PR #9, opened 2017-06-21
Update teleport.qasm

esabo:
Use of cregs inconsistent with other example files. Updated such that the Circuit reads  [('c', 0), ('c',1),('c',2)] and not read [('c0', 0), ('c1',0),('c2',0)].


awcross1:
Hi Eric, I'd like to keep this example synchronized with the document. After one of us updates the corresponding figure in the document, I'll accept this change. Thanks!

awcross1:
A simpler solution is to accept the change under a different file name, since there would be a couple examples to change in the document otherwise.

PR #15, opened 2017-07-27
translate qasm2.tex to qasm2.rst

ewinston:


jaygambetta:
@ewinston we should change this pull request a bit. I would rather not just convert the tex to rst but change it to rst so that we can edit it over time and now always have to convert the original white paper. The paper we have placed on the arxiv now so that it has a reference if need be.

ewinston:
@jaygambetta I fixed some things and removed the tex. GitHub doesn't generate it's view quite as well as the intended Sphinx; latex equations and numbered figure references don't appear like the Sphinx html output.

PR #26, opened 2018-01-25
Fixing issue in RandomCircuits.py: claimed SU(4) was actually U(4)

ajavadia:
Now it is SU(4)


ajavadia:
Issue reported by @levbishop

PR #28, opened 2018-03-14
Add Benchmark Tool and Scenario

hitomitak:
Add benchmark tool, benchmark/run_simbench.py and some benchmark scenario (qasm files). 
The detailed description about the tools, please see README.md file.


jaygambetta:
I think these are good. But should they go in openqasm or qiskit.

ajavadia:
@hitomitak Thank you very much for these!

Some comments:
1- There are 2 folders named SAT and sat, which is redundant. Can you please delete SAT? 

2- Please make sure every folder has a single generator script in python, with consistent naming (e.g. foldername_gen.py). Please remove the .sh scripts, the python scripts should suffice for creating problem instances. 

3- Similarly, please remove `cc/common.py` and `bv/common.py` if they are not doing something essential.

4- The random_circuits, I think we can call them "quantum_volume" or ("volume" for short) so it is more descriptive of what they are trying to accomplish.

5- I don't really understand what the "-l" flag does. Can you clarify a bit?

6- There are notebook implementations `bv` and `cc` as well, can you add them to the bottom of the README like you have for `qft`?

hitomitak:
@ajavadia Thank you for the comments. I updated the code as follows.

- Delete SAT dir
- Rename random app to quantum_volume 
- Delete sh scripts and rename generator python script
- Merge common.py and generator python script.
- Update README file 
- Add option to set seed number

ajavadia:
Thank you for the contribution @hitomitak! This is a very useful set of benchmarks to have.

I have merged it here. But can you also make a pull request in the QISKit repo? It would be good to have these under `test/qasm`. But there are currently too many files for that repo. Can you just keep the generator script, plus 5-10 examples from each app? This way we would keep the QISKit package clean, but still provide a way for others to generate their own instances.

PR #30, opened 2018-04-12
Update quantum_volume generation file and remove scenario files

hitomitak:
Add a function to quantum_volume/quantum_volume.py to execute simulation. 
If add 'r' option, quantum_volume.py execute a simulation instead of generating qasm file. 

Moreover,  we changed the following items:
- Delete some scenario files  ( https://github.com/QISKit/openqasm/pull/28 )
- Modify code style for PEP8


ajavadia:
Thanks!

PR #33, opened 2018-05-15
Corrected a small spelling mistake.

RolfHuisman:
The grammar written down is not really BNF as stated, but based on the context I think there is an 'x' missing.


ajavadia:
Thanks!

PR #34, opened 2018-05-22
fix bugs in quantum_volume.py and add random seed

itoko:


jaygambetta:
closed as moved to qiskit

PR #35, opened 2018-06-14
Support ProjectQ and Q#

hitomitak:
We supported ProjectQ and Q# as a benchmark backend. 
If ProjectQ or Q# are installed in the host, we can run the benchmark on them.
We can specify the backend as --b option as follows. 
```
$ python3 run_simbench.py -a qft -b Qsharp -s 10 -e 20
$ python3 run_simbench.py -a qft -b ProjectQ -s 10 -e 20
```


jaygambetta:
closed as benchmark should be in qiskit

PR #47, opened 2019-07-25
bring over examples from terra

ajavadia:


jaygambetta:
i dont think we need these examples

PR #53, opened 2020-04-10
chsh game only for ibmq_qasm_simulator .

AppTrain:
This is an implementation of the CHSH Game demonstrating the CHSH Inequality.

The example uses the H gate on the first 2 qubits to randomize Alice and Bob's inputs. Two more gates are entangled to store their outputs. The input gates are measured, Then the IBM 'if Gate' executes both their strategies, then the output gates are measured. 

This will only run on the ibmq_qasm_simulator as it uses the 'If operation' .  It passes the conformance tests.

### DIFF
--- a/examples/generic/chsh.qasm
+++ b/examples/generic/chsh.qasm
@@ -1,0 +1,30 @@
+OPENQASM 2.0;
+include "qelib1.inc";
+
+qreg q[4];
+creg c[4];
+
+// Bob's input
+h q[0];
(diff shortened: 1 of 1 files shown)


delapuente:
I'm retargeting this PR to the [OpenQASM2.x](https://github.com/Qiskit/openqasm/tree/OpenQASM2.x) branch.

jaygambetta:
closing as I dont think this is a general example.

PR #58, opened 2020-11-03
Add infrastructure and first text for OpenQASM live spec

delapuente:
This PR adds the infrastructure for building a live specification and the base text for OpenQASM 3.0.

The tool for generating the specification is Sphinx and the base text is a conversion made with [pandoc](https://pandoc.org/) fo the original LaTeX files. The base text needs manual fixes to accurately mimic the `.tex` files.


jaygambetta:
cant we delete all the tex now we have rst and not use that going forward. Hence do we need the tools not just the content

jaygambetta (approved):

jaygambetta:
im going to approve so that we get this going -- I make an issue to remove the building files

PR #77, opened 2020-11-10
removing old include files

jaygambetta:


delapuente (approved):
This is fine, we should also review other outdated example files.

PR #81, opened 2020-11-11
Comments and RFC for OpenQASM 3.0 Types

steleman:
Request For Comments for OpenQASM 3.0 Types.

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
 percent sign, or an element from the Unicode character categories
 Lu/Ll/Lt/Lm/Lo/NI :cite:`noauthorUnicodeNodate`.
 Continuation characters may contain numbers. Variable identifiers may
-not override a reserved identifier. All qubits are global variables.
+not override a reserved identifier.
+
+All qubits are global variables.
+
 When qubits are declared, their state is initially undefined. Qubits
 cannot be declared within gates or subroutines. This simplifies OpenQASM
 significantly since there is no need for quantum memory management.
-However, it also means that users or compiler have to explicitly manage
-the quantum memory. In addition to being assigned values within a
-program, all of the classical types can be initialized on declaration.
+However, it also means that users or compilers have to explicitly manage
+the quantum memory.
+
+OpenQASM supports all scalar classical types as defined in the C99 Standard:
+http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
+
+Additionally, OpenQASM 3.0 supports compile-time fixed-size C-style bit
+arrays.  Variable-Length Arrays (VLA's) are not supported in OpenQASM 3.0.
+
+Missing - Needs Clarification:
+------------------------------
+Does an Array automatically decay to the pointer of its first element, as in C?
+-------------------------------------------------------------------------------
+Can an Array be passed as argument to a function or subroutine, as in C?
+------------------------------------------------------------------------
+Signed / Unsigned
+-----------------
+Overflow / Underflow
+--------------------
+Wrap-Around (unsigned)
+----------------------
+
+In addition to being assigned values within a program, all scalar classical
+types can be initialized at declaration.
+
+Missing - Needs Specfication:
+-----------------------------
+Full enumeration of supported scalar classical types and their sizes.
+---------------------------------------------------------------------
+
 Multiple comma-separated declarations can occur after the typename with
-optional assignment on declaration for each. Any classical variable or
-Boolean that is not explicitly initialized is undefined. Classical types
-can be mutually cast to one another using the typename . We use the
+optional assignment on declaration for each. The value of any scalar
+classical variable or Boolean that is not explicitly initialized is
+undefined.
+
+Classical types can be cast to one another using the typename. We use the
 notation to denote the width and precision of fixed point numbers, where
 is the number of sign bits, is the number of integer bits, and is the
 number of fractional bits. It is necessary to specify low-level
 classical representations since OpenQASM operates at the intersection of
 gates/analog control and digital feedback and we need to be able to
 explicitly transform types to cross these boundaries. Classical types
-are scoped to the braces within which they are declared.
+declared within a braced block are scoped to the braces within which they
+are declared. Globally-declared scalar classical types have Translation
+Unit scope.
+
+Missing - Needs clarification:
+------------------------------
+Casting and conversion rules for scalar classical types.
+--------------------------------------------------------
 
 Quantum types
 -------------
-
 Qubits
 ~~~~~~
 
@@ -114,6 +152,23 @@ conversion will be done assuming little-endian bit ordering.
    // Declare a 32 bit signed integer
    int[32] my_int;
 
+Needs clarification:
+--------------------
+If OpenQASM supports "classical types", then it implicitly declares
+-------------------------------------------------------------------
+support for normative, known fixed-size C types.  Paragraph above seems
+-----------------------------------------------------------------------
+inconsistent.
+-------------
+Can Bit Arrays be cast (implicitly or explicitly) to an Integer type
+--------------------------------------------------------------------
+or a Floating-Point type?
+-------------------------
+What happens during a cast between two scalar types of different bit width?
+---------------------------------------------------------------------------
+What is the bit width of the resulting type?
+--------------------------------------------
+
 Signed fixed-point numbers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -139,6 +194,16 @@ vendors may not support manipulating these values at run-time.
    // Declare a single-precision 32-bit float
    float[32] my_float = π;
 
+Needs clarification:
+--------------------
+IEEE-754 implies support for implicit 32-bit float and 64-bit double.
+---------------------------------------------------------------------
+We can apply IEEE-754 rounding rules to fixed-width floating-point
+------------------------------------------------------------------
+types via GNU MPFR:
+-------------------
+https://www.mpfr.org/
+
 Fixed-point angles
 ~~~~~~~~~~~~~~~~~~
 
@@ -157,6 +222,11 @@ compatible with run-time values on some platforms.
    // Declare an angle with 20 bits of precision
    angle[20] my_angle;
 
+Needs clarification:
+--------------------
+"angle" is a reserved keyword? (based on the example above, yes it is).
+-----------------------------------------------------------------------
+
 Boolean types
 ~~~~~~~~~~~~~
 
@@ -170,6 +240,20 @@ be true and 0 will be false.
    bool my_bool;
    // Assign a cast bit to a boolean
    my_bool = bool(my_bit);
+
+Needs clarification:
+--------------------
+Wouldn't a C-style cast
+-----------------------
+
+.. code-block:: c
+  my_bool = (bool) my_bit;
+
+be more consistent with C? The cast example above is more like an
+-----------------------------------------------------------------
+explicit construction / conversion from C++.
+--------------------------------------------
+
 
 Real constants
 ~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


source/language/types.rst
```diff
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
...
-the quantum memory. In addition to being assigned values within a
-program, all of the classical types can be initialized on declaration.
+However, it also means that users or compilers have to explicitly manage
+the quantum memory.
+
+OpenQASM supports all scalar classical types as defined in the C99 Standard:
+http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
+
+Additionally, OpenQASM 3.0 supports compile-time fixed-size C-style bit
+arrays.  Variable-Length Arrays (VLA's) are not supported in OpenQASM 3.0.
```
blakejohnson, line 29: Correct, but I do want the ability to refer to an arbitrary fixed-size array in the type signature to a `def` subroutine, i.e. `foo(int[32] a[])` should allow `a` to be any fixed-size array of int[32]s.

source/language/types.rst
```diff
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
...
 cannot be declared within gates or subroutines. This simplifies OpenQASM
 significantly since there is no need for quantum memory management.
-However, it also means that users or compiler have to explicitly manage
-the quantum memory. In addition to being assigned values within a
-program, all of the classical types can be initialized on declaration.
+However, it also means that users or compilers have to explicitly manage
+the quantum memory.
+
+OpenQASM supports all scalar classical types as defined in the C99 Standard:
+http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
```
blakejohnson, line 26: As written, the spec supports more than that because the field sizes are arbitrary, i.e. `int[10]` is valid OpenQASM3. This is more akin to numeric types in VHDL and Verilog.

source/language/types.rst
```diff
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
...
+
+Additionally, OpenQASM 3.0 supports compile-time fixed-size C-style bit
+arrays.  Variable-Length Arrays (VLA's) are not supported in OpenQASM 3.0.
+
+Missing - Needs Clarification:
+------------------------------
+Does an Array automatically decay to the pointer of its first element, as in C?
```
blakejohnson, line 33: What do you mean by "decay"? We don't have pointers in OpenQASM3, so I am not sure what you mean.

source/language/types.rst
```diff
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
...
+-------------------------------------------------------------------------------
+Can an Array be passed as argument to a function or subroutine, as in C?
```
blakejohnson, line 35: Yes.

source/language/types.rst
blakejohnson: I will follow up to note that particular implementations are allowed to support a finite set of these types, i.e. your compiler might only allow `int[8]`, `int[16]`, `int[32]`, etc.

source/language/types.rst
```diff
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
...
+--------------------
+Wrap-Around (unsigned)
+----------------------
+
+In addition to being assigned values within a program, all scalar classical
+types can be initialized at declaration.
+
+Missing - Needs Specfication:
+-----------------------------
+Full enumeration of supported scalar classical types and their sizes.
```
blakejohnson, line 49: This should be a compiler specification as opposed to a language specification.

source/language/types.rst
```diff
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
...
+Additionally, OpenQASM 3.0 supports compile-time fixed-size C-style bit
+arrays.  Variable-Length Arrays (VLA's) are not supported in OpenQASM 3.0.
+
+Missing - Needs Clarification:
+------------------------------
+Does an Array automatically decay to the pointer of its first element, as in C?
+-------------------------------------------------------------------------------
+Can an Array be passed as argument to a function or subroutine, as in C?
+------------------------------------------------------------------------
+Signed / Unsigned
```
blakejohnson, line 37: +1

source/language/types.rst
```diff
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
...
+-----------------
+Overflow / Underflow
+--------------------
+Wrap-Around (unsigned)
```
blakejohnson, line 41: Good point. At least with the `angle[n]` type we explicitly want wrap-around throwing an errors. For consistency sake, we might want to adopt this behavior everywhere.

source/language/types.rst
```diff
@@ -12,28 +12,66 @@ Variable identifiers must begin with a letter [A-Za-z], an underscore, a
...
 gates/analog control and digital feedback and we need to be able to
 explicitly transform types to cross these boundaries. Classical types
-are scoped to the braces within which they are declared.
+declared within a braced block are scoped to the braces within which they
+are declared. Globally-declared scalar classical types have Translation
+Unit scope.
+
+Missing - Needs clarification:
+------------------------------
+Casting and conversion rules for scalar classical types.
```
blakejohnson, line 70: +1

source/language/types.rst
```diff
@@ -114,6 +152,23 @@ conversion will be done assuming little-endian bit ordering.
...
+-----------------------------------------------------------------------
+inconsistent.
+-------------
+Can Bit Arrays be cast (implicitly or explicitly) to an Integer type
+--------------------------------------------------------------------
+or a Floating-Point type?
+-------------------------
+What happens during a cast between two scalar types of different bit width?
+---------------------------------------------------------------------------
+What is the bit width of the resulting type?
```
blakejohnson, line 169: This last question seems like it is determined by the cast. i.e. if I write `int[16](x)` where `x` is an `uint[8]` then I am explicitly asking for 16-bit integer output.

source/language/types.rst
```diff
@@ -114,6 +152,23 @@ conversion will be done assuming little-endian bit ordering.
...
 
+Needs clarification:
+--------------------
+If OpenQASM supports "classical types", then it implicitly declares
+-------------------------------------------------------------------
+support for normative, known fixed-size C types.  Paragraph above seems
+-----------------------------------------------------------------------
+inconsistent.
+-------------
+Can Bit Arrays be cast (implicitly or explicitly) to an Integer type
```
blakejohnson, line 163: I would certainly like this. We should propose syntax for it.

source/language/types.rst
```diff
@@ -157,6 +222,11 @@ compatible with run-time values on some platforms.
    // Declare an angle with 20 bits of precision
    angle[20] my_angle;
 
+Needs clarification:
+--------------------
+"angle" is a reserved keyword? (based on the example above, yes it is).
```
blakejohnson, line 227: Yes, it is a reserved keyword.

source/language/types.rst
```diff
@@ -171,6 +241,20 @@ be true and 0 will be false.
...
+Wouldn't a C-style cast
+-----------------------
+
+.. code-block:: c
+  my_bool = (bool) my_bit;
+
+be more consistent with C? The cast example above is more like an
+-----------------------------------------------------------------
+explicit construction / conversion from C++.
+--------------------------------------------
```
blakejohnson, line 255: I definitely prefer the conversion syntax we have proposed. We will need a more functional form to support some of the things above where we need to define how a value is extended or contracted into smaller/larger types.

source/language/types.rst
```diff
@@ -222,6 +306,24 @@ namespace are listed in table `1 <#tab:real-constants>`__.
       | Euler’s number                | euler_gamma  | :math:`e`    | 2.7182818284...     |
       +-------------------------------+--------------+--------------+---------------------+
 
+Needs clarification:
+--------------------
+What is the default width of the my_const variable in the following
```
blakejohnson, line 311: We are thinking about making `const`s explicitly typed as well. Otherwise it has an implicit type of BigFloat/BigInt like constants in python.

source/language/types.rst
```diff
@@ -252,6 +354,13 @@ timing resolution stage of the compiler, higher order stretches will
 suppress lower order stretches whenever they appear in the same scope on
 the same qubits.
 
+Needs clarification:
+--------------------
+'length' and 'stretch' are reserved keywords.
+---------------------------------------------
+What is the default width for 'length' and 'stretch' types?
```
blakejohnson, line 361: They don't have one. These are compile-time values.

source/language/types.rst
steleman: I wasn't clear with what I mean by "decay", so, here goes:

In C - and C-like derivatives - an array of type T automatically "decays" to a pointer of type T* to said array's first element. This means I can write something like this:

```
int array[10];
int* p = array;
```
Now I can access the elements of array either by index (```array[0], array[1]```, etc), or by pointer arithmetic (```*p, *(p+1), (*p+2)```, etc, etc, etc).

Why is this important: Since OpenQASM 3.0 supports passing arrays as arguments to functions and subroutines, we need to specify **how** these array arguments are passed. If we want to do it "like C does it", then we must recognize that, in C, array function arguments are passed as pointers, not as arrays. I.e. only the pointer to the array's first element is passed as argument. If we don't want to do it "like C does it", then we are implicitly doing it "like FORTRAN does it". Which means, the entire array is passed in as argument.

These two ways of passing arrays as function arguments are mutually exclusive.

Since you indicated above that you would like to have the ability to write ```foo(int[32], a[])``` -- i.e. pass an array of arbitrary size as a function argument -- that implies support for implicit array-to-pointer conversion. I.e. array "decay". I.e. we are implicitly supporting pointers.

source/language/types.rst
steleman: What about the case where there's loss:

```
uint[16] x;
int[8](x);
```

Possibilities:

- We throw an error because the cast to 8 bits from 16 bits won't fit.
- We silently truncate and continue with potentially catastrophic results.
- We truncate, we issue a warning and we continue with potentially catastrophic results.

If we truncate:

- Which set of bits does the resulting cast get? The most significant bits or the least significant bits?

source/language/types.rst
taalexander: We also do not support some types like `short` which are encompassed by `int[]`.

ndebeaudrap:
Is OpenQASM3 intended to be 100% backwards compatible with OpenQASM2? For instance, is it a design constraint that a gate declaration with no type annotations such as `gate rotZ (theta) q { U(0,0,theta) q; }` be valid OpenQASM3?

taalexander:
> Is OpenQASM3 intended to be 100% backwards compatible with OpenQASM2? For instance, is it a design constraint that a gate declaration with no type annotations such as `gate rotZ (theta) q { U(0,0,theta) q; }` be valid OpenQASM3?

It is not intended to 100% backward compatible. We are comfortable making breaking changes where it makes sense. For example, the `opaque` keyword has been dropped in favor of `defcal` statements and `gate` declarations.

ndebeaudrap:
I think that the ability to support a range of significantly different (and somewhat unusual) numerical types is a nice idea. However, any hardware platform is likely to support only a few such types. For instance, it would be pretty reasonable for one platform to support only float[16], float[32], and int[32], and another to support only angle[8], fixed[0,32], and int[64]. Do you provide these different types merely to facilitate the final stages of code-generation, e.g., by allowing OpenQASM in principle to express a range of register types that different hardware may support? Or do you have in mind to develop a standard for how code written to support one set of numerical types should be interpreted on hardware which accommodates other types?

taalexander:
> I think that the ability to support a range of significantly different (and somewhat unusual) numerical types is a nice idea. However, any hardware platform is likely to support only a few such types. For instance, it would be pretty reasonable for one platform to support only float[16], float[32], and int[32], and another to support only angle[8], fixed[0,23], and uint[64]. Do you provide these different types merely to facilitate the final stages of code-generation, e.g., by allowing OpenQASM in principle to express a range of register types that different hardware may support? Or do you have in mind to develop a standard for how code written to support one set of numerical types should be interpreted on hardware which accommodates other types?

I would agree that this falls more into the category of "nice idea" than a necessary idea. Predominantly the use cases are for targeting backend hardware by allowing experimentalist and calibration services to finetune routines for target hardware and easing the lowering process to the target hardware. These concepts are much less relevant at the logical program level. 

It should also be noted that each set of hardware will likely only support some small subset of these types regardless. Perhaps the concept of a system size like `isize`, `usize` etc. is more convenient. Writing a standard for conversion of these types in a standard way to runtime systems would be a daunting task and I would prefer to avoid it. In practice I imagine parameterized types would just be cast to the system size with precision being lost in the process.


My apologies for the delay in my response.

PR #82, opened 2020-11-13
Types and Casting review

ismaelfaro:
- Reorganize part of the description of the **Qubit** definition
- add extra examples in **Register concatenation and slicing**


source/language/types.rst
```diff
@@ -49,6 +47,12 @@ of this register, where
 Qubits are initially in an undefined state. A quantum operation is one
 way to initialize qubit states.
 
+All qubits are global variables.
+Qubits cannot be declared within gates or subroutines. This simplifies OpenQASM
+significantly since there is no need for quantum memory management.
+However, it also means that users or compiler have to explicitly manage
+the quantum memory. 
+
```
ismaelfaro, line 55: This parragraph was in in the Generalities, and I considered that is better to move it to the **Qubit** section, because it make reference to the Qubits itself.

ismaelfaro (commented):
comments about the changes

blakejohnson (approved):

PR #85, opened 2020-11-13
Restore dropped \code{..} sections from TeX to RST conversion.

blakejohnson:
The conversion through pandoc dropped some `\code{}` sections. This PR should restore all of them. The sections reviewed are:

* `gates`
* `types`
* `insts`
* `classical`
* `subroutines`
* `pulses`
* `delays`

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -150,10 +150,10 @@ vendors may not support manipulating these values at run-time.
 Fixed-point angles
 ~~~~~~~~~~~~~~~~~~
 
-Fixed-point angles are interpreted as :math:`2\pi` times a 0:1:n-1
+Fixed-point angles are interpreted as 2π times a 0:1:n-1
 fixed-point number. This represents angles in the interval
 :math:`[0,2\pi)` up to an error :math:`\epsilon\leq \pi/2^{n-1}` modulo
-:math:`2\pi`. The statement ``angle[size] name;`` declares an n-bit angle. OpenQASM3
+2π. The statement ``angle[size] name;`` declares an n-bit angle. OpenQASM3
 introduces this specialized type because of the ubiquity of this angle
 representation in phase estimation circuits and numerically controlled
 oscillators found in hardware platform. Note that defining gate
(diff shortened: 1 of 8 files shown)


source/language/types.rst
```diff
@@ -226,9 +226,9 @@ namespace are listed in table `1 <#tab:real-constants>`__.
       +-------------------------------+--------------+--------------+---------------------+
       | Constant                      | Alphanumeric | Unicode      | Approximate Base 10 |
       +-------------------------------+--------------+--------------+---------------------+
-      | (r)1-1(lr)2-2(rl)3-3(l)4-4 Pi | pi           | :math:`\pi`  | 3.1415926535...     |
+      | Pi                            | pi           | π            | 3.1415926535...     |
       +-------------------------------+--------------+--------------+---------------------+
-      | Tau                           | tau          | :math:`\tau` | 6.283185...         |
+      | Tau                           | tau          | τ            | 6.283185...         |
       +-------------------------------+--------------+--------------+---------------------+
       | Euler’s number                | euler_gamma  | :math:`e`    | 2.7182818284...     |
```
taalexander, line 233: I think this is incorrect and the Euler gamma is Euler's constant rather than Euler's number. This should probably just be `euler`. I am not sure if there is a good unicode character for Euler's number other than https://www.compart.com/en/unicode/U+0065.

blakejohnson:
There are missing code blocks in the other files. Attempting to fix.

blakejohnson:
Okay, I think I grabbed all the missing `\code{...}` blocks.

source/language/types.rst
blakejohnson: Fair point, but I wonder if we should actually drop this one from the table and instead just add `exp(...)`` to the built-in scientific calculator functions?

taalexander: 👍  to that.

delapuente (approved):
This is great! Thank you a lot @blakejohnson

source/language/types.rst
delapuente: @taalexander @blakejohnson maybe we can open a follow up for reviewing symbols, in general, trying to use more semantic ones, if possible.

PR #87, opened 2020-11-17
Add code of conduct

delapuente:


ismaelfaro:
I think that can be better link to the Qiskit code of conduct, if finally the content is the same

https://github.com/Qiskit/qiskit/blob/master/CODE_OF_CONDUCT.md

delapuente:
@ismaelfaro done

delapuente:
I've merged this since it is the same case as in https://github.com/Qiskit/qiskit.org/pull/989

PR #89, opened 2020-11-25
ANTLR Grammar

zachschoenfeld33:
### Summary
ANTLR grammar for OpenQASM3. 

### Details
- Tested by generating Python ANTLR parser
- Updated documentation
- Unit tests which compare parsed snippets of OpenQasm3 to expected parse trees
- Examples parsed and validated that no error is thrown my the grammar

### ToDo
- More unit tests (so far just have a few)
- Validate rest of examples (so far just have `adder.qasm`)


ismaelfaro:
we should pass all the "examples", https://github.com/Qiskit/openqasm/tree/master/examples 

we could generate the parser for python or and do a Github action to test all of them 😁. @delapuente any suggestion on that?

source/grammar/qasm3.g4
```diff
@@ -1,98 +1,461 @@
...
+
+// Register Concatenation and Slicing
+concatenateExpression
+    : ASSIGN
+    ( Identifier range
+    | Identifier '||' Identifier
+    | Identifier LBRACKET expressionList RBRACKET )
+    ;
+
+range
```
ismaelfaro, line 189: it's necessary change this "range" identifier to don't collide with Python range.
- error: symbol "range" conflicts with generated code in target language or runtime

source/grammar/qasm3.g4
```diff
@@ -1,98 +1,461 @@
...
+LPAREN : '(' ;
+RPAREN : ')' ;
+
+COLON: ':' ;
+SEMICOLON : ';' ;
+
+DOT : '.' ;
+COMMA : ',' ;
+
+AnyString : .*? ;
```
ismaelfaro, line 443: warning: non-fragment lexer rule AnyString can match the empty string

source/grammar/qasm3.g4
```diff
@@ -1,98 +1,461 @@
...
+calibration
+    : calibrationGrammarDeclaration
+    | calibrationDefinition
+    ;
+
+calibrationGrammarDeclaration
+    : 'defcalgrammar' calibrationGrammar SEMICOLON
+    ;
+
+calibrationDefinition
```
ismaelfaro, line 409: warning: rule calibrationDefinition contains an optional block with at least one alternative that can match an empty string

ismaelfaro (commented):
review with ANTLR Python3

delapuente:
For testing the parser we should construct some AST and take them as reference. Sure the parser should work for the examples but I don't know if they are the best suite for testing and debugging since the ASTs of the examples are extremely complex. We could:

1. Sart by just building them and trusting the output is correct.
2. Create a series of "unit tests" for small snippets exercising the main aspects of the parser.
3. Rebuild the first set as we find problems with the current implementations.

source/grammar/qasm3.g4
zachschoenfeld33: Addressed in https://github.com/Qiskit/openqasm/pull/89/commits/c3b3b28409aec991623c590a5ecfe28c7527d14c.

source/grammar/qasm3.g4
zachschoenfeld33: Addressed in https://github.com/Qiskit/openqasm/pull/89/commits/c3b3b28409aec991623c590a5ecfe28c7527d14c.

source/grammar/qasm3.g4
zachschoenfeld33: Addressed in https://github.com/Qiskit/openqasm/pull/89/commits/c3b3b28409aec991623c590a5ecfe28c7527d14c.

zachschoenfeld33:
Have created a number of unit tests and passed `adder.qasm` example.

delapuente (dismissed):
There is a lot of work here and we should merge it. Next steps being integrating the test in our development cycle for when changes happen in the grammar they are automatically run.

It would be also convenient to automatize the load an execution of `yaml` tests, without explicitly listing them in the test class.

However, former issues can be addressed in follow-ups. This PR presents a solid foundation on which continue working.

Nice work @zachschoenfeld33.

zachschoenfeld33:
Incorporates https://github.com/Qiskit/openqasm/pull/122.

zachschoenfeld33:
@taalexander @lcapelluto @delapuente @blakejohnson @steleman @mbhealy

zachschoenfeld33:
We will also need to make sure to update this as updates are made to the spec.

delapuente (approved):
Moving forward, this is a good foundation. Thank you a lot @zachschoenfeld33!

PR #103, opened 2020-12-17
Fix typos I found while reading OpenQASM 3.0 spec

stevenheidel:
I also pinned the version of sphinxcontrib-bibtex since they just released a new version which broke the build.

PR #104, opened 2020-12-17
Clarify that division is integer division

stevenheidel:
Assuming this is the case, so I clarified that and updated the example.


stevenheidel:
Build fails because the version of -bibtex needs to be pinned like in my other PR #103

blakejohnson (approved):
I agree that this makes sense.

blakejohnson (approved):

PR #108, opened 2020-12-22
fix wrong notations in examples/

hhorii:
This PR fixes several wrong statements in `examples`.

Comment: I'm not sure whether `stretch` can be in an expression because it is a type.

### DIFF
--- a/examples/dd.qasm
+++ b/examples/dd.qasm
@@ -5,9 +5,9 @@
 OPENQASM 3.0;
 include "stdgates.inc";
 
-length start_stretch = -0.5 * lengthof({x %0;}) + stretch
-length middle_stretch = -0.5 * lengthof({x %0;}) - 5 * lengthof({y %0;} + stretch
-length end_stretch = -0.5 * lengthof({y %0;}) + stretch
+length start_stretch = -0.5 * lengthof({x %0;}) + stretch;
+length middle_stretch = -0.5 * lengthof({x %0;}) - 5 * lengthof({y %0;}) + stretch;
+length end_stretch = -0.5 * lengthof({y %0;}) + stretch;
 
 box {
   delay[start_stretch] %0;
--- a/examples/varteleport.qasm
+++ b/examples/varteleport.qasm
@@ -27,7 +27,7 @@ rz(pi / 4) input;
 
 let io = input;
 for i in [0: n_pairs - 1] {
-  let bp = q[2 * i, 2 * i + 1];
+  let bp = q[2 * i: 2 * i + 1];
   bit pf[2];
   bellprep bp;
   cx io, bp[0];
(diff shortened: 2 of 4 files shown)


examples/vqe.qasm
```diff
@@ -59,7 +59,7 @@ def trial_circuit qubit[n]:q {
 /* Apply VQE ansatz circuit and measure a Pauli operator
  * given by spec. Return the number of 1 outcomes.
  */
-def counts_for_term(bit[2*n]:spec) qubit[n] -> uint[prec] {
+def counts_for_term(bit[2*n]:spec) qubit[n]:q -> uint[prec] {
```
taalexander, line 62: I think these examples have additional errors. I do not believe we support calling subroutines like gates, and the qubit `qubit: q` parameter should be included within the parentheses. The notation for specifying the size of the qubit array seems incorrect as well. I believe it should be something like:
```suggestion
def counts_for_term(bit[2*n]:spec, qubit[]:q) -> uint[prec] {
```
For now `n` is a constant that can be used within the subroutine for iteration bounds, however, the language should likely support a `length/sizeof` subroutine builtin for inquiring the size of an array.

taalexander (changes requested):

examples/vqe.qasm
hhorii: @taalexander do you mean that [the description of subroutine](https://qiskit.github.io/openqasm/language/subroutines.html#subroutines) will be changed? An example `xcheck` looks like a gate.
```
def xcheck qubit[4]:d, qubit:a -> bit {
  reset a;
  for i in [0: 3] cx d[i], a;
  return measure a;
}
```

examples/dd.qasm
```diff
@@ -5,9 +5,9 @@
 OPENQASM 3.0;
 include "stdgates.inc";
 
-length start_stretch = -0.5 * lengthof({x %0;}) + stretch
-length middle_stretch = -0.5 * lengthof({x %0;}) - 5 * lengthof({y %0;} + stretch
-length end_stretch = -0.5 * lengthof({y %0;}) + stretch
+length start_stretch = -0.5 * lengthof({x %0;}) + stretch;
```
zachschoenfeld33, line 8: I do not think `stretch` should be allowed to be defined w/ out a variable like this. I believe this should only occur for `stretchinf`. Instead we should have a variable defined above `stretch s;` and then use that

examples/varteleport.qasm
```diff
@@ -27,7 +27,7 @@ rz(pi / 4) input;
 
 let io = input;
 for i in [0: n_pairs - 1] {
-  let bp = q[2 * i, 2 * i + 1];
+  let bp = q[2 * i: 2 * i + 1];
```
zachschoenfeld33, line 30: The comma syntax is valid--see https://qiskit.github.io/openqasm/language/types.html#register-concatenation-and-slicing. it picks out the particular entries of interest. I think they are actually equivalent in this case, as the range is inclusive.

examples/vqe.qasm
zachschoenfeld33: I believe we resolved that qubit args are not in the parentheses--see https://github.com/Qiskit/openqasm/pull/115/files.

hhorii:
All the examples pointed in this PR were fixed.

Issue #102, opened 2020-12-16
defcal for reset


jaygambetta:
Currently, we do not support a defcal for reset. 

options 
add three gates, reset measurement 

or 
can we get it all under one defcal

blakejohnson:
Just to expand upon what Jay wrote a bit. The central issue here is: what is an acceptable target for a `defcal`. Right now it is `gate`s + `measure` as a special case. However, that leaves open the question of how to handle `reset`, which definitely is problematic because we have typically said that `defcal`s shouldn’t contain control flow, because that breaks the single-threaded control model that is implicit in OpenQASM3.

So, a couple options that were discussed:
1. Allow `defcal` to also target subroutines (`def`). _Runs into the issue above about control flow._
2. Introduce a concept of a “operation” to QASM3 which is broader than `gate`  but narrower than `def`. Add `measure` and `reset` to this family but also allow the user to add other things.
3. Add an additional special case for `reset` to `defcal` .

jaygambetta:
As we discuss I initially preferred 3 but now like 2 as we can add operations to qasm and then define them to be used like gates. For instance I could envision that we may want to add `D_phase(gamma)` where we do a dephasing channel which could be implements by a `defcal` which could be done by adding white noise to a parameter that sets the qubit frequency or entangles with a second qubit and measures it there by doing this operation back on the qubit of interest. 

I think we should start going this way (not sure of why for applications yet - maybe there is some bath engineering examples that could be interesting) but for studying noise effects could be interesting and as such I would add operation to qasm which is a bigger than a gate but less than a subroutine but it can not support conditional operations.

stevenheidel:
In (2) there could also be additional `measure`-like operations that return level 0 or level 1 data, rather than a `bit.

taalexander:
> In (2) there could also be additional `measure`-like operations that return level 0 or level 1 data, rather than a `bit.

This could add significant complexity - two potential use-cases:
- Returning qudit measurements
- Real-time control/measurement bootstrapping routines in OpenQASM3.

levbishop:
To flesh out Blake's suggestion (2):
we could allow `defcal` to also target a new thing more general than unitary gate. Call it "foo" for now because choosing the correct name may be tricky. Just as 
```
gate mygate(type1: param1, type2: param2 ) q1, q2 
{ 
   <mygate_body>
}```
defines a unitary gate in order to teach the compiler what the meaning of mygate is, lets have
```
[...]

PR #109, opened 2020-12-28
Pulse-level control instructions

stevenheidel:
Proposal to add pulse-level control instructions to OpenQASM, in the form of an OpenPulse defcalgrammar. This represents several weeks of weeks of work from the OpenPulse working group.

Co-authored with @psivaraj

Tracking issue: #110

### DIFF
--- a/source/language/index.rst
+++ b/source/language/index.rst
@@ -22,3 +22,4 @@ Appendix `[app:semantics] <#app:semantics>`__ gives formal semantics.
    directives
    delays
    pulses
+   openpulse
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -126,7 +126,7 @@ The relevant instructions are:
 - ``set_phase`` and ``shift_phase``
 - ``set_scale`` and ``shift_scale``
 
-Here's an example of using ``shift_scale`` to calibrate an ``rz`` gate:
+Here's an example of using ``shift_phase`` to calibrate an ``rz`` gate:
 
 .. code-block:: none
 
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -1,0 +1,316 @@
+OpenPulse Grammar
+=================
+
+In addition to OpenQASM instructions, ``defcal`` blocks may contain OpenPulse 
+instructions. Or equivalently stated: OpenPulse instructions cannot appear
+outside of a ``defcal`` block.
+
+These instructions are motivated by the original OpenPulse specification which
+was defined as a JSON wire-format for pulse-level quantum programs in the paper
+`Qiskit Backend Specifications for OpenQASM and OpenPulse Experiment <https://arxiv.org/abs/1809.03452>`_. 
+The text format described here has several advantages over the equivalent JSON 
+format:
+
+- It is more readable
+- Absolute time is handled through built-in timekeeping instructions
+- Gates and classical instructions can be mixed in with the pulses to create far richer calibrations
+- Pulse definitions are tied to circuit instructions rather than circuit programs
+- Richer handling of frames
+
+Channels
+--------
+
+Channels map to a hardware resource, which can play pulses to manipulate a qubit
+or capture signal from the qubit to perform a measurement. There is a many-to-many
+relationship between qubits and channels. One qubit may have multiple channels
+connecting to it. Pulses on different channels would have different physical
+interactions with that qubit. A channel may also have many qubits. For instance,
+a channel could manipulate the coupling between two neighboring qubits, or
+could even reference multiple qubits in a chain.
+
+Channels are vendor-specific and device-specific. It is expected that vendors
+of quantum hardware provide the appropriate channel names and qubit mappings
+for use by end users. Each channel may also have associated static settings,
+such as local-oscillator frequencies, which do not vary throughout program
+execution. Again it is expected that vendors of quantum hardware provide a
+method for manipulating those static settings if appropriate.
+
+There are two kinds of channels: transmit channels (sending input to a quantum
+device) and receive channels (reading output from a quantum device).
+
+The ``channel`` type is a compile-time type representing a channel.
+
+A value of type ``channel`` is retreived with the special functions
+``transmit`` and ``receive``. These functions take an optional string argument
+followed by 1 or more physical qubits.
+
+The hardware vendor, during compilation, will substitute valid calls to
+``transmit`` and ``receive`` with their corresponding channels.
+
+A recommended convention: calling ``transmit`` with a single qubit argument
+should return the channel for driving single qubit gates, and calling
+``receive`` with a single qubit argument should return the channel for reading
+out the state of that qubit.
+
+.. code-block:: none
+
+   // Get the default input channel for qubit 0
+   transmit(%0)
+   // Get the default output channel for qubit 1
+   receive(%1)
+   // Get an input channel named "measure" for qubit 1
+   // Could be used to specify the measurement stimulus channel
+   transmit("measure", %1)
+   // Get an output channel named "cr1" for qubit 0
+   receive("cr1", %0)
+   // Get the default output channel for qubits 0 and 1
+   receive(%0, %1)
+
+Frames
+------
+
+It turns out to be quite useful to keep track of a set of carrier signals
+throughout the execution of a program. These carrier signals are called "frames"
+and are defined by :math:$A*e^{i\left(2\pi f t + \theta\right)}$, where `f` is
+frequency, `theta` is phase, and `A` is a scaling factor.
+
+Frames are attached to channels by a 1:1 correspondence. Every frame must 
+specify a frequency before any pulses can be played on that channel. The special 
+``prelude`` defcal is a good place to place these instructions.
+
+.. code-block:: none
+
+   defcal prelude {
+      // Required: set the frequency to 5GHz
+      initframe transmit(%1), 5e9;
+      // Optional: set the starting phase (otherwise it would start at 0)
+      set_phase transmit(%1), pi;
+   }
+
+Frames track the set of (frequency, phase, scale) throughout the program and 
+are manipulated using the following `set_` and `shift_` commands. Then when
+a pulse is played the carrier signal of the frame is mixed with the pulse itself
+to create the final signal which will be applied on the channel.
+
+Frequency has no default since it must be set at the beginning. Phase defaults
+to 0 and is a value of type ``angle``. Scale defaults to 1.0 and is a value of
+type ``float``. The exact precision of these three parameters is hardware
+specific.
+
+frameof command
+~~~~~~~~~~~~~~~
+
+The ``frame`` type is a compile-time type represnting a frame. It is useful
+when playing a pulse using the frame of a different channel.
+
+A value of type ``frame`` is retrieved by calling the ``frameof`` command
+and passing in a channel.
+
+Frame manipulation
+~~~~~~~~~~~~~~~~~~
+
+The (frequency, phase, scale) of a frame can be manipulated throughout program
+execution with the following instructions. After a call to one of these
+instructions, future references to that frame will have the new value.
+
+All of the following instructions have the same form. The first parameter can 
+either be a ``frame`` type or a ``channel`` type (in which case the 
+corresponding frame will be used). The second parameter is the value to shift
+or set. It must be the appropriate type, ``float`` for frequency, ``angle`` for
+phase, and ``float`` for scale. Again, the exact precision of these calculations
+is hardware specific.
+
+The relevant instructions are:
+
+- ``set_frequency`` and ``shift_frequency``
+- ``set_phase`` and ``shift_phase``
+- ``set_scale`` and ``shift_scale``
+
+Here's an example of using ``shift_scale`` to calibrate an ``rz`` gate:
+
+.. code-block:: none
+
+   // Shift phase of qubit 0's frame by pi/4, eg. an rz gate with angle -pi/4
+   shift_phase transmit(%0), pi/4;
+
+   // Equivalent, but more verbose
+   frame fr = frameof(transmit(%0));
+   shift_phase fr, pi/4;
+
+   // Define a calibration for the rz gate on all physical qubits
+   defcal rz(angle[20]:theta) %q {
+     shift_phase transmit(%q), -theta;
+   }
+
+Here's an example qubit spectroscopy experiment.
+
+.. code-block:: none
+
+   qubit q;
+
+   const shots = 1000;
+   const dfreq = 1e6; # 1MHz per point
+   const points = 50; # Sweep over 50MHz
+
+   complex[32] iq, average;
+   complex[32] output[points];
+
+   for p in [0 : points-1] {
+     average = 0;
+     for i in [0 : shots-1] {
+       // Assumes suitable calibrations for reset, x, and measure_iq
+       reset q;
+       x q;
+       measure_iq q -> iq;
+
+       average = (average * i + iq) / (i + 1);
+     }
+     shift q;
+     output[p] = average;
+   }
+
+   defcal prelude {
+      set_frequency transmit(%q) 5e9;
+   }
+
+   defcal shift %q {
+     shift_frequency transmit(%q) dfreq;
+   }
+
+Pulses
+------
+
+Pulses have two representations:
+
+- An array of complex samples which define the points for the pulse envelope
+- A``pulse`` type, which describes an abstract mathematical function
+  representing a pulse. This will later be materialized into a list of complex
+  samples, either by the compiler or the hardware using the parameters provided
+  to the pulse template.
+
+A value of type ``pulse`` is retrieved by calling one of the built-in pulse
+template functions. Note that each of these functions takes a type ``length``
+as a first argument, since pulses need to have a definite length. Using the
+hardware dependent ``dt`` unit is recommended, since the compiler may need to
+down-sample a higher precision pulse to physically realize it.
+
+.. code-block:: none
+
+   // amp is pulse amplitude at center
+   // center is the mean of pulse
+   // sigma is the standard deviation of pulse
+   gaussian(length:l, complex[float[32]]:amp, length:center, length:sigma)
+
+   // amp is pulse amplitude at center
+   // center is the mean of pulse
+   // sigma is the standard deviation of pulse
+   sech(length:l, complex[float[32]]:amp, length:center, length:sigma)
+
+   // amp is pulse amplitude at center
+   // center is the mean of pulse
+   // square_width is the width of the square pulse component
+   // sigma is the standard deviation of pulse
+   gaussian_square(length:l, complex[float[32]]:amp, length:center, length:square_width, length:sigma)
+
+   // amp is pulse amplitude at center
+   // center is the mean of pulse
+   // sigma is the standard deviation of pulse
+   // beta is the Y correction amplitude, see the DRAG paper
+   drag(length:l, complex[float[32]]:amp, length:center, length:sigma, float[32]:beta)
+
+   // Define a boxcar (aka. constant) pulse of length l
+   boxcar(l:length)
+
+Play instruction
+~~~~~~~~~~~~~~~~
+
+Pulses are scheduled using the ``play`` instruction. Play instructions have two
+required parameters:
+
+- the channel on which to play the pulse
+- a value of type ``pulse`` representing the pulse envelope
+
+Optionally the ``play`` instruction can also take a third parameter: the frame
+to use for the pulse. If the frame is not specified then the corresponding frame
+for the channel will be used instead, as if ``frameof`` was called for that
+channel.
+
+.. code-block:: none
+
+   // Play a 3 sample pulse on qubit 0
+   play transmit(%0), [1+0*j, 0+1*j, 1/sqrt(2)+1/sqrt(2)*j];
+
+   // Play a gaussian on qubit 0 using qubit 1's frame
+   frame f1 = frameof(transmit(%1));
+   play transmit(%0), gaussian(...), f1;
+
+Capture Instruction
+-------------------
+
+Acquisition is scheduled by a ``capture`` instruction. This is a special
+``kernel`` function which is specified by a hardware vendor. The measurement
+process is difficult to describe generically due to the wide variety of
+hardware and measurement methods.
+
+The only required parameter is an ``channel``, ie. a channel returned from
+calling the ``receive`` function.
+
+The following are possible parameters that might be included:
+
+- A "duration" of type ``length``, if it cannot be inferred from other parameters
+- A "filter", which is dot product-ed with the measured IQ the distill the
+  result into a single IQ value
+- A "tag", which could be used to identify which branch of an if statement was
+  traversed
+
+Again it is up to the hardware vendor to determine the parameters and write a
+kernel definition at the top-level, such as:
+
+.. code-block:: none
+
+   // Minimum requirement
+   kernel capture(channel output) -> complex[32];
+
+   // A capture command with more features
+   kernel capture(channel output, pulse filter) -> complex[32];
+
+The return type of a ``capture`` command varies. It could be a raw trace, ie a
+list of samples taken over a short period of time. It could be some averaged IQ
+value. It could be a classified bit. Or it could even have no return value,
+pushing the results into some buffer which is then accessed outside the program.
+
+Timing
+------
+
+Each channel maintains its own "clock". When a pulse is played the clock for 
+that channel advances by the length of the pulse. The same is true
+for output channels. Pulses on a single channel cannot be played simultaneously,
+although pulses on multiple channels for the same qubit can.
+
+For channels, everything behaves analogous to qubits in the 
+`Delays <delays.html>`_ section of this specification. There are however some
+small differences.
+
+The ``delay`` instruction may take a channel instead of a qubit. If a ``delay``
+instruction is applied to the qubit, this is the same as applying the delay to
+all channels on the qubit simultaneously.
+
+The ``barrier`` instruction on a qubit implies a barrier on all channels defined
+for that qubit. A barrier instruction will advance the clocks on all channels of
+the qubit to the channel with the highest clock.
+
+``defcal`` blocks have an implicit barrier on every qubit argument, meaning
+that clocks are guaranteed to be aligned at the start of the ``defcal`` block.
+These blocks also need to have a well-defined length, similar to the ``boxas``
+block.
+
+.. code-block:: none
+
+   pulse p = ...;
+
+   defcal simultaneous_pulsed_gate %0 {
+     play transmit("channel1", %0), p;
+     delay[20dt] transmit("channel2", %0);
+     // Starts the 100dt pulse 20dt into "channel1" already playing it
+     play transmit("channel1", %0), pulse;
+   }
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -82,7 +82,7 @@ specify a frequency before any pulses can be played on that channel. The special
 
    defcal prelude {
       // Required: set the frequency to 5GHz
-      initframe transmit(%1), 5e9;
+      set_frequency transmit(%1), 5e9;
       // Optional: set the starting phase (otherwise it would start at 0)
       set_phase transmit(%1), pi;
    }
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -1,27 +1,42 @@
 OpenPulse Grammar
 =================
 
-In addition to OpenQASM instructions, ``defcal`` blocks may contain OpenPulse 
-instructions. Certain OpenPulse instructions are further restricted to only
-appearing inside of a ``defcal`` block.
+OpenQASM allows users to provide the target system's implementation of quantum operations
+with ``defcal`` blocks . Calibration grammars are open to extension for system implementors. In
+this document, we outline one such grammar, OpenPulse, which maybe selected within a supporting
+compiler through the declaration defcalgrammar "openpulse" <VERSION>.
 
-These instructions are motivated by the original OpenPulse specification which
-was defined as a JSON wire-format for pulse-level quantum programs in the paper
-`Qiskit Backend Specifications for OpenQASM and OpenPulse Experiment <https://arxiv.org/abs/1809.03452>`_. 
-The text format described here has several advantages over the equivalent JSON 
-format:
+This grammar is motivated by the original OpenPulse specification, a JSON wire-format for
+pulse-level quantum programs defined in the paper `Qiskit Backend Specifications for OpenQASM
+and OpenPulse Experiment`_:cite:`mckay2018`. The textual format described
+here has several advantages over the original JSON format:
 
-- It is more readable
-- Absolute time is handled through built-in timekeeping instructions
-- Gates and classical instructions can be mixed in with the pulses to create far richer calibrations
-- Pulse definitions are tied to circuit instructions rather than circuit programs
-- Richer handling of frames
+- Improved readability
+- Pulse timing is based on instruction ordering and works with programs containing branching
+  control flow
+- Reusable gate calibrations enable more succinct calibration descriptions.
+- Pulse definitions are declared as a calibration for individual circuit instructions attached to
+  physical qubits enabling the microcoding of gate level operations.
+- Richer ability to compose complex pulses through natural DSP-like operations.
+- Clearly defined relationship between pulses, channels and the phases of compilation to
+  hardware resources.
+- Use of multiple frames on a single channel at the same time (see :ref:`Geometric gate`)
+
+
+Openpulse provides a flexible programming model that should extend to many quantum control schemes
+and hardware. At the core of the OpenPulse grammar are the concepts of ``frames``.
+``frames`` are used to schedule the playing of a ``waveform`` or capturing of data
+via a ``channel`` on the target device. A channel is a software abstraction which allows the
+programmer to be agnostic to complexities of the device's underlying pulse generation hardware. It
+is the responsibility of the target device's compiler to map ``frames`` to the applied channels in
+target hardware.
+
 
 Channels
 --------
 
 Channels map to a hardware resource, which can play pulses to manipulate a qubit
-or capture signal from the qubit to perform a measurement. There is a many-to-many
+or capture data from the qubit to perform a measurement. There is a many-to-many
 relationship between qubits and channels. One qubit may have multiple channels
 connecting to it. Pulses on different channels would have different physical
 interactions with that qubit. A channel may also have many qubits. For instance,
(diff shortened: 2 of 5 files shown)


delapuente:
@stevenheidel as you indicate, this is a proposal and it's fine to have it here. However, for driving the discussion, could you open an issue for it, please?

stevenheidel:
Sure, I opened #110

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+     shift_freq drive(%q) dfreq;
+   }
+
+Capture Instruction
+-------------------
+
+Acquisition is scheduled by a ``capture`` instruction.
+
+The ``capture`` instruction takes two parameters, a channel and the filter to
+apply to the returned signal. The length of the filter in dt will determine how
```
caryan, line 238: How does the user get at the underlying raw trace besides a basis of filters? It would be handy for calibration but more generally separating capture and integration would allow users to specify something different to a linear filter - e.g. an optimal Bayes or ML approach.

caryan (commented):
Looks great @stevenheidel.  I'd like to start to make a library of examples to exercise the grammar to better think through some use cases. Have you started any?

stevenheidel:
> Looks great @stevenheidel. I'd like to start to make a library of examples to exercise the grammar to better think through some use cases. Have you started any?

That is exactly the right next step. @karalekas volunteered to also try it out with some examples. I would recommend you both do so independently (along with anyone else who is interested). If you both independently arrive at similar-looking programs then that is a good sign for the grammar. If you both come up with similar suggestions then that is a good sign for the suggestion.

source/language/pulses.rst
```diff
@@ -72,16 +69,6 @@ the most specific definition found for a given operation. Thus, given,
 the operation ``rx(pi/2) %0`` would match to (3), ``rx(pi) %0`` would
 match (2), ``rx(pi/2) %1`` would match (1).
 
-Users specify the grammar used inside ``defcal`` blocks with a ``defcalgrammar "name"`` declaration, or by
```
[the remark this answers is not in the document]
stevenheidel: Why have the ability to specify a grammar? Supporting multiple grammars to specify the same thing means a marked increase in complexity.

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+two kinds of channels: input channels, which send signal into the system to
+drive the qubit, and output channels, which measure signal from the system to
+measure qubits.
+
+Each channel has a user-specified local oscillator frequency which is assumed to
+have been specified outside of the context of OpenQASM.
+
+A value of type ``channel`` can be retrieved with the ``drive`` (input channels) 
+and ``acquire`` (output channels) functions, a reference to the physical qubit, 
+and the name of the channel. If the name of the channel is omitted then the
```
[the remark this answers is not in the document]
stevenheidel: Two canonical channels are included in OpenPulse so I included them here too

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+The length of the pulse will be the length of the array multiplied by the unit
+``dt``, which specifies the sample rate.
+
+.. code-block:: none
+
+   // Play a 3 sample pulse on qubit 0's drive channel
+   play drive(%0), [1+0*j, 0+1*j, 1/sqrt(2)+1/sqrt(2)*j];
+
+Specifying a full list of samples for real-life pulses can be unwieldy, so we
+include several built-in pulse shape functions as well:
```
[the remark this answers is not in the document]
stevenheidel: Seems like a worthwhile design space to explore. This would require two things:
- a `shape` type that is not immediately turned into an array of samples on creation
- an addition of a length parameter on the `play` command which right now is implicit

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+
+   // Shift phase of qubit 0 by pi/4, eg. an rz gate with angle -pi/4
+   shift_phase drive(%0), pi/4;
+
+   // Define a calibration for the rz gate on all physical qubits
+   defcal rz(angle[20]:theta) %q {
+     shift_phase drive(%q), -theta;
+   }
+
+Shift Frequency Instruction
```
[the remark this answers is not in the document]
stevenheidel: Some thoughts on a bunch of related comments on frequency and phase:

When I wrote the Quil-T proposal at Rigetti I came up with this idea where frequency, phase, and scale were all implicitly tracked in the background for each channel. See here: https://github.com/rigetti/quil/pull/5/files#diff-eb8ffcda694c97f435a9132c5eb133ffc50a9005c3f98ddcf4079b2a0aef4774R34

This led to that table on line 63 where each of (frequency, phase, scale) behaved a little differently. In particular, set_frequency was required to be called before a pulse.

I still consider this to be an elegant solution despite needing that table. I also think the design space hasn't fully been explored there yet - it's definitely worth hearing some more ideas.

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+
+Capture Instruction
+-------------------
+
+Acquisition is scheduled by a ``capture`` instruction.
+
+The ``capture`` instruction takes two parameters, a channel and the filter to
+apply to the returned signal. The length of the filter in dt will determine how
+long the capture channel is open. The ``capture`` instruction returns the dot
+product of the measured IQ values and the filter. (Note that the filter is
```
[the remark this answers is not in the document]
stevenheidel: I'll update "other contexts" to what I really meant which was "some literature uses the term kernel for what we refer to as a filter, since kernel in OpenQASM is already used to refer to a custom classical function"

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+------
+
+Each channel maintains its own "clock". When a pulse is played the clock for 
+that channel advances by the number of samples in the pulse. The same is true
+for output channels based on the length of the capture filter. Pulses on a
+single channel cannot be played simultaneously, although pulses on multiple
+channels for the same qubit can.
+
+For channels, everything behaves analogous to qubits in the 
+`Delays <delays.html>`_ section of this specification. There are however some
```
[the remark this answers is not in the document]
stevenheidel: The only issue with negative delays is that you can't enforce the restriction that only 1 pulse is playing on a channel at one time, so that would have to be pushed to a runtime error.

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+   }
+
+Shift Frequency Instruction
+---------------------------
+
+A ``shift_freq`` instruction schedules a frequency advance for all the following
+pulses on that channel. This is useful for defining spectroscopy experiments.
+
+The ``shift_freq`` instruction takes two parameters, a channel and the
+requested frequency change of type float in units of GHz.
```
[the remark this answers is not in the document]
stevenheidel: I took the GHz units from OpenPulse but I have no personal opinion, Hz seems reasonable

source/language/pulses.rst
stevenheidel: It's clear that getting raw traces is a useful calibration to run, the question is whether to include it in the language? OpenPulse has this concept of measurement levels (0 for raw trace, 1 for IQ, 2 for classified value). This spec supports 1 and 2: 1 through the `capture` instruction and 2 through writing some classical function to do the discrimination.

stevenheidel:
@caryan - thanks for the feedback! Lots of comments related to frequency, phase, or scale so clearly more work to do there.

For the record, here are the other things I've been pondering since writing this:
- Do we need a `frequency` type with units like the `length` type? This would solve the Hz GHz problem
- Similarly, is there a need for a shorthand IQ type? (a complex number with a hardware-specified amount of precision)
- Timing and length of pulses is underspecified. There's maybe a sketch of some special syntax for defining the lengths of things in OpenQASM (see `rotary` function in delays.rst) but doesn't look like this has been fully specified yet.

source/language/pulses.rst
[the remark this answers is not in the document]
karalekas: afaict (from the existing examples at least) there is also a `measure` channel:  [defcal.qasm#L19](https://github.com/Qiskit/openqasm/blob/master/examples/defcal.qasm#L19)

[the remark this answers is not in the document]
stevenheidel: I renamed `measure` to `capture` since `measure` is already a keyword in openqasm

lcapelluto (commented):
We might consider switching the order of the arguments to match the OpenQasm syntax. `shift_phase(π), drive(%0);` The downside is that you'd have to define your waveforms first if you wanted your `play` instructions to be readable (i.e., `play([long list of samples]), drive(%0);` will be harder to read).

re: your ponderings, I'd say

 - Freq units: That seems like a cool idea. Normally I wouldn't prefer it, but like you say, OpenQasm has it already
 - You make a good point about the user not knowing the precision of the returned value 🤔 Leaving it as a particular type should work nicely since the compiler for the hardware backend should be able to substitute it for the proper precision -- what happens with arithmetic done on the return value? Would it get cast to `complex` anyway, or do all the other values get cast to the return type?
 - The instructions we have so far seem to me to have fully specified lengths in terms of dt. Are you referring to the lack of explicit alignment or something else?

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+classically-controlled stimulus fields used to interact with a qubit. There are
+two kinds of channels: input channels, which send signal into the system to
+drive the qubit, and output channels, which measure signal from the system to
+measure qubits.
+
+Each channel has a user-specified local oscillator frequency which is assumed to
+have been specified outside of the context of OpenQASM.
+
+A value of type ``channel`` can be retrieved with the ``drive`` (input channels) 
+and ``acquire`` (output channels) functions, a reference to the physical qubit,
```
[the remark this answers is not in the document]
psivaraj: Or indeed you might have an architecture with control built to address specific subsets of qubits (e.g. rows/columns of a neutral atom array).

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+   defcal shift %q {
+     shift_freq drive(%q) dfreq;
+   }
+
+Capture Instruction
+-------------------
+
+Acquisition is scheduled by a ``capture`` instruction.
+
+The ``capture`` instruction takes two parameters, a channel and the filter to
```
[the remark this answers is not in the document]
psivaraj: I think the concept of a filter is both general and useful (e.g. filtering the image data using summing a la' boxcar or even a matched filter). But tying it to `time` is too specific.

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+long the capture channel is open. The ``capture`` instruction returns the dot
+product of the measured IQ values and the filter. (Note that the filter is
+sometimes referred to as "kernel" in other contexts, but this is not related in
+any way to the ``kernel`` instruction in OpenQASM).
+
+.. code-block:: none
+    
+   complex[float[32]] filter = [1, 1, 1, 1, 1];
+   // Capture for 5 samples
+   iq = capture acquire(%0), filter;
```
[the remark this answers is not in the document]
psivaraj: If it's a stimulus, then why not just a regular `play drive(%0, "measure")`? I don't think it needs to be coupled to the `capture`.

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+--------
+
+Channels (sometimes referred to as "frames") describe the different forms of
+classically-controlled stimulus fields used to interact with a qubit. There are
+two kinds of channels: input channels, which send signal into the system to
+drive the qubit, and output channels, which measure signal from the system to
+measure qubits.
+
+Each channel has a user-specified local oscillator frequency which is assumed to
+have been specified outside of the context of OpenQASM.
```
psivaraj, line 108: Do we also expect to specify other properties like the sample rate for the `channel`? Especially useful if we do delayed resolve to samples [as per this discussion](https://github.com/Qiskit/openqasm/pull/109/files#r550954765).

source/language/pulses.rst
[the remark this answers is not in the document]
lcapelluto: Right, Steven renamed `drive(%0, "measure")` to `acquire`, so the first of those three lines is equivalent to that

source/language/pulses.rst
```diff
@@ -72,16 +69,6 @@ the most specific definition found for a given operation. Thus, given,
 the operation ``rx(pi/2) %0`` would match to (3), ``rx(pi) %0`` would
 match (2), ``rx(pi/2) %1`` would match (1).
 
-Users specify the grammar used inside ``defcal`` blocks with a ``defcalgrammar "name"`` declaration, or by
-an optional grammar string in a ``defcal`` definition, e.g.
-
-.. code-block:: c
-
-   defcalgrammar "openpulse";
```
lcapelluto, line 80: Could we follow the qasm grammar? A qasm file could be started with:
```
OPENQASM 3;
OPENPULSE 3;
```
https://qiskit.github.io/openqasm/grammar/index.html

lcapelluto:
I'm starting a defcal generating script so we could easily obtain defcal statements from Qiskit `Schedule`s

> > Looks great @stevenheidel. I'd like to start to make a library of examples to exercise the grammar to better think through some use cases. Have you started any?
> 
> That is exactly the right next step. @karalekas volunteered to also try it out with some examples. I would recommend you both do so independently (along with anyone else who is interested). If you both independently arrive at similar-looking programs then that is a good sign for the grammar. If you both come up with similar suggestions then that is a good sign for the suggestion.

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+   boxcar(l:length)
+
+Timing
+------
+
+Each channel maintains its own "clock". When a pulse is played the clock for 
+that channel advances by the number of samples in the pulse. The same is true
+for output channels based on the length of the capture filter. Pulses on a
+single channel cannot be played simultaneously, although pulses on multiple
+channels for the same qubit can.
```
taalexander, line 267: Sometimes it is necessary to apply crosstalk compensation pulses in our systems that effectively occupy the same hardware channel at once such as in the case of a geometric gate or stark tone. While these could be captured by constructing a complex signal envelope it may be nice to consider a more graceful way of expressing this. I am ok pushing this down the road.

taalexander (commented):
Thank you very much for making this contribution @stevenheidel, we're excited to build out the openpulse grammar. I think this is an excellent start.

Some general thoughts:
- We should split this grammar proposal into a separate document referenced from `pulses.rst`.
- It would be great to have a general "execution model" section that lays the groundwork - qubits, channels, pulses, frames, and instruction timing.
- Need to spend some time discussing qubit measurement, and output formats. There is a large tradespace here.
- It would be informative to write out a mock calibration for a target system (possibly with some interesting features) in terms of defcals as a working demo of the proposal. This could be a replacement for this [example](https://github.com/Qiskit/openqasm/blob/master/examples/defcal.qasm).

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+Channels
+--------
+
+Channels (sometimes referred to as "frames") describe the different forms of
+classically-controlled stimulus fields used to interact with a qubit. There are
+two kinds of channels: input channels, which send signal into the system to
+drive the qubit, and output channels, which measure signal from the system to
+measure qubits.
+
+Each channel has a user-specified local oscillator frequency which is assumed to
```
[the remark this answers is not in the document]
stevenheidel: Creative! I will sketch that out a bit and include it in the next update.

source/language/pulses.rst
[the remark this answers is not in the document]
stevenheidel: Sounds good, Hz it is

source/language/pulses.rst
stevenheidel: Is that a necessary constraint? Another version of the `measure` instruction (perhaps named differently) could be calibrated to return the "level 1" value at least

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
 corresponding ``gate`` definitions. However, if a user provides a ``defcal`` for a symbol
 without a corresponding ``gate``, then we treat such operations like the ``opaque`` gates
 of prior versions of OpenQASM.
+
+OpenPulse instructions
```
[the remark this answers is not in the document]
stevenheidel: Will do

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+format:
+
+- It is more readable
+- Absolute time is handled through built-in timekeeping instructions
+- Gates and classical instructions can be mixed in with the pulses to create far richer calibrations
+
+Channels
+--------
+
+Channels (sometimes referred to as "frames") describe the different forms of
```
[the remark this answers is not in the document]
stevenheidel: This is a great explanation, thank you! It ties in well with the (freq, phase, scale) tracking we discussed above and could clean up a few things. I will make the necessary revisions.

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+
+Specifying a full list of samples for real-life pulses can be unwieldy, so we
+include several built-in pulse shape functions as well:
+
+.. code-block:: none
+
+   // amp is pulse amplitude at center
+   // center is the mean of pulse
+   // sigma is the standard deviation of pulse
+   gaussian(length:l, complex[float[32]]:amp, length:center, length:sigma)
```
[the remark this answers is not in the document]
stevenheidel: That question plus your question on multiple sample rates are both going to be interesting challenges

source/language/pulses.rst
```diff
@@ -92,3 +79,217 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+.. code-block:: none
+
+   // Define a boxcar (aka. constant) filter of length l
+   boxcar(l:length)
+
+Timing
+------
+
+Each channel maintains its own "clock". When a pulse is played the clock for 
+that channel advances by the number of samples in the pulse. The same is true
```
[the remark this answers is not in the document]
stevenheidel: The challenge here as you mention above is in defining a defcal for `reset.` Maybe that means `reset` isn't a gate but rather a subroutine?

stevenheidel:
Alright I have pushed a new version which has hopefully addressed the above comments. It especially improves the treatment of channels and frames.

Changelog:
- Moved OpenPulse grammar to a separate page
- Added restrictions to the defcal blocks. They must be deterministic and have a well-defined, compiler-inferable length. I made one exception to the determinism requirement: reset calibrations can have an if statement with two branches of equal length. Open to hearing other solutions for reset calibrations.
- Added some comments on level 0/1/2 measurements - namely that you can define a gate and corresponding calibration that returns any type you like, including raw traces.
- I separated channels and frames and expanded the explanation of what they are and how they are related. I then brought over my 3-tuple from Quil-T for (frequency, phase, and scale). Also completed the set of `set_` and `shift_` commands.
- Frequency has no default value, so it is now specified in a `prelude` calibration which runs before any other instructions in the program. (Thanks to @taalexander for the creative suggestion)
- Channels can now have arbitrary names and reference multiple qubits.
- Introduced a `pulse` type for defining an abstract pulse whose actual evaluation will be deferred to either the compiler or hardware.
- Added a `frame` parameter to the `play` command to cover the example of playing a pulse on one channel using another's frame.
- Based on @caryan and @taalexander discussion, turned `capture` into a generic `kernel` function which returns an arbitrary classical value. This pushes the responsibility of making decisions on how capture works to whoever is implementing the hardware.

There's a few opportunities for bikeshedding:
- the names of the `transmit` and `receive` functions which are used to retrieve channels
- the names of the special `prelude` and `postlude` defcals
- names and parameters of the built-in pulses, how many to include

And there are a couple topics that need more thought:
- timing, samples rates, `dt`
- hardware precision of frequency, phase, and scale
- custom pulse functions

Looking forward to hearing more of your feedback!

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Frames
+------
+
+It turns out to be quite useful to keep track of a set of carrier signals
+throughout the execution of a program. These carrier signals are called "frames"
+and are defined by :math:$A*e^{i\left(2\pi f t + \theta\right)}$, where `f` is
+frequency, `theta` is phase, and `A` is a scaling factor.
+
+Frames are attached to channels by a 1:1 correspondence. Every frame must
```
blakejohnson, line 77: Is it 1:1? Clearly a channel can have multiple frames, but a frame can only have one channel.

stevenheidel: The `frameof` operation only really works if there is a 1:1 mapping between channels and frames. I suppose nothing prevents us from creating additional frames that are not attached to channels.

source/language/pulses.rst
taalexander: I agree let's explore in #102.

source/language/pulses.rst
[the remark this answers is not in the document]
taalexander: Let's move this discussion to #102.

source/language/openpulse.rst
blakejohnson: It is perhaps worth noting that the existing OpenPulse spec does not allow for multiple frames per channel, but I find this to be problematic and worth changing.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+specific.
+
+frameof command
+~~~~~~~~~~~~~~~
+
+The ``frame`` type is a compile-time type represnting a frame. It is useful
+when playing a pulse using the frame of a different channel.
+
+A value of type ``frame`` is retrieved by calling the ``frameof`` command
+and passing in a channel.
```
blakejohnson, line 107: Per my comment above about multiple frames per channel, if we go with that suggestion then `frameof` would need an additional argument to identify a frame.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+phase, and ``float`` for scale. Again, the exact precision of these calculations
+is hardware specific.
+
+The relevant instructions are:
+
+- ``set_frequency`` and ``shift_frequency``
+- ``set_phase`` and ``shift_phase``
+- ``set_scale`` and ``shift_scale``
+
+Here's an example of using ``shift_scale`` to calibrate an ``rz`` gate:
```
blakejohnson, line 129:
```suggestion
Here's an example of using ``shift_phase`` to calibrate an ``rz`` gate:
```

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+
+       average = (average * i + iq) / (i + 1);
+     }
+     shift q;
+     output[p] = average;
+   }
+
+   defcal prelude {
+      set_frequency transmit(%q) 5e9;
+   }
```
blakejohnson, line 174: Any thoughts on whether `prelude` should get called automatically, or should it appear explicitly in the program? i.e. immediately prior to line 158 you could have
```
prelude;
for p in [0: points-1] {
// ...
```

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+The only required parameter is an ``channel``, ie. a channel returned from
+calling the ``receive`` function.
+
+The following are possible parameters that might be included:
+
+- A "duration" of type ``length``, if it cannot be inferred from other parameters
+- A "filter", which is dot product-ed with the measured IQ the distill the
+  result into a single IQ value
+- A "tag", which could be used to identify which branch of an if statement was
+  traversed
```
blakejohnson, line 264: I'm struggling to imagine a scenario where the measurement operation would need this info. Did you have something particular in mind?

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+value. It could be a classified bit. Or it could even have no return value,
+pushing the results into some buffer which is then accessed outside the program.
+
+Timing
+------
+
+Each channel maintains its own "clock". When a pulse is played the clock for 
+that channel advances by the length of the pulse. The same is true
+for output channels. Pulses on a single channel cannot be played simultaneously,
+although pulses on multiple channels for the same qubit can.
```
blakejohnson, line 288: Is it the channel or the frame the maintains the clock? It seems like the latter would be better, because then I can express things like multiplexed measurement as instructions to separate frames that all share a common channel.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+all channels on the qubit simultaneously.
+
+The ``barrier`` instruction on a qubit implies a barrier on all channels defined
+for that qubit. A barrier instruction will advance the clocks on all channels of
+the qubit to the channel with the highest clock.
+
+``defcal`` blocks have an implicit barrier on every qubit argument, meaning
+that clocks are guaranteed to be aligned at the start of the ``defcal`` block.
+These blocks also need to have a well-defined length, similar to the ``boxas``
+block.
```
blakejohnson, line 305: How do barriers on qubits get converted into barriers on channels? I am again thinking about the case of multiplexed measurement where i want:
```
measure %0 -> c[0];
measure %1 -> c[1];
```
where qubits 0 and 1 are on a multiplexed feedline to still issue simultaneously.

stevenheidel: That's a great question. The solution described in the current diff says something like "a barrier on a qubit applies to all channels for that qubit", which fails to provide a way to do multiplexed measurement.

There might even be a bigger problem here: if you're doing multiplexed readout of multiple qubits at once, how do you communicate that back to the individual measure instructions?
ie. this is really doing something like `measure %0 -> c[0], %1 -> c[1];`

source/language/openpulse.rst
stevenheidel: No need to have it called automatically. Having it called explicitly would mean one less special case

source/language/openpulse.rst
blakejohnson: I'm not looking for a single instruction that does multiplexed measurement, but rather I want to make sure we build the grammar and semantics so that at the gate/operation-level I can issue independent `measure` instructions and have that be compatible with a pulse-level implementation where the signals share a channel because the physical implementation is multiplexed.

source/language/pulses.rst
```diff
@@ -16,10 +16,7 @@ instructions to the underlying microcoded
 :cite:`wilkesBestWayDesign1989` stimulus programs emitted by
 the controllers to implement each operation. In OpenQASM we expose
 access to this level of control with pulse-level definitions of gates
-and measurement with user-selectable pulse grammar. A future document
-will define a textualized representation of one such grammar, OpenPulse.
-Here we restrict ourselves to defining the necessary interfaces within
-OpenQASM to these pulse-level definitions of gates and measurement.
+and measurement using a `text representation of OpenPulse <openpulse.html>`_
```
blakejohnson, line 19: Having a user-selectable pulse grammar is important to allow systems with very different control requirements (e.g. trapped ions) to plug in an appropriate pulse description.

source/language/pulses.rst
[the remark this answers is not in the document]
blakejohnson: The idea was supposed to be that a compiler vendor could choose to support some particular pulse grammar or grammars. Just like you can embed various kinds of assembly language in a C program (x86, power, ARM, etc.).

source/language/pulses.rst
```diff
@@ -63,25 +64,15 @@ physical qubits with different parameter values, e.g.
...
 the operation ``rx(pi/2) %0`` would match to (3), ``rx(pi) %0`` would
 match (2), ``rx(pi/2) %1`` would match (1).
 
-Users specify the grammar used inside ``defcal`` blocks with a ``defcalgrammar "name"`` declaration, or by
-an optional grammar string in a ``defcal`` definition, e.g.
-
-.. code-block:: c
-
-   defcalgrammar "openpulse";
-   defcal "openpulse" measure %q -> bit { ... }
```
blakejohnson, line 81: I'm fine with dropping this version.

source/language/pulses.rst
```diff
@@ -92,3 +83,63 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The contents of ``defcal`` bodies are subject to some restrictions. This is
+similar to how boxed expressions work.
+
+- They must have a definite length, regardless of the parameters passed in or
+  the state of the system when called. This allows the compiler to properly
+  resolve ``lengthof(...)`` calls and allowing for optimizations.
+- No control flow is allowed within a ``defcal`` block (with one exception,
+  see below). This follows from the definite length requirement.
```
blakejohnson, line 97: We need a better way to express this constraint that doesn't feel like such a kludge. I don't have a good suggestion for improvement right now, though.

blakejohnson:
Thanks for contributing this, @stevenheidel. I've finished my first pass through it and will turn my attention now to your questions.

source/language/index.rst
```diff
@@ -21,4 +21,5 @@ Appendix `[app:semantics] <#app:semantics>`__ gives formal semantics.
    subroutines
    directives
    delays
-   pulses
\ No newline at end of file
+   pulses
+   openpulse
```
taalexander, line 25: Could this go within a grammar subsection within `pulses`?

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
+OpenPulse Grammar
+=================
+
+In addition to OpenQASM instructions, ``defcal`` blocks may contain OpenPulse 
+instructions. Or equivalently stated: OpenPulse instructions cannot appear
+outside of a ``defcal`` block.
```
taalexander, line 6: Was the grammar definition removed? I think this is important to start off with.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+- Gates and classical instructions can be mixed in with the pulses to create far richer calibrations
+- Pulse definitions are tied to circuit instructions rather than circuit programs
+- Richer handling of frames
+
+Channels
+--------
+
+Channels map to a hardware resource, which can play pulses to manipulate a qubit
+or capture signal from the qubit to perform a measurement. There is a many-to-many
+relationship between qubits and channels. One qubit may have multiple channels
```
taalexander, line 25: 👍

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+
+.. code-block:: none
+
+   // Get the default input channel for qubit 0
+   transmit(%0)
+   // Get the default output channel for qubit 1
+   receive(%1)
+   // Get an input channel named "measure" for qubit 1
+   // Could be used to specify the measurement stimulus channel
+   transmit("measure", %1)
```
taalexander, line 63: It might be better to swap the order here to more easily define the string as an optional extension.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+   // Get the default output channel for qubits 0 and 1
+   receive(%0, %1)
+
+Frames
+------
+
+It turns out to be quite useful to keep track of a set of carrier signals
+throughout the execution of a program. These carrier signals are called "frames"
+and are defined by :math:$A*e^{i\left(2\pi f t + \theta\right)}$, where `f` is
+frequency, `theta` is phase, and `A` is a scaling factor.
```
taalexander, line 75: Possibly cite virtual Z gate as the canonical motivation?

source/language/openpulse.rst
taalexander: I agree frame-per-channel was a mistake and necessitates having every channel be a "logical" overlay channel of which many would be mapped to the same hardware channel to support multiple frames. This is in turn is very inflexible and makes applying a pulse at the frame of another channel very difficult.

Solutions I see:
 - Frames are explicit values that can be manipulated/assigned to channels. This is the approach I took in the document circulated privately.
 - Channels are derivable from existing channels with new frame relationships. This leads to an explosion of "channels" and the line between hardware/software entities may become blurred.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Frequency has no default since it must be set at the beginning. Phase defaults
+to 0 and is a value of type ``angle``. Scale defaults to 1.0 and is a value of
+type ``float``. The exact precision of these three parameters is hardware
+specific.
+
+frameof command
+~~~~~~~~~~~~~~~
+
+The ``frame`` type is a compile-time type represnting a frame. It is useful
```
taalexander, line 103: If a frame can be later assigned to a channel I think this resolves the discussion above.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+   gaussian_square(length:l, complex[float[32]]:amp, length:center, length:square_width, length:sigma)
+
+   // amp is pulse amplitude at center
+   // center is the mean of pulse
+   // sigma is the standard deviation of pulse
+   // beta is the Y correction amplitude, see the DRAG paper
+   drag(length:l, complex[float[32]]:amp, length:center, length:sigma, float[32]:beta)
+
+   // Define a boxcar (aka. constant) pulse of length l
+   boxcar(l:length)
```
taalexander, line 222: I would argue `constant` may be more immediately recognizable but this may be my own bias.

source/language/openpulse.rst
taalexander: *edit* - I now see the ability to supply a frame below to `play`.

source/language/openpulse.rst
taalexander: I believe the rationale was if there is data we could return in real-time to the circuit level that the experimentalist wishes to later extract and debug offline, like the full sampled signal. The measurement in question could be captured.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+
+For channels, everything behaves analogous to qubits in the 
+`Delays <delays.html>`_ section of this specification. There are however some
+small differences.
+
+The ``delay`` instruction may take a channel instead of a qubit. If a ``delay``
+instruction is applied to the qubit, this is the same as applying the delay to
+all channels on the qubit simultaneously.
+
+The ``barrier`` instruction on a qubit implies a barrier on all channels defined
```
taalexander, line 298: Are barriers allowed on `channels`? A scenario I would have in mind is synchronizing a multiplexed measurement trigger reset while still allow single-qubit gates that don't have this dependency to not be blocked globally on the qubit. The tricky part is communicating these complex dependencies to the circuit-level scheduler.

source/language/pulses.rst
```diff
@@ -16,7 +16,7 @@ instructions to the underlying microcoded
 :cite:`wilkesBestWayDesign1989` stimulus programs emitted by
 the controllers to implement each operation. In OpenQASM we expose
 access to this level of control with pulse-level definitions of gates
-and measurement using a text representation of OpenPulse.
+and measurement using a `text representation of OpenPulse <openpulse.html>`_
```
taalexander, line 19:
```suggestion
and measurement using a `textual representation of OpenPulse <openpulse.html>`_ 
```

taalexander (commented):
Changes look great 👍 . Thanks, @stevenheidel. Some more comments attached.

source/language/openpulse.rst
lcapelluto: I think it's still in (or is under discussion and would be added back to) `source/language/pulses.rst`

source/language/openpulse.rst
caryan: I'm struggling a bit with why we need both channels and frames. The notion of channels that "map to hardware resources" seems to mix in implementation concerns and makes it feel like the channel is either a data converter or a sequencer.  Can't we get away with claiming that everything is just played on an abstract frame at this point?  For example, if a hardware vendor supplies multiple frames that map to the same data converter then it's up to them to support that with multiple sequencers and an adder tree, or to merge the waveforms in software or however they want to support it. Or is the notion that you have a more limited set of frames and cover the combinatorial crosstalk issue with the product of a channel and frame?

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+     output[p] = average;
+   }
+
+   defcal prelude {
+      set_frequency transmit(%q) 5e9;
+   }
+
+   defcal shift %q {
+     shift_frequency transmit(%q) dfreq;
+   }
```
karalekas, line 178: This doesn't seem right to me -- the point of a `defcal` is to "specify a microcoded implementation of a `gate`, `measure`, or `reset` instruction", not to define a new OpenQASM-level symbol that expands into OpenPulse instructions. That being said, what is the right way to do spectroscopy then? Stuffing the `shift_frequency` call in a `defcal` for `reset` or `measure` feels like the wrong solution, but the only other thing I can think of atm requires breaking your requirement that OpenPulse instructions exist only in `defcal` blocks.

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+corresponding frame will be used). The second parameter is the value to shift
+or set. It must be the appropriate type, ``float`` for frequency, ``angle`` for
+phase, and ``float`` for scale. Again, the exact precision of these calculations
+is hardware specific.
+
+The relevant instructions are:
+
+- ``set_frequency`` and ``shift_frequency``
+- ``set_phase`` and ``shift_phase``
+- ``set_scale`` and ``shift_scale``
```
karalekas, line 127: What about `swap_phases`?

karalekas (commented):
Summary:

- Frames vs. channels confusion. I feel like what I really care about when playing a pulse is (1) the physical qubits, (2) some label that allows me to do frame tracking and map down to physical channels, and (3) the waveform. To me, (1) + (2) = frame.

- Addition of the `swap_phase` operation on frames.

- Blocking vs nonblocking pulses. Are we timekeeping per frame or are we timekeeping per qubit?

- OpenPulse instructions outside of `defcal`s. How do we do spectroscopy? Can people write/run OpenPulse programs or do they have to run OpenQASM programs with `defcals` in them?

- It would be great to see more treatment of multi-qubit operations. I'll try and cook up some more examples in addition to the one below.

Here's an example defcal in which I try to address a couple of the points above (and propose a solution):

```
defcal cz %0 %1 {
   play (%0, %1), "cz", erfsquare();
   shift_phase (%0), "xy" -0.123;
   play (%0, %1), "cz", erfsquare();
   shift_phase (%0), "xy", -0.456;
   shift_phase (%1), "xy" -0.789;
}
```

 In this example, I'm implementing a CZ gate as a pair of flux pulses sandwiching a shift phase, followed by some single-qubit RZ corrections. In this example, I want the first `shift_phase` to come after the first `play`, and the remaining `shift_phase`s to come after the second. I'd also like to do this without specifying delays, which leads to my point about blocking vs nonblocking pulses & how we're keeping time. If I want to do per-qubit timekeeping and have `play`s block all physical qubits that they involve, it's natural to just refer to those physical qubits directly in the `play` command rather than defining a channel or frame elsewhere (which is what I did in the snippet). I can look at this program and immediately understand its "moment" structure (by moment I'm thinking of [`cirq.Moment`](https://github.com/quantumlib/Cirq/blob/master/cirq/ops/moment.py)), and I would still be able to do so even if we swapped out the `shift_phase`s for single-qubit operations that take a non-zero amount of time.

stevenheidel:
Thanks everyone. Next steps for me on this PR are to make some of the smaller changes that have been requested.

The frames discussion is best moved to an issue at this point to keep it organized. I'll summarize the discussion so far and open that In a bit.

source/language/openpulse.rst
taalexander: Does this seem like a rather specific case of a more common operation? What if the phase relationship required is not a direct swap but a linear combination? Why do I need to swap the phase between two channels rather than copy from one channel to another and store the target channels original phase in a temporary location?

source/language/openpulse.rst
taalexander: +1 this feels wrong, but I don't see a solution around this. Perhaps spectroscopy is an edge-case that we have to hack into the circuit model? A similar case worth exploring is how do we express a Rabi experiment?

source/language/openpulse.rst
```diff
@@ -0,0 +1,316 @@
...
+value. It could be a classified bit. Or it could even have no return value,
+pushing the results into some buffer which is then accessed outside the program.
+
+Timing
+------
+
+Each channel maintains its own "clock". When a pulse is played the clock for 
+that channel advances by the length of the pulse. The same is true
+for output channels. Pulses on a single channel cannot be played simultaneously,
+although pulses on multiple channels for the same qubit can.
```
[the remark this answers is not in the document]
taalexander: I believe blocking/ non-blocking behavior can be orchestrated via barriers that synchronize pulses across channels. At a first glance Blocking/Nonblocking seems less-powerful then barriers. For example if I had three channels for a qubit how would I instruct that two pulses occur at the same time while the third channel happens ASAP withoutdelays? With a barrier it is
```
// some pulses before
barrier c0, c1;
play c0 pulse1; play c1 pulse1; // start at the same time
play c2 pulse2; // start at earliest possible time.
```
With blocking/non-blocking (please correct me if I get this wrong).
```
delay c0 dur1; delay c1 dur2; // required to align
NONBLOCKING play c0 pulse1; NONBLOCKING play c1 pulse1; // start at the same time
NONBLOCKING play c2 pulse2; // start at earliest possible time.
```

source/language/openpulse.rst
taalexander: I am hardly tied to the concept of a frame. The problems that channels + frames are attempting to solve are scheduling constraints in hardware. 

If I have multiple frames per-physical channel at the openpulse level it is not clear if resources are shared at scheduling time (this is a philosophical choice if we wish to make the language-aware of this, internally we have a system like this and we refer to these as logical channels). I am ok with this as a lowering step in the compiler could be to map multiple "frames" to a single frame via sideband mixing (or error if constraints cannot be met) in software as you describe.

However, it is not possible to express hardware that is capable of supporting multiple togglable frames going to a single output DAC and this responsibility would have to fall to a further lower-level representation to express toggling frames on a single channel. 

In either representation (frames, or frames+channels) it is important that the frame exist as a persistent entity that accumulates states throughout the execution in real-time to enable mapping of input arguments to carrier phases. It is also necessary to be able to reference and manipulate the state of these carrier frames across channels. I believe both proposals contain this (eg., `playwfr`, `SWAP-PHASE`).

source/language/openpulse.rst
stevenheidel: Ya basically. I'm predicting that in the near term the accumulation of results will be done mostly as a side effect (ie. you call capture without assigning it to a variable) and the results just buffer somewhere outside the language. Then it would be useful to tag multiple measurements.

Or in the case of active reset: the `reset` statement doesn't return anything, but maybe you're interested in how many rounds of active reset you do on average.

In any case with the generic capture instruction if a hardware vendor doesn't want to implement this they don't need to.

source/language/openpulse.rst
stevenheidel: Having thought about this more it seems obvious to me now that frames should maintain the clock.

source/language/openpulse.rst
stevenheidel: +1, another good argument for implementing timing on frames rather than channels.

source/language/pulses.rst
stevenheidel: Will add back in

stevenheidel:
New version ready for review. I decided to try out a different approach to channels and frames. Before that, quickly running through the small changes:

- Added defcalgrammar back in, with a version number
- Changed `play` syntax from `play target, pulse` to `play(pulse) target` to better match gate and measurement statements
- Renamed `boxcar` to `constant`

### Frames and channels

This conversation convinced me to drop channels: https://github.com/Qiskit/openqasm/pull/109#pullrequestreview-562957613

Also, think about how these things are implemented in the control hardware. The phase and frequency are just memory locations. Therefore why not treat them like any other memory, rather than having specialized shift and set instructions?

The `frame` object described in the new version means shifting phase looks like this:
```
driveframe.phase += pi;
```
rather than this:
```
shift_phase pi;
```

This cleans up a ton of things:
- It drops at least 6 statements [shift, set] x [phase, scale, frequency] from the grammar
- It solves the "swap phase" and "linear combination of phases" discussion by Peter and Thomas above
- Since manipulating the frame now == manipulating classical memory, there is no need to keep these within `defcal` blocks. This makes the qubit spectroscopy experiment look way nicer.
- For the same reason, we can drop the magic 'prelude' gate.
- Timing becomes clearer, there's also no need for specialized nonblocking vs blocking instructions

I imagine the most controversial part of this will be mapping frames to channels. I'd argue that that is an implementation concern akin to static instrument settings. There's always going to be a section of hardware specific settings that live "outside" OpenPulse, my argument is that the frame->channel mapping belongs there as well. Another argument: it's better to start without channels in the grammar and add them in later than it is to go the other direction.

Looking forward to your feedback!

source/language/openpulse.rst
blakejohnson: I typically don't want to just store the original phase, though, but to allow it to continue accumulating at its frame frequency while I operate on another frame. `swap_phases` has an easily articulated use case in SWAP and iSWAP gates.

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+of quantum hardware provide the appropriate channel names and qubit mappings
+as configuration information to end users. Each channel may also have associated 
+static settings, such as local-oscillator frequencies, which do not vary 
+throughout program execution. Again it is expected that vendors of quantum 
+hardware provide a method for manipulating those static settings if appropriate.
+
+There are two kinds of channels: transmit channels (sending input to a quantum
+device) and receive channels (reading output from a quantum device).
+
+Interacting with channels directly is onerous and makes the pulse programs less
```
psivaraj, line 41: One problem I see with restricting events to frames is an assumption that you only wish to schedule events (e.g. `play` or `capture`) on something that tracks phase. There are instances in which I need deterministic scheduling outside this paradigm. Some examples include wanting to schedule a `capture` at a very specific time that maps to a camera or a single-photon detector acquisition, or schedule a `play` that tunes a magnetic field coil during a gate. Although one could argue that these would just have a `frequency` and `phase` of 0, it seems funny to have to over-specify such properties -- I would probably ask why I can't just use a `channel`.

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+portable. The OpenPulse defcalgrammar does not directly work with channels and
+instead works with abstract "frames", described below. The static settings above 
+should also include a mapping of these abstract frames to the correct channels.
```
psivaraj, line 44: One problem I see with restricting events to frames is an assumption that you only wish to schedule events (e.g. play or capture) on something that tracks phase. There are instances in which I need deterministic scheduling outside this paradigm. Some examples include wanting to schedule a capture at a very specific time that maps to a camera or a single-photon detector acquisition, or schedule a play that tunes a magnetic field coil during a gate. Although one could argue that these would just have a frequency and phase of 0, it seems funny to have to over-specify such properties -- I would probably ask why I can't just use a channel.

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+Pulses have two representations:
+
+- An array of complex samples which define the points for the pulse envelope
+- A``pulse`` type, which describes an abstract mathematical function
+  representing a pulse. This will later be materialized into a list of complex
+  samples, either by the compiler or the hardware using the parameters provided
+  to the pulse template.
+
+A value of type ``pulse`` is retrieved by calling one of the built-in pulse
+template functions. Note that each of these functions takes a type ``length``
```
psivaraj, line 203: Can a vendor support additional built-in template functions?

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+- a value of type ``pulse`` representing the pulse envelope
+- the frame to use for the pulse
+
+.. code-block:: none
+
+   // Play a 3 sample pulse on qubit 0's "drive" frame
+   play([1+0*j, 0+1*j, 1/sqrt(2)+1/sqrt(2)*j]) frameof("drive", %0);
+
+   // Play a gaussian on qubit 1's "drive" frame
+   frame f1 = frameof("drive", %1);
```
psivaraj, line 252: This would now need a `const` qualifier.

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+Again it is up to the hardware vendor to determine the parameters and write a
+kernel definition at the top-level, such as:
+
+.. code-block:: none
+
+   // Minimum requirement
+   kernel capture(frame output) -> complex[32];
+
+   // A capture command with more features
+   kernel capture(frame output, pulse filter) -> complex[32];
```
psivaraj, line 283: Maybe it's just me, but when I first look at this the terms `output` and `filter` look confusing because  my inclination is to assume they are qualifiers. I think it would be more instructive to :fire: those additional terms and add an explicit example.

source/language/pulses.rst
```diff
@@ -92,3 +94,30 @@ circuit. Most of the time symbols in the ``defcal`` table will also have
...
+
+- They must have a definite length, regardless of the parameters passed in or
+  the state of the system when called. This allows the compiler to properly
+  resolve ``lengthof(...)`` calls and allowing for optimizations.
+- No control flow is allowed within a ``defcal`` block (with one exception,
+  see below). This follows from the definite length requirement.
+
+The sole exception to the second rule is a ``reset`` gate. The ``defcal`` for a
+``reset`` gate is permitted to have a single if statement, provided each branch
+of the if statement has definite and equivalent length.
```
psivaraj, line 112: To play devil's advocate, why is it just restricted to a `reset` gate? Why not allow for any `defcal` block to have control-flow provided each branch has definite and equivalent length. Taking this one step further, for loops that have a compile-time resolvable `lengthof(...)` could also qualify. If the compiler is already resolving branches for a `reset`,  I don't see why it couldn't do the same for other control flow constructs.  I'm not sure if it would resolve @blakejohnson's `kludge` concern, but it captures the ultimate intention of definite length.

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+These blocks also need to have a well-defined length, similar to the ``boxas``
+block.
+
+.. code-block:: none
+
+   pulse p = ...; // some 100dt pulse
+
+   defcal simultaneous_pulsed_gate %0 {
+     play(p) frameof("drive0", %0);
+     delay[20dt] frameof("drive1", %0);
```
lcapelluto, line 314: I think this syntax matches the gate syntax, right? We don't allow other QASM gates within `defcal`s, do we? Perhaps we could keep the syntax for `delay` the same as the syntax for `play`? `play(arg) frame;` and `delay(arg) frame;`?

lcapelluto:
Are pragmas allowed?

source/language/openpulse.rst
stevenheidel: Relates to this issue: https://github.com/Qiskit/openqasm/issues/121

source/language/openpulse.rst
stevenheidel: More thought needs to go into these template functions and the type they return.

source/language/openpulse.rst
stevenheidel: Take a look at the latest diff and let me know how awkward that is to do now.

source/language/pulses.rst
blakejohnson: Agreed. We started discussing this particular issue over in #102. In an internal discussion last week I think we came up with a workable alternative that @levbishop was going to write down. But the gist was allowing `defcal` or `defchannel` for anything with a compile-time computable duration.

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
+OpenPulse Grammar
+=================
+
+In addition to OpenQASM instructions, ``defcal`` blocks may contain OpenPulse 
+instructions. Certain OpenPulse instructions are further restricted to only
```
zachschoenfeld33, line 5: I thought OpenPulse instructions can only occur within a `defcal`?

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+all future pulses on that frame.
+
+The ``frame`` type is a compile-type type representing a *reference* to a frame.
+The frame is composed of four parts:
+
+1. A frequency ``frequency`` of type ``float``. It is not initialized and must
+be set before any pulses can be played on that frame.
+2. A phase ``phase`` of type ``angle``. This is initialized to the value 0.
+3. An amplitude ``scale`` of type ``float``. This is initilized to the value 1.0.
+4. A time of type ``dt`` which is manipulated implicitly and cannot be modified
```
zachschoenfeld33, line 65: Should it be clarified that `dt` is the sample clock time? As done here: https://qiskit.github.io/openqasm/language/delays.html#length-and-stretch-types.

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+.. code-block:: none
+
+   qubit q;
+
+   const shots = 1000;
+   const start = 5e9; // 5 GHz
+   const end = 6e9; // 6 GHz
+   const points = 50;
+
+   complex[32] iq, average;
```
zachschoenfeld33, line 161: I don't think `complex` types have yet been defined in OpenQASM. Probably a good idea.

source/language/openpulse.rst
zachschoenfeld33: I think that's a good idea, as they may want their own custom functions

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+   const temp = frame1.phase;
+   frame1.phase = frame2.phase;
+   frame2.phase = temp;
+
+Pulses
+------
+
+Pulses have two representations:
+
+- An array of complex samples which define the points for the pulse envelope
```
zachschoenfeld33, line 196: It might be clearer to have a single `pulse` type, which is either an array of samples or the abstract mathematical function?

source/language/openpulse.rst
```diff
@@ -0,0 +1,317 @@
...
+
+Play instructions have two required parameters:
+
+- a value of type ``pulse`` representing the pulse envelope
+- the frame to use for the pulse
+
+.. code-block:: none
+
+   // Play a 3 sample pulse on qubit 0's "drive" frame
+   play([1+0*j, 0+1*j, 1/sqrt(2)+1/sqrt(2)*j]) frameof("drive", %0);
```
zachschoenfeld33, line 249: Seems a bit odd to me that the frame is not within the `play` parentheses. Was there any particular rationale for doing this?

source/language/pulses.rst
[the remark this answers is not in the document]
zachschoenfeld33: I do like the idea of not specifying a required calibration grammar. This allows the OpenQASM model to be valid on any qubit implementation. We can use `openpulse` as the default for microwave pulse specification, but I don't think it adds much extra complexity to say other options are allowed? They would of course need to write their own compiler to handle the varying specifics of the `defcal`.

source/language/pulses.rst
zachschoenfeld33: I'm not sure about the resolution, but I do think including a version number is important as the pulse grammar (or whatever calibration grammar is used) will likely evolve in time.

taalexander:
I've posted an overview of what a "signal" formalism could look like in #126, please have a look 😄

zachschoenfeld33:
Added a PR to allow complex types in OpenQASM: https://github.com/Qiskit/openqasm/pull/175. This will be needed in any pulse grammar

stevenheidel:
Most of this conversation is outdated, so I'm going to close this and open a new PR

PR #115, opened 2021-01-06
Clarify assignment for subroutines

zachschoenfeld33:
Subroutine/kernel assignment should be done by `=` or another assignment operator `+=, *=`, etc and not by `->`. Parentheses should be required for classical arguments to kernels/subroutines and only be dropped if there are no such arguments.

### DIFF
--- a/examples/gateteleport.qasm
+++ b/examples/gateteleport.qasm
@@ -25,7 +25,7 @@ rz(pi/4) a;
 cx q, a;
 
 // measure out the ancilla
-r = logical_meas;
+r = logical_meas a;
 
 // if we get a logical |1> then we need to apply a logical Z correction
 if (r == 1) z q;
--- a/examples/gateteleport.qasm
+++ b/examples/gateteleport.qasm
@@ -4,13 +4,13 @@ include "stdgates.inc";
 
 // declarations
 const n = 3;
-kernel vote bit[n] -> bit;
+kernel vote(bit[n]) -> bit;
 
 def logical_meas qubit[3]:d -> bit {
     bit c[3];
     bit r;
     measure d -> c;
-    vote c -> r;
+    r = vote(c);
     return r;
 }
 
@@ -25,7 +25,7 @@ rz(pi/4) a;
 cx q, a;
 
 // measure out the ancilla
-logical_meas a -> r;
+r = logical_meas;
 
 // if we get a logical |1> then we need to apply a logical Z correction
 if (r == 1) z q;
--- a/examples/qec.qasm
+++ b/examples/qec.qasm
@@ -7,11 +7,11 @@ qubit a[2];
 bit c[3];
 bit syn[2];
 
-def syndrome qubit[3]:d, qubit[2]:a -> bit[2] { 
+def syndrome qubit[3]:d, qubit[2]:a -> bit[2] {
   bit[2] b;
-  cx d[0], a[0]; 
-  cx d[1], a[0]; 
-  cx d[1], a[1]; 
+  cx d[0], a[0];
+  cx d[1], a[0];
+  cx d[1], a[1];
   cx d[2], a[1];
   measure a -> b;
   return b;
@@ -21,7 +21,7 @@ reset a;
 x q[0]; // insert an error
 barrier q;
 syn = syndrome q, a;
-// also valid: syndrome q, a -> syn;
+
 if(int(syn)==1) x q[0];
 if(int(syn)==2) x q[2];
 if(int(syn)==3) x q[1];
--- a/examples/scqec.qasm
+++ b/examples/scqec.qasm
@@ -14,9 +14,9 @@ const n = d^2;       // number of code qubits
 
 uint[32] failures;  // number of observed failures
 
-kernel zfirst creg[n - 1], int, int;
-kernel send creg[n -1 ], int, int, int;
-kernel zlast creg[n], int, int -> bit;
+kernel zfirst(creg[n - 1], int, int);
+kernel send(creg[n -1 ], int, int, int);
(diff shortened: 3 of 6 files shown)


stevenheidel:
Out of curiosity: why have two syntaxes that mean the same thing?

taalexander:
I would prefer to remove the `->` syntax for return values and keep `=` only. Originally `->` was kept for measurement assignment for OpenQASM2.0 backwards compatibility but we should not have two ways of doing the same thing. `->` is reserved for signature declarations.

Are you ok with this @awcross1?

jaygambetta:
I actually like the arrow but I agree we need to have one. My preference for this is why change I dont see enough reasons

taalexander:
I prefer `=` as it is more common (c/python/...). Most modern languages have assignments on the LHS and this can be interpreted as `let x = y`. This approach also reserves `->` for the signature definition.

jaygambetta:
Ok,  I think we go with only one and if we go with '=' and either have a note for backward somewhere that only measurement supports '->' (not in the main text everywhere) or we remove. Otherwise we are making it complicated for no good reason, I would prefer remove if we go with '=' so I guess I support you @taalexander.

stevenheidel:
Or use '->' solely for measurement operations, to indicate that a quantum -> classical operation has occurred.

Definitely agree though that the results of kernels and subroutines should use '=' which is much more common for assignments.

taalexander:
> Or use '->' solely for measurement operations, to indicate that a quantum -> classical operation has occurred.

This is an interesting concept. The interface between the quantum and classical computation *is* special and perhaps it is useful to have a special notation to signify this. I think I would still prefer LHS assignment `result <- measure q;` for consistency when reading. This is also related to #102, WRT to output from defcals.

zachschoenfeld33:
Based on this, I will only allow `=` for assignment in the ANTLR grammar, except in the case of `measure` where we can continue to have the `->` for backwards compatibility.

I don't think a special notation for quantum to classical is necessary. Also some subroutines contain this bridge while others will not and we shouldn't use two different notations for subroutine assignment.

zachschoenfeld33:
PR has been updated

examples/gateteleport.qasm
```diff
@@ -25,7 +25,7 @@ rz(pi/4) a;
 cx q, a;
 
 // measure out the ancilla
-logical_meas a -> r;
+r = logical_meas;
```
taalexander, line 28: Qubit and parentheses are missing here.

examples/qec.qasm
```diff
@@ -7,11 +7,11 @@ qubit a[2];
 bit c[3];
 bit syn[2];
 
-def syndrome qubit[3]:d, qubit[2]:a -> bit[2] { 
+def syndrome qubit[3]:d, qubit[2]:a -> bit[2] {
```
taalexander, line 10: parantheses missing

examples/qec.qasm
```diff
@@ -21,7 +21,7 @@ reset a;
 x q[0]; // insert an error
 barrier q;
 syn = syndrome q, a;
```
taalexander, line 23: missing on call here.

examples/scqec.qasm
```diff
@@ -83,12 +83,12 @@ for shot in [1: shots] {
 
   // Initialize
   reset data;
-  cycle data, ancilla -> layer;
+  layer = cycle data, ancilla;
```
taalexander, line 86: parentheses missing here and below.

taalexander (changes requested):

examples/qec.qasm
stevenheidel: Do qubits go inside or outside the parens on subroutines?

taalexander: This is a good point and I am incorrect according to the current specification. I know we had previously had a discussion on unifying classical and quantum arguments within parentheses in subroutines and I thought the outcome had been in favour but the document doesn't reflect this. 

I am no longer sure how I feel on this subject and I will resolve these comments.

taalexander: I've spent some time thinking about this today after reading some feedback from @ndebeaudrap and I am no longer opposed to leaving qubits out of brackets in the spirit of backward compatibility with OpenQASM2. It does make differentiating between gates and subroutines more difficult.

zachschoenfeld33: I hadn't replied yet because of the potential ambiguity and wanted to see how antlr might parse it (which I haven't gotten to do yet). I feel it is reasonable to leave qubits out of parentheses but require classical arguments to be in the parentheses. If there are no classical arguments we can ignore the parenthesis. It is true in this case that you can't distinguish between kernel, subroutine or gate calls. However, if you put everything in parentheses for subroutines, you would still have ambiguity on whether it is a subroutine or a kernel. 

I don't think you necessarily need to be clear which is which at call time. This is similar to not knowing whether a gate has a `defcal` or a unitary `gate` implementation.

taalexander: I agree with the first paragraph but 

> I don't think you necessarily need to be clear which is which at call time. This is similar to not knowing whether a gate has a defcal or a unitary gate implementation.

Defcals are never called, they are a root level gate implementation passed to the compiler. Calls to defcals are indirectly performed through gates, and linked by the compiler. Only gates/measure/reset (operations) and subroutines/kernels are called.

taalexander: Resolving, the current spec is correct.

zachschoenfeld33:
Updated

examples/gateteleport.qasm
zachschoenfeld33: Fixed.

Issue #129, opened 2021-01-20
Feedback on spec


Travis-S-IBM:
In reading through the specification today, some items stood out as benefiting from clarification or editing. Sharing here.

## [Types and Casting](https://qiskit.github.io/openqasm/language/types.html)

* The text implicitly introduces the idea of "program qubits" via juxtaposition with "physical qubits". Presumably this distinction is conceptually important for the grammar and specification. To that end, I'd suggest a paragraph on program qubits, their properties, and any available information on mapping from program to physical qubits.

Separately, is the intention of this specification to introduce "program qubits" as a standard terminology into the circuit compilation lexicon? If so, then it's all the more important to call it out explicitly.

* There's no example on this page of how to use the `stretch` type. While there are some examples in the [Circuit timing page](https://qiskit.github.io/openqasm/language/delays.html), there should be an example here.

## [Gates](https://qiskit.github.io/openqasm/language/gates.html)

* The page introduces the idea of "quantum instruction" without definition. The [Built in quantum instructions page](https://qiskit.github.io/openqasm/language/insts.html) also doesn't really define them. Would there be a straightforward explanation that could be added?

* The example for a CX gate uses \oplus, not \otimes. Is that intentional? Separately, this example also introduces a different bit ordering than is commonly used; it should be called out more clearly.

* It's unclear to me why the example below (taken from the page) is the case. In particular, it seems `qr0` is a 1-qubit register, while `qr2` and `qr3` are 2-qubit, and `qr2` is 3-qubit. So wouldn't both applications of g be incorrect?
```
gate g qb0, qb1, qb2, qb3
{
  // body
}
qubit qr0[1];
qubit qr1[2];
qubit qr2[3];
qubit qr3[2];
g qr0[0], qr1, qr2[0], qr3; // ok
g qr0[0], qr2, qr1[0], qr3; // error!
```

## [Built-in quantum instructions](https://qiskit.github.io/openqasm/language/insts.html)

* The declaration of the classical bits in the example below is incorrect, no? Shouldn't it be `bits bit[10]`?
```
// Initialize, flip and measure an array of 10 qubits
qubit qubits[10];
bit[10] bits;
x qubits;
bits = measure qubits
```

## [Classical instructions](https://qiskit.github.io/openqasm/language/classical.html)

* In the example below, the loop goes over every _even_ integer, no?
```
// loop over every positive number from 0 to 20 using an indexset
for i in [0:2:20] {
   // do something
}
```

[...]

PR #132, opened 2021-01-28
Clarify use of "program qubits"

Travis-S-IBM:
This commit adjusts the first pagraph in the "Qubits" section
in two ways:
* Explicitly says that the qubits described in the paragraph are
"program qubits".
* Moves the sentence talking about quantum registers to later in
the paragraph, once the idea is introduced.

This is in response to feedback in #129.


taalexander:
@ajavadia and @awcross1 what do you think about the term "program qubits"?

zachschoenfeld33:
Want to follow up on this PR as it hasn't been touched for a while--I'm not sure we need this clarification?

zachschoenfeld33:
@levbishop based on the last TSC meeting, would you take a look at this and add a comment as needed?

source/language/types.rst
```diff
@@ -36,10 +36,10 @@ Qubits
 ~~~~~~
 
 There is a quantum bit (``qubit``) type that is interpreted as a reference to a
-two-level subsystem of a quantum state. Quantum registers are static
-arrays of qubits that cannot be dynamically resized. The statement ``qubit name;``
-declares a reference to a quantum bit. The statement ``qreg name[size];`` or ``qubit name[size];`` declares a
-quantum register with the given name identifier. The keyword ``qreg`` is included
+two-level subsystem of a quantum state. The statement ``qubit name;``
+declares a reference to a quantum bit. These qubits are referred to as "program qubits" (in distinction to "physical qubits" on actual hardware; see below.) The statement ``qreg name[size];`` or ``qubit name[size];`` declares a
```
levbishop, line 40:
```suggestion
declares a reference to a quantum bit. These qubits are referred to as "virtual qubits" (in distinction to "physical qubits" on actual hardware; see below.) The statement ``qreg name[size];`` or ``qubit name[size];`` declares a
```

The OpenQASM paper uses the term "virtual qubits" to distinguish from "physical qubits" and I like this better than "program qubits". Otherwise this looks good.

levbishop (changes requested):

source/language/types.rst
```diff
@@ -36,10 +36,10 @@ Qubits
 ~~~~~~
 
 There is a quantum bit (``qubit``) type that is interpreted as a reference to a
-two-level subsystem of a quantum state. Quantum registers are static
-arrays of qubits that cannot be dynamically resized. The statement ``qubit name;``
-declares a reference to a quantum bit. The statement ``qreg name[size];`` or ``qubit name[size];`` declares a
-quantum register with the given name identifier. The keyword ``qreg`` is included
+two-level subsystem of a quantum state. The statement ``qubit name;``
+declares a reference to a quantum bit. These qubits are referred to as "virtual qubits" (in distinction to "physical qubits" on actual hardware; see below.) The statement ``qreg name[size];`` or ``qubit name[size];`` declares a
```
blakejohnson, line 40: `[size]` should be attached to `qubit` rather than to the identifier:
```suggestion
declares a reference to a quantum bit. These qubits are referred to as "virtual qubits" (in distinction to "physical qubits" on actual hardware; see below.) The statement ``qreg name[size];`` or ``qubit[size] name;`` declares a
```

blakejohnson (dismissed):

taalexander (approved):

PR #135, opened 2021-01-29
add power operator

### LINKED ISSUE #123: definition of ^ operator
A subsection [Classical bits and registers](https://qiskit.github.io/openqasm/language/classical.html#classical-bits-and-registers) describes as

> Classical registers and bits support bitwise operators and the corresponding assignment operators between registers of the same size: and `&`, or `|`, xor `^`.

and a subsection [Integers](https://qiskit.github.io/openqasm/language/classical.html#integers) describes as

> Integer types support addition `+`, subtraction `-`, multiplication `*`, and integer division `/`; the corresponding assignments `+=`, `-=`, `*=`, and `/=`; as well as increment `++` and decrement `--`.

In a surface code in [examples](https://github.com/Qiskit/openqasm/blob/master/examples/scqec.qasm), `^` is used as "power" operation for integers, I guess.

```
const d = 3;         // code distance
const m = 10;        // number of syndrome measurement cycles
const shots = 1000;  // number of samples
const n = d^2;       // number of code qubits
```

If operator is like Python, `^` is XOR and `**` is POWER.

hhorii:
This PR add power `**` as an integer operation to resolve #123.


blakejohnson (dismissed):
LGTM

hhorii:
This PR should not be merged because the latest ANTLR file supports ordering of classical instructions. A new grammar for power instruction should be in the current expression hierarchy.

zachschoenfeld33:
@hhorii can we incorporate this into https://github.com/Qiskit/openqasm/pull/180/files?

hhorii:
I added `powerExpression` as the first precedence as [python](https://github.com/antlr/grammars-v4/blob/master/python/python3-py/Python3.g4#L244) does.

zachschoenfeld33:
@hhorii this makes sense. Can you add the power operation `**` to https://qiskit.github.io/openqasm/language/classical.html#evaluation-order? We also need to add `**` as allowed for each of the classical types (for instance as done in https://qiskit.github.io/openqasm/language/classical.html#integers).

We should add the assignment `**=` as well...

zachschoenfeld33:
Maybe a test can be added for the power operation too?

source/language/classical.rst
```diff
@@ -82,9 +82,9 @@ an index set, for example ``i in {0,3}`` returns ``true`` if i equals 0 or 3 and
 Integers
 ~~~~~~~~
 
-Integer types support addition ``+``, subtraction ``-``, multiplication ``*``, integer division [2]_ ``/``
-and modulo ``%``; the corresponding assignments ``+=``, ``-=``, ``*=``, ``/=`` and ``%=``; as well as
-increment ``++`` and decrement ``--``.
+Integer types support addition ``+``, subtraction ``-``, multiplication ``*``, integer division [2]_ ``/``,
+modulo ``%``, and power ``**``; the corresponding assignments ``+=``, ``-=``, ``*=``, ``/=``, ``%=`` and
```
zachschoenfeld33, line 86: Power should also be defined for floating types right?

source/language/classical.rst
```diff
@@ -122,6 +123,8 @@ OpenQASM evaluates expressions from left to right.
       +----------------------------------------+------------------------------------+
       | Operator                               | Operator Types                     |
       +----------------------------------------+------------------------------------+
+      | ``**``                                 | Power                              |
```
zachschoenfeld33, line 126: Should this be switched? `expressionTerminator` should be the highest precedence, then power.

zachschoenfeld33 (commented):
Few more comments. Also pls add "Power" to the expression hierarchy comment in the `qasm3.g4` file. I believe this is line 296 currently...

zachschoenfeld33 (approved):
LGTM. Nice work @hhorii!

blakejohnson (approved):

PR #150, opened 2021-02-22
Change physical qubit identifier

zachschoenfeld33:
Use `$0, $1, ..., $n` as physical qubit identifiers. This avoids a clash w/ the modulo operator `%`.

### DIFF
--- a/examples/defcal.qasm
+++ b/examples/defcal.qasm
@@ -1,36 +1,36 @@
 defcalgrammar "openpulse";
 
-defcal x %0 {
-   play drive(%0), gaussian(...);
+defcal x %q0 {
+   play drive(%q0), gaussian(...);
 }
 
-defcal x %1 {
-  play drive(%1), gaussian(...);
+defcal x %q1 {
+  play drive(%q1), gaussian(...);
 }
 
-defcal rz(angle[20]:theta) %q {
-  shift_phase drive(%q), -theta;
+defcal rz(angle[20]:theta) q {
+  shift_phase drive(q), -theta;
 }
 
-defcal measure %0 -> bit {
+defcal measure %q0 -> bit {
   complex[int[24]] iq;
   bit state;
   complex[int[12]] k0[1024] = [i0 + q0*j, i1 + q1*j, i2 + q2*j, ...];
-  play measure(%0), flat_top_gaussian(...);
-  iq = capture acquire(%0), 2048, kernel(k0);
+  play measure(%q0), flat_top_gaussian(...);
+  iq = capture acquire(%q0), 2048, kernel(k0);
   return threshold(iq, 1234);
 }
 
-defcal zx90_ix %0, %1 {
-  play drive(%0, "cr1"), flat_top_gaussian(...);  // uses a non-default
-                                                  // frame labeled "cr1"
+defcal zx90_ix %q0, %q1 {
+  play drive(%q0, "cr1"), flat_top_gaussian(...);  // uses a non-default
+                                                   // frame labeled "cr1"
 }
 
-defcal cx %0, %1 {
-  zx90_ix %0, %1;
-  x %0;
-  shift_phase drive(%0, "cr1");
-  zx90_ix %0, %1;
-  x %0;
-  x %1;
+defcal cx %q0, %q1 {
+  zx90_ix %q0, %q1;
+  x %q0;
+  shift_phase drive(%q0, "cr1");
+  zx90_ix %q0, %q1;
+  x %q0;
+  x %q1;
 }
--- a/examples/defcal.qasm
+++ b/examples/defcal.qasm
@@ -1,36 +1,36 @@
 defcalgrammar "openpulse";
 
-defcal x %q0 {
-   play drive(%q0), gaussian(...);
+defcal x $0 {
+   play drive($0), gaussian(...);
 }
 
-defcal x %q1 {
-  play drive(%q1), gaussian(...);
+defcal x $1 {
+  play drive($1), gaussian(...);
 }
 
-defcal rz(angle[20]:theta) q {
-  shift_phase drive(q), -theta;
+defcal rz(angle[20]:theta) $q {
+  shift_phase drive($q), -theta;
 }
 
-defcal measure %q0 -> bit {
+defcal measure $0 -> bit {
   complex[int[24]] iq;
   bit state;
   complex[int[12]] k0[1024] = [i0 + q0*j, i1 + q1*j, i2 + q2*j, ...];
-  play measure(%q0), flat_top_gaussian(...);
-  iq = capture acquire(%q0), 2048, kernel(k0);
+  play measure($0), flat_top_gaussian(...);
+  iq = capture acquire($0), 2048, kernel(k0);
   return threshold(iq, 1234);
 }
 
-defcal zx90_ix %q0, %q1 {
-  play drive(%q0, "cr1"), flat_top_gaussian(...);  // uses a non-default
+defcal zx90_ix $0, $1 {
+  play drive($0, "cr1"), flat_top_gaussian(...);  // uses a non-default
                                                    // frame labeled "cr1"
 }
 
-defcal cx %q0, %q1 {
-  zx90_ix %q0, %q1;
-  x %q0;
-  shift_phase drive(%q0, "cr1");
-  zx90_ix %q0, %q1;
-  x %q0;
-  x %q1;
+defcal cx $0, $1 {
+  zx90_ix $0, $1;
+  x $0;
+  shift_phase drive($0, "cr1");
+  zx90_ix $0, $1;
+  x $0;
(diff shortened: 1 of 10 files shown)


zachschoenfeld33:
Question here: Can we do 
`defcal rz(theta) q {}` so long as we call it as
```
rz(3.14) %q0;
rz(3.14) %q1;
```
or is the intent to define a new `defcal` per qubit. It is nice to reuse if the calibration pulse sequences are the same.

zachschoenfeld33:
@steleman

steleman:
```
defcal rz(theta) q {}
```
```
rz(3.14) %q0;
rz(3.14) %q1;
```
This is what the current version of the OpenQASM 3 Spec says:

> The entry point to such gate and measurement definitions is the defcal keyword analogous to the gate keyword, but where the defcal body specifies a pulse-level instruction sequence on **physical qubits**.

Doesn't that imply that the ```q``` argument in the defcal declaration should be, in fact, ```%q```?

The Spec also states the following:

> Some operations require further specialization on parameter values, so we also allow multiple declarations on the same physical qubits with different parameter values, e.g. [ ... ]

> Given multiple definitions of the same symbol, the compiler will match the most specific definition found for a given operation.

Which is really another way of saying **defcals can be overloaded based on parameters**, just like C++ Templates or C++ Template Functions.

For this particular example, it is unclear to me whether the explicit defcal call to ```rz(3.14) %q0;``` would even resolve, because the original defcal declaration takes a ```theta``` parameter, but the explicit defcal calls are passed ```pi``` (or a crude approximation thereof) as parameter.

So, I think we need quite a bit more details in the Spec about _defcal parameter overloading resolution rules_, just like in C++. One of the most important rules being the one that tells us _at which point do we give up because we can't find a close enough match?_

zachschoenfeld33:
In regards to `q` vs `%q`: I was thinking that you could define a `defcal` on a variable provided it is called on an actual physical qubit. So in the above examples, `q` becomes either `%q0` or `%q1`. You could make `defcal` variables have `%`, but the issue is suppose you want a `defcal` on three qubits.

If you do `defcal name %q0, %q1, %q2` that's now bound to a 0, 1, 2 ordering. But we might want flexibility to change the ordering. In this case it makes sense to define
```
defcal name q0, q1, q2
```

You could then call it as `name %q0, %q1, %q2` or `name %q2, %q0, %q1`, etc.

blakejohnson:
Ah, I guess we missed that in our discussion: `%alphanumeric` was reserved for a physical qubit variable (i.e. it can be bound to any physical qubit). This may suggest that dropping `%` as the modulus operator is the better option.

steleman:
On Tue, Feb 23, 2021 at 3:17 PM Blake Johnson <notifications@github.com>
wrote:

> Ah, I guess we missed that in our discussion: %alphanumeric was reserved
> for a physical qubit variable (i.e. it can be bound to any physical qubit).
> This may suggest that dropping % as the modulus operator is the better
> option.
>

We don't have to drop % as the modulo operator. It's just that it's more
difficult to write a parser that can make the distinction between the two
semantic contexts.

-- 
Stefan Teleman
stefan.teleman@gmail.com

stevenheidel:
> Ah, I guess we missed that in our discussion: `%alphanumeric` was reserved for a physical qubit variable (i.e. it can be bound to any physical qubit). This may suggest that dropping `%` as the modulus operator is the better option.

My vote would also be to drop % as the modulus operator.

steleman:
On Tue, Feb 23, 2021 at 5:48 PM Steven Heidel <notifications@github.com>
wrote:

>
> My vote would also be to drop % as the modulus operator.
>

What would be the technical justification for importing (and allowing) all
the operators from C except the % (modulo) operator?


-- 
Stefan Teleman
stefan.teleman@gmail.com

zachschoenfeld33:
A simple alternative would be choose something other than `%`. `$` (as was previously suggested) could work pretty well. `#` or `@` could work too.

source/language/pulses.rst
```diff
@@ -27,8 +27,8 @@ instruction sequence on *physical* qubits, e.g.
 
 .. code-block:: c
 
-   defcal rz(angle[20]:theta) %q { ... }
-   defcal measure %q -> bit { ... }
+   defcal rz(angle[20]:theta) $0 { ... }
+   defcal measure $0 -> bit { ... }
```
blakejohnson, line 31: This should still be in terms of a generic physical qubit, i.e. `$q`.

blakejohnson: Oh, i see you were introducing the generic version down below. Fine.

source/language/pulses.rst
```diff
@@ -38,47 +38,57 @@ interchangeable at the pulse level. Due to varying physical qubit
...
+QASM achieves this by prefixing qubit references with ``$`` to indicate
+a specific qubit on the device, e.g. ``$2`` would refer to physical
+qubit 2.
+
+One can define a `defcal` using an arbitrary `$` identifier, provided that gate is called using physical
+qubits. For instance, to define an equivalent `rz` calibration on qubits 0 and 1, we could write
+
+.. code-block:: c
+
+   defcal rz(angle[20]:theta) $q { ... }
```
blakejohnson, line 50: The example would be clearer with some whitespace between the `defcal` and the uses below. even inserting a line with a comment would help. i.e.
```
// we've defined 'rz' on arbitrary physical qubits, so we can already do:
...
```

zachschoenfeld33:
Have replaced `%` w/ `$` (thanks to @mbhealy)

examples/t1.qasm
```diff
@@ -15,26 +15,26 @@ kernel tabulate(int[32], int[32], int[32]);
 
 bit c0, c1;
 
-// define a gate calibration for an X gate on qubit 1
-defcal "openpulse" x %0 {
-   play drive(%0), gaussian(100, 30, 5);
+// define a gate calibration for an X gate on any qubit
+defcal "openpulse" x $q {
```
blakejohnson, line 19: While we are at it, we should remove the `"openpulse"` from the `defcal`, which has been dropped from the spec.

examples/defcal.qasm
```diff
@@ -1,36 +1,36 @@
...
+  iq = capture acquire($0), 2048, kernel(k0);
   return threshold(iq, 1234);
 }
 
-defcal zx90_ix %0, %1 {
-  play drive(%0, "cr1"), flat_top_gaussian(...);  // uses a non-default
-                                                  // frame labeled "cr1"
+defcal zx90_ix $0, $1 {
+  play drive($0, "cr1"), flat_top_gaussian(...);  // uses a non-default
+                                                   // frame labeled "cr1"
```
blakejohnson, line 26: Seems like the extra space mis-aligns the comments?

blakejohnson (changes requested):
Just a couple of minor fixes needed.

examples/t1.qasm
zachschoenfeld33: How does the user know then what pulse grammar to use? It still appears at the bottom of https://qiskit.github.io/openqasm/language/pulses.html as well?

examples/defcal.qasm
zachschoenfeld33: Addressed in https://github.com/Qiskit/openqasm/pull/150/commits/7ba67251679c80311bc500d812ce712fff8e1b24

source/language/pulses.rst
zachschoenfeld33: Addressed in https://github.com/Qiskit/openqasm/pull/150/commits/7ba67251679c80311bc500d812ce712fff8e1b24.

examples/t1.qasm
blakejohnson: With a `defcalgrammar` statement. I should fix that section you point out as well.

taalexander (commented):
I am a fan of the $ syntax. The only question I am still wondering is if we should go `$q<idx>` to and default to normal identifiers for wild-card qubits in defcals.

blakejohnson:
Just to follow up on the suggested change from the original proposal of `%qn` is that this doesn't actually solve the parsing problem for modular arithmetic against classical type identifies that start with 'q', e.g.
```
int[16] q7 = 7;
int[16] foo = (5 * 10) % q7;
```

Regarding @taalexander's suggestion of:
> I am still wondering is if we should go `$q<idx>`

I would just say that the `$` symbol is sufficient to distinguish it, so I would prefer not to introduce additional characters.

> default to normal identifiers for wild-card qubits in defcals.

I need to ponder this one a bit more.

examples/t1.qasm
zachschoenfeld33: Ok--I didn't realize that was the sole option now. Will update.

zachschoenfeld33: @blakejohnson I fixed that issue as well in this PR

blakejohnson (approved):

PR #154, opened 2021-02-25
Clarify what we mean by "C-like"

blakejohnson:
cc @stevenheidel


jwoehr:
"C-like" is possibly the most uninformative phrase in the document and has generated a troubling volume of discussion and effort.

PR #156, opened 2021-03-03
subroutineCall should always have parentheses

hhorii:
[Current grammar of `subroutineCall`](https://github.com/Qiskit/openqasm/blob/02e62e91509fad3ff480be91329f740ea85bbea6/source/grammar/qasm3.g4#L390-L392) allows not to have parenthesis.

```
// if have subroutine w/ out args, is ambiguous; may get matched as identifier
subroutineCall
    : Identifier ( LPAREN expressionList? RPAREN )? expressionList
    ;
```

As commented there, if no argument exists, parser cannot distinguish a subroutine call with a normal variable (`expressionTerminator->Identifier`).

```
expressionTerminator
    : Constant
    | MINUS? ( Integer | RealNumber )
    | Identifier
    | StringLiteral
    | timingTerminator
    ;
```

For example, this qasm file is parsed wrongly.
```
OPENQASM 3;
include "stdgates.inc";
const n = 1;
const m = n - 1;
```
Its tree of `const m = n - 1` is
```
  statement
    classicalDeclarationStatement
      constantDeclaration
        const
        equalsAssignmentList
          m
          equalsExpression
            =
            expression
              subroutineCall
                n
                expressionList
                  expression
                    -
                    expression
                      expressionTerminator
                        1
      ;
```

`n` is recognized as `subroutineCall` though it is a constant variable. 

To parse correctly, parentheses are always necessary for a subroutine call even if it does not have any arguments.

### DIFF
--- a/examples/vqe.qasm
+++ b/examples/vqe.qasm
@@ -73,7 +73,7 @@ def counts_for_term(bit[2*n]:spec) qubit[n]:q -> uint[prec] {
 // Estimate the expected energy
 def estimate_energy qubit[n]:q -> fixed[prec,prec] {
   fixed[prec, prec] energy;
-  uint[prec] npaulis = get_npaulis;
+  uint[prec] npaulis = get_npaulis();
   for t in [0:npaulis-1] {
     bit spec[2*n] = get_pauli(t);
     uint[prec] counts;
--- a/source/grammar/qasm3.g4
+++ b/source/grammar/qasm3.g4
@@ -388,7 +388,7 @@ subroutineBlock
 
 // if have subroutine w/ out args, is ambiguous; may get matched as identifier
 subroutineCall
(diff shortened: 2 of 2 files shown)


zachschoenfeld33:
@hhorii I agree w/ this. Currently, however, the spec does not require them. See https://qiskit.github.io/openqasm/language/subroutines.html (for instance, the `xcheck` method). 

It probably does make sense to have the `()` to recognize a subroutine/kernel. I suggest filing an issue and we can try to advocate for a modification of the spec.

hhorii:
I would like to close this PR because #180 includes this.

I believe that ANTLR will leave some ambiguity in its parsed tree. `Identifier` will be a variable, a kernel call, or a subroutine call. However, I do not think that`subroutineCall` should include a variable.

PR #159, opened 2021-03-05
Modify power call syntax

taalexander:
This PR updates the power syntax to be more consistent with `lengthof` and other function calls that operate on types.

Some rationale:

The existing `[]` syntax seems like special casing. I've changed to look like a function `ctrl @ pow(2) @ gate` which accepts a gate type and produces a gate type. `@` can be viewed as infix notation for calling transformation macros/functions like so `ctrl(pow(2, gate))`. It may be clearer that `pow(4, gate)` is `pow(2, pow(2, gate))`, rather than `pow[2](pow[2](gate))` which special cases the input parameter to `pow` unnecessarily.


zachschoenfeld33 (approved):
LGTM.

blakejohnson (approved):

PR #175, opened 2021-04-07
Add complex types 

zachschoenfeld33:
### Summary
This PR adds complex types to OpenQASM3. These are particularly important in pulse implementations, but may see other applications as well.

### Details
- Declare using the imaginary number ``im``. ``complex`` declaration can be any expression as for other types
- Example: ``complex[64] c = 2.1+5.1im`` which defines a 64-bit complex number composed of a real part (``float[64]`` 2.1) and an imaginary part (``float[64]`` 5.1).


zachschoenfeld33:
Starting to think using `a+b*I` (or another special imaginary constant char) could be better--would reduce potential confusion w/ arrays and make literals much simpler. Also just easier to reason through

steleman:
> Starting to think using `a+b*I`

I am not a fan of this. I do not see why **I** or **i** would have to be explicitly expressed.

complex[size] seems much more intuitive to me.

zachschoenfeld33:
> > Starting to think using `a+b*I`
> 
> I am not a fan of this. I do not see why **I** or **i** would have to be explicitly expressed.
> 
> complex[size] seems much more intuitive to me.

Yes that would be the type signature for declaration. But how do you do initialize? You can do 
```
complex[64] a;
```

But then how do you initialize the value. Many languages do it w/ some reserved value for imaginary. In python, it's `j`. In Julia, it's `im`. In C, it's `I`. I am suggesting doing 
```
complex[64] a = 1.5 + 2.0I
```
vs
```
complex[64] a = [1.5, 2.0]
```

steleman:
> > > Starting to think using `a+b*I`
> > 
> > 
> > I am not a fan of this. I do not see why **I** or **i** would have to be explicitly expressed.
> > complex[size] seems much more intuitive to me.
> 
> Yes that would be the type signature for declaration. But how do you do initialize? You can do
> 
> ```
> complex[64] a;
> ```
> 
> But then how do you initialize the value. Many languages do it w/ some reserved value for imaginary. In python, it's `j`. In Julia, it's `im`. In C, it's `I`. I am suggesting doing
```
> complex[64] a = 1.5 + 2.0I
> ```
> 
> vs
> 
> ```
> complex[64] a = [1.5, 2.0]
> ```
```

Then we have to allow construction initialization with all the arithmetic operators:

```
#include <stdio.h>
#include <complex.h>

int main()
{
  double complex z = 3.2 * 5.0 * I;
  (void) fprintf(stderr, "z: %.2f * %.2fi\n", creal(z), cimag(z));

  double complex c = 3.2 + 4.1 * I;
  (void) fprintf(stderr, "c: %.2f + %.2fi\n", creal(c), cimag(c));

  return 0;
}
```
gcc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9):
clang version 11.1.0:

z: 0.00 * 16.00i
c: 3.20 + 4.10i

jwoehr:
I like the `i` but why not just use `i` instead of `I `which is how we're taught from childhood.

zachschoenfeld33:
> I like the `i` but why not just use `i` instead of `I `which is how we're taught from childhood.

Well `i` is a far more common identifier. C uses `I`. But I'm open to different options. Yes @steleman I understand it could be more challenging for the compiler. Another option would be 2d tuples `(a,b)`

zachschoenfeld33:
Syntax updated to use ``im``

1ucian0:
It seems counter intuitive to me that `complex[64] c` is 128-bit long

blakejohnson:
Wasn't this discussed in a fairly recent live discussion? I thought we had settled on full type names inside the brackets, e.g. `complex[float[32]]` if you want a complex number composed of a 32-bit floating real part and 32-bit floating imaginary part.

zachschoenfeld33:
> Wasn't this discussed in a fairly recent live discussion? I thought we had settled on full type names inside the brackets, e.g. `complex[float[32]]` if you want a complex number composed of a 32-bit floating real part and 32-bit floating imaginary part.

@blakejohnson, yes you are right. I've fixed it.

blakejohnson:
@zachschoenfeld33 does the grammar also allow use of a machine-sized `int` or `float`, e.g. `complex[int]` or `complex[float]`?

zachschoenfeld33:
> @zachschoenfeld33 does the grammar also allow use of a machine-sized `int` or `float`, e.g. `complex[int]` or `complex[float]`?

Not yet. I believe we should add a fix for https://github.com/Qiskit/openqasm/issues/210 (ie support for machine-sized types) and then add this. I have marked this on #210

taalexander:
> @zachschoenfeld33 does the grammar also allow use of a machine-sized int or float, e.g. complex[int] or complex[float]?

Do we really require both? It seems like we are complicating the language when we do not have a clear and pressing use case.

steleman:
> > @zachschoenfeld33 does the grammar also allow use of a machine-sized int or float, e.g. complex[int] or complex[float]?
> 
> Do we really require both? It seems like we are complicating the language when we do not have a clear and pressing use case.

Not only we don't need both, but a declaration of the form:

```
complex[int[32]] C  = 13 / 6 im;
```

creates a number of problems, while providing no obvious benefit.

1. For starters, 13 / 6 will not evaluate to an Integer. It will evaluate to a Real.

2. Now that we're dealing with Complex Real numbers: what is their bit width? The programmer only told us the bit width of the Integers. They did not tell us the resulting bit width of their implicit conversion to a Real.

blakejohnson (dismissed):

1ucian0:
The TSC consensus is to merge this PR as it is and delegate on the type WG to solve the int/float issue.

blakejohnson (approved):

taalexander (approved):

taalexander:
Pinging @zachschoenfeld33 to sign the CLA so this may be merged.

jakelishman:
@taalexander: I think Zach had already signed the CLA, the problem is that the email address used in these commits isn't linked to the account, so it doesn't register correctly.

PR #158, opened 2021-03-03
[ANTLR] remove space between 'stretch' and N

hhorii:
This PR will fix issue #145 by adding a definition of Lexer grammar.

```
timingType
    : 'length'
    | StretchN
    ;

StretchN : 'stretch'Digit* ;
Identifier : FirstIdCharacter GeneralIdCharacter* ;
```

This qasm file and a generated tree from the file are here:
```
OPENQASM 3;
include "stdgates.inc";

stretch s;
stretch1 s1;
stretch10 s10;
stretch100 s100;
stretch255 s255;
```

```
program
  header
    version
      OPENQASM
      3
      ;
    include
      include
      "stdgates.inc"
      ;
  statement
    classicalDeclarationStatement
      classicalDeclaration
        noDesignatorDeclaration
          noDesignatorType
            timingType
              stretch
          identifierList
            s
      ;
  statement
    classicalDeclarationStatement
      classicalDeclaration
        noDesignatorDeclaration
          noDesignatorType
            timingType
              stretch1
          identifierList
            s1
      ;
  statement
    classicalDeclarationStatement
      classicalDeclaration
        noDesignatorDeclaration
          noDesignatorType
            timingType
              stretch10
          identifierList
            s10
      ;
  statement
    classicalDeclarationStatement
      classicalDeclaration
        noDesignatorDeclaration
          noDesignatorType
            timingType
              stretch100
          identifierList
            s100
      ;
  statement
    classicalDeclarationStatement
      classicalDeclaration
        noDesignatorDeclaration
          noDesignatorType
            timingType
              stretch255
          identifierList
            s255
      ;
```

PR #180, opened 2021-04-19
[ANTLR] Unite expression hierarchy

### LINKED ISSUE #145: [ANTLR] wrong definition of stretch type
Current stretch type is defined [as follows](https://github.com/Qiskit/openqasm/blob/0b3fe89214dedd6831926ee56ca5cbf599f423f5/source/grammar/qasm3.g4#L406-L409):
```
timingType
    : 'length'
    | 'stretch' Integer?
    ;
```
I think this is wrong because no space exists between `stretch` and `N` in [the spec](https://qiskit.github.io/openqasm/language/types.html#stretch).
> We distinguish different “orders” of stretch via stretchN types, where N is an integer between 0 to 255. stretch0 is an alias for the regular stretch.

The most naive (and not-smart) way to solve this issue is listing all the stretch types.
```
timingType
    : 'length'
    | 'stretch'
    | 'stretch0'
    | 'stretch1'
    ...
```
I'm not sure we can define`stretchN` with fragment `Digit*`.

### LINKED ISSUE #177: [ANTLR] subroutineCall should take a indexIdentifierList instead of expressionList
Currently, `subroutineCall` takes a `expressionList `.

```
subroutineCall
    : Identifier ( LPAREN expressionList? RPAREN )? expressionList
```

I guess that this `expressionList` should be replaced with `indexIdentifierList` as `quantumGateCall` does.

```
quantumGateCall
    : quantumGateModifier* quantumGateName ( LPAREN expressionList? RPAREN )? indexIdentifierList
    ;
```

hhorii:
This PR adds `booleanExpression` into the `expression` hierarchy.

Previously, `booleanExpression` is independent from the `expression` hierarchy and `branchingStatement` used it for test expression. However, `branchingStatement`  may take `subroutineCall` that returns `true` or `false`. Also, `bool` is a type and `expression` needs to include `booleanExpression` to assign a boolean value.

Here are details of changes in qasm.g4.

**1. Divide logical operator expression**
In original, logical operators (`&&`, `||`) are in a definition as `booleanExpression logicalOperator comparsionExpression`. However, this is wrong in most of programming languages. `&&` is evaluated before `||`. This PR divide this logical operator expression into `logicalAndExpression` and `logicalOrExpression`.

I believe that `logicalOrExpression` is the weakest. So, `expression` starts `logicalOrExpression` as follows:

```
expression:
  : ...
    | logicalAndExpression
    | expression '||' logicalAndExpression
    ;

logicalAndExpression
    : comparisonExpression
    | logicalAndExpression '&&' comparisonExpression
    ;
```

**2. `relationalOperator` includes `in`**
In [Python](https://github.com/antlr/grammars-v4/blob/master/python/python3-py/Python3.g4#L232-L235), `in` in `membershipTest` is a comparison operator because a result of `membershipTest` is also an expression (`true` or `false`). For example, following statement should be acceptable in QASM3:
```
a in { b, c } == d == e;
```
In Python, it is evaluated as `(((a in [ b, c ] )== d) == e);`.
Because `relationalOperator` includes `in`, `expressionTerminator` needs to have `setDeclaration`.

```
expressionTerminator
    : ...
    | setDeclaration
    | ...
    ;
```

**3. Remove `booleanExpression`**
In original, `booleanExpression` is used in tests of branches and while loops. Because they may take subroutine or kernel calls, their tests should be `expression`.


source/grammar/qasm3.g4
```diff
@@ -470,7 +503,7 @@ pragma
 
 timingType
     : 'length'
-    | 'stretch' Integer?
+    | StretchN
```
zachschoenfeld33, line 506: Please note the other PR's this is pulling in; will make it easier to review. Let's also focus on getting those merged in first

source/grammar/qasm3.g4
```diff
@@ -276,34 +279,73 @@ logicalOperator
     | '||'
     ;
 
+shiftOperator
```
zachschoenfeld33, line 282: Don't think this is necessary. It will just make the resulting parse tree a longer hierarchy. I think the way it was before was clear.

source/grammar/qasm3.g4
```diff
@@ -276,34 +279,73 @@ logicalOperator
...
+    ;
+
 expressionStatement
     : expression SEMICOLON
     ;
 
 expression
     // include terminator/unary as base cases to simplify parsing
     : expressionTerminator
-    | unaryExpression
```
zachschoenfeld33, line 286: I have these base cases to simplify the parse tree--see the test results you generated vs the ones that were there before. Your parse trees are much longer are more complicated. If we have a terminator or a unary expr, we don't necessarily need to go through the entire expression hierarchy

source/grammar/qasm3.g4
```diff
@@ -276,34 +279,73 @@ logicalOperator
...
+    : orExpression  // if (expression)
+    | comparisonExpression relationalOperator orExpression
+    ;
+
+orExpression
+    : xorExpression
+    | orExpression '|' xorExpression
+    ;
+
+xorExpression
```
zachschoenfeld33, line 346: I prefer the camel case `xOrExpression`. Is consistent w the rest of the ANTLR grammar.

source/grammar/qasm3.g4
```diff
@@ -276,34 +279,73 @@ logicalOperator
...
-xOrExpression
+booleanExpression
+    : membershipTest
+    | comparisonExpression
+    | booleanExpression logicalOperator comparisonExpression
+    ;
+
+comparisonExpression
+    : orExpression  // if (expression)
+    | comparisonExpression relationalOperator orExpression
```
zachschoenfeld33, line 338: Again, the reason I used `expression` here is that for simple cases like expression terminator and unary, you don't need to go through the entire hierarchy. Makes the resulting parse tree far clearer and is still accurate.

source/grammar/qasm3.g4
```diff
@@ -276,34 +279,73 @@ logicalOperator
...
   *  grammar: https://github.com/antlr/grammars-v4/blob/master/c/C.g4
   * Order (first to last evaluation):
     Terminator (including Parens),
     Unary Op,
+    Power
     Multiplicative
     Additive
     Bit Shift
     Bit And
     Exlusive Or (xOr)
```
zachschoenfeld33, line 323:
```suggestion
    Exclusive Or (xOr)
```

source/grammar/qasm3.g4
```diff
@@ -276,34 +279,73 @@ logicalOperator
...
     Bit Or
+    Comparison
```
zachschoenfeld33, line 325: I'd just replace these w/ boolean expression

source/grammar/qasm3.g4
```diff
@@ -276,34 +279,73 @@ logicalOperator
...
 expressionStatement
     : expression SEMICOLON
     ;
 
 expression
     // include terminator/unary as base cases to simplify parsing
     : expressionTerminator
-    | unaryExpression
     // expression hierarchy
-    | xOrExpression
```
zachschoenfeld33, line 288: I might just add `powerExpression` as another base case w terminator and unary

source/grammar/qasm3.g4
```diff
@@ -276,34 +279,73 @@ logicalOperator
...
-    | expression '|' xOrExpression
+    | booleanExpression
     ;
 
 /**  Expression hierarchy for non-terminators. Adapted from ANTLR4 C
   *  grammar: https://github.com/antlr/grammars-v4/blob/master/c/C.g4
   * Order (first to last evaluation):
     Terminator (including Parens),
     Unary Op,
+    Power
```
zachschoenfeld33, line 318: Isn't power above unary? Guess that would depend on the convention we choose, but at least in some languages it is

source/grammar/tests/outputs/branch_binop.yaml
```diff
@@ -10,50 +10,83 @@ reference: |
...
-            &&
-          comparsionExpression
-            expression
-              expressionTerminator
-                expressionTerminator
-                  spec
-                [
-                expression
-                  xOrExpression
+            booleanExpression
```
zachschoenfeld33, line 15: Note difference in length

source/grammar/tests/outputs/loop.yaml
```diff
@@ -14,16 +14,30 @@ reference: |
         loopSignature
           while
           (
-          booleanExpression
```
zachschoenfeld33, line 17: Note difference in length

source/grammar/tests/outputs/unary_expr.yaml
```diff
@@ -7,9 +7,19 @@ reference: |
     statement
       expressionStatement
         expression
-          unaryExpression
-            unaryOperator
-              !
-            expressionTerminator
-              my_var
+          booleanExpression
```
zachschoenfeld33, line 10: Note difference in length

zachschoenfeld33 (commented):
Added some initial comments. The changes make sense. Some, however, add to the complexity of the parse tree without adding to correctness, so don't think these are necessary. We should also try to get the PR's that this is dependent on merged to simplify the diff for review purposes

source/grammar/qasm3.g4
hhorii: This is in #158. It can be merged soon, I believe.

hhorii:
I believe that long depth of a parsed tree is not a problem because human does not read it. Simplicity and explainability of the grammar is prioritized.

source/grammar/qasm3.g4
hhorii: Original `booleanExpression` is not correct. LogicalAND (&&) must be prioritized than LogicalOR (||) if QASM3 follows C.

I do not think `membershipTest` is necessary in the hierarchy because `in` is a kind of `logicalOperator` in other language such as Python. If we introduce an array types, `membershipTest` will become `Identifier 'in' orExpression`, which is mostly same with `comparisonExpression`: `comparisonExpression relationalOperator orExpression`. In this PR, I proposed `relationalOperator` includes `in` and simplified the hierarchy by merging `membershipTest` into `comparisonExpression`.

Original `booleanExpression`:
```
booleanExpression
    : membershipTest
    | comparisonExpression
    | booleanExpression logicalOperator comparisonExpression
    ;

comparisonExpression
    : expression  // if (expression)
    | expression relationalOperator expression
    ;
```

zachschoenfeld33: Fair pt

source/grammar/qasm3.g4
```diff
@@ -340,6 +360,7 @@ expressionTerminator
     | Integer
     | RealNumber
     | Identifier
+    | setDeclaration
```
zachschoenfeld33, line 363: I think this needs to be `membershipTest`. I don't think `{1,2,3}` is an expression on its own. Should be the boolean result of a measurement test

hhorii: `{1,2,3}` looks an array. Though current spec does not say any about array-type, initialization of its variables will be this form.

zachschoenfeld33 (commented):
Two changes:
1. `in` keyword only in for loop for now.
2. register slicing syntax only for Alias (`let` statements)

zachschoenfeld33:
Can you also add a comment in the specification (in the classical types section) stating that the order of operation hierarchy is formally defined in the ANTLR grammar? (you can list it out too, up to you) It would be good to have that officially outlined in the specification though

hhorii:
>     1. `in` keyword only in for loop for now.
>     2. register slicing syntax only for Alias (`let` statements)

@zachschoenfeld33 4e8df07 covers 1 but not 2 because some examples use slicing not in alias statements.

https://github.com/Qiskit/openqasm/blob/3b23fa7a0b74ef7a721cfb8172232edf1fc8df40/examples/adder.qasm#L41

https://github.com/Qiskit/openqasm/blob/3b23fa7a0b74ef7a721cfb8172232edf1fc8df40/examples/msd.qasm#L27

https://github.com/Qiskit/openqasm/blob/3b23fa7a0b74ef7a721cfb8172232edf1fc8df40/examples/msd.qasm#L126

https://github.com/Qiskit/openqasm/blob/3b23fa7a0b74ef7a721cfb8172232edf1fc8df40/examples/msd.qasm#L130

Probably, slicing syntax should be allowed in the spec for any qubit index specification.

source/language/classical.rst
```diff
@@ -91,7 +91,7 @@ increment ``++`` and decrement ``--``.
    int[32] a = 2;
    int[32] b = 3;
 
-   a * b; // 5
+   a * b; // 6
```
zachschoenfeld33, line 94: Good catch!

source/language/classical.rst
```diff
@@ -110,6 +110,40 @@ multiplication, and division and the corresponding assignment operators.
    angle[10] c;
    c = angle(a + b); // cast to angle[10]
 
+
+Evaluation order
+~~~~~~~~~~~~~~~~
+
+OpenQASM evaluates expressions from left to right.
+The operator precedence is based on C.
```
zachschoenfeld33, line 118:
```suggestion
```
I say remove the line about C. We don't want to be held to that. The order is defined by the table below (and the accompanying ANTLR .g4 file)

source/language/classical.rst
```diff
@@ -110,6 +110,40 @@ multiplication, and division and the corresponding assignment operators.
...
+
+   .. table:: [operator-precedence] operator precedence in OpenQASM.
+
+      +-------------------------------+
+      | Operator                      |
+      +-------------------------------+
+      | +x, -x, ~x                    |
```
zachschoenfeld33, line 125: Can you make two columns? One for Operator Type and one for examples. So something like
```
Operator Type: Unary
Example: !, ~, ...
```

zachschoenfeld33 (commented):
@hhorii good point on the slicing. You are right.

source/grammar/qasm3.g4
```diff
@@ -256,15 +256,18 @@ quantumGateCall
 /*** Classical Instructions ***/
 
 unaryOperator
-    : '~' | '!'
+    : '~' | '!' | '-'
```
zachschoenfeld33, line 259: Good change

source/grammar/qasm3.g4
```diff
@@ -294,18 +297,43 @@ expression
...
+    Logical and
+    Logical or
 **/
+
+logicalAndExpression
+    : orExpression
+    | logicalAndExpression '&&' orExpression
+    ;
+
+orExpression
```
zachschoenfeld33, line 314: Should be `bitOrExpression`

source/grammar/qasm3.g4
```diff
@@ -389,8 +403,8 @@ assignmentOperator
     | '+=' | '-=' | '*=' | '/=' | '&=' | '|=' | '~=' | '^=' | '<<=' | '>>='
     ;
 
-membershipTest
-    : Identifier 'in' setDeclaration
+equalsAssignmentList
```
zachschoenfeld33, line 406: Remove `equalsAssignmentList` (not used anywhere)--this is a recent change...

source/grammar/tests/outputs/assignment.yaml
```diff
@@ -6,6 +6,9 @@ source: |
   a[0] = b[1];
   x += int[10](a[1]);
   a = measure q[1:2];
+  b = a == 0;
+  a = b in { c, d } == e == f;
```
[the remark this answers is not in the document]
zachschoenfeld33: This is fixed. `in` only allowed in `for` loops for now

source/grammar/qasm3.g4
zachschoenfeld33: This has been fixed.

source/language/classical.rst
```diff
@@ -110,6 +110,43 @@ multiplication, and division and the corresponding assignment operators.
...
+OpenQASM evaluates expressions from left to right.
+
+   .. table:: [operator-precedence] operator precedence in OpenQASM.
+
+      +-------------------------------+---------------------------------------+
+      | Operator                      | Descriptions                          |
+      +-------------------------------+---------------------------------------+
+      | (), [], ++, --                | postfix for call, index, incrementors |
+      +-------------------------------+---------------------------------------+
+      | +x, -x, ~x                    | unary operations                      |
```
zachschoenfeld33, line 126:
```suggestion
      | !, ~, -, **                   | Unary                      |
```

source/language/classical.rst
```diff
@@ -110,6 +110,43 @@ multiplication, and division and the corresponding assignment operators.
...
+~~~~~~~~~~~~~~~~
+
+OpenQASM evaluates expressions from left to right.
+
+   .. table:: [operator-precedence] operator precedence in OpenQASM.
+
+      +-------------------------------+---------------------------------------+
+      | Operator                      | Descriptions                          |
+      +-------------------------------+---------------------------------------+
+      | (), [], ++, --                | postfix for call, index, incrementors |
```
zachschoenfeld33, line 124:
```suggestion
      | (), ++, cast, literals, etc...                | Expression terminator |
```

source/language/classical.rst
```diff
@@ -110,6 +110,43 @@ multiplication, and division and the corresponding assignment operators.
...
+
+Evaluation order
+~~~~~~~~~~~~~~~~
+
+OpenQASM evaluates expressions from left to right.
+
+   .. table:: [operator-precedence] operator precedence in OpenQASM.
+
+      +-------------------------------+---------------------------------------+
+      | Operator                      | Descriptions                          |
```
zachschoenfeld33, line 122:
```suggestion
      | Operator                      | Operator Type                          |
```

source/language/classical.rst
```diff
@@ -110,6 +110,43 @@ multiplication, and division and the corresponding assignment operators.
...
+      +-------------------------------+---------------------------------------+
+      | (), [], ++, --                | postfix for call, index, incrementors |
+      +-------------------------------+---------------------------------------+
+      | +x, -x, ~x                    | unary operations                      |
+      +-------------------------------+---------------------------------------+
+      | *, /                          | Multiplicative operations             |
```
zachschoenfeld33, line 128:
```suggestion
      | *, /, %                          | Multiplicative             |
```
(Forgot modulo)

source/language/classical.rst
zachschoenfeld33: Forgot `**`

source/language/classical.rst
```diff
@@ -110,6 +110,43 @@ multiplication, and division and the corresponding assignment operators.
...
+      +-------------------------------+---------------------------------------+
+      | +, -                          | Additive operations                   |
```
zachschoenfeld33, line 130: Can take the word "operations/operation" out of all of these

source/language/classical.rst
```diff
@@ -110,6 +110,43 @@ multiplication, and division and the corresponding assignment operators.
...
+      +-------------------------------+---------------------------------------+
+      | <<, >>                        | Shift operations                      |
```
zachschoenfeld33, line 132:
```suggestion
      | <<, >>                        | Bit Shift operations                      |
```

source/language/classical.rst
zachschoenfeld33: (I do know by the way that further operator precedence may need to be defined at the terminator level, but lets leave that for a future PR...)

zachschoenfeld33 (commented):
Almost there. Can you add `%` as a valid operation (modulo) to integers and `**` (power) to both integers and floats/angles (this was agreed upon in https://github.com/Qiskit/openqasm/pull/135). 

I suggested two other minor changes to the syntax, and then we should be good

source/grammar/qasm3.g4
```diff
@@ -294,18 +297,43 @@ expression
     Multiplicative
     Additive
     Bit Shift
+    Comparison
+    Equality
     Bit And
     Exlusive Or (xOr)
     Bit Or
+    Logical and
```
zachschoenfeld33, line 305:
```suggestion
    Logical And
```

source/grammar/qasm3.g4
```diff
@@ -294,18 +297,43 @@ expression
...
+    Logical or
```
zachschoenfeld33, line 306:
```suggestion
    Logical Or
```

source/language/classical.rst
hhorii: I intentionally take `**` from this PR because we need discussion.

In python, `**` is evaluated before unary. `-2 ** 2` is evaluated as `- (2 ** 2)` resulting `-4`. In addition, it is natural to add `**=` as a new assignment operator if `**` is introduced.

I will send a new PR that follows this to add `**`.

source/language/classical.rst
hhorii: We do not need to list all the "expression terminator" in this table because some (such as literal and numbers) do not need operator precedence.

source/language/classical.rst
```diff
@@ -110,6 +110,43 @@ multiplication, and division and the corresponding assignment operators.
...
+OpenQASM evaluates expressions from left to right.
+
+   .. table:: [operator-precedence] operator precedence in OpenQASM.
+
+      +-------------------------------+---------------------------------------+
+      | Operator                      | Operator Types                        |
+      +-------------------------------+---------------------------------------+
+      | (), [], ++, (type)(x)         | call, index, incrementors, cast       |
+      +-------------------------------+---------------------------------------+
+      | +, -, ~                       | unary                                 |
```
zachschoenfeld33, line 126:
```suggestion
      | !, ~, - (minus)                       | Unary                                 |
```

source/language/classical.rst
```diff
@@ -91,7 +91,7 @@ increment ``++`` and decrement ``--``.
    int[32] a = 2;
    int[32] b = 3;
 
-   a * b; // 5
```
zachschoenfeld33, line 94: Can you add that integers also support the modulo operator?

source/language/classical.rst
```diff
@@ -110,6 +110,43 @@ multiplication, and division and the corresponding assignment operators.
...
+~~~~~~~~~~~~~~~~
+
+OpenQASM evaluates expressions from left to right.
+
+   .. table:: [operator-precedence] operator precedence in OpenQASM.
+
+      +-------------------------------+---------------------------------------+
+      | Operator                      | Operator Types                        |
+      +-------------------------------+---------------------------------------+
+      | (), [], ++, (type)(x)         | call, index, incrementors, cast       |
```
zachschoenfeld33, line 124:
```suggestion
      | (), [], ++, (type)(x)         | Call, index, incrementors, cast       |
```

zachschoenfeld33 (commented):
Few more very minor comments. I thought maybe we should clarify the difference between unary minus (-) and the binary subtraction sign.

Also we need to add power--up to you if you'd like to add that here, or in another PR. It is currently here: https://github.com/Qiskit/openqasm/pull/135

source/language/classical.rst
zachschoenfeld33: Ok sure

source/language/classical.rst
zachschoenfeld33: Ah got it

source/language/classical.rst
hhorii: good point :-)

source/language/classical.rst
```diff
@@ -110,6 +111,43 @@ multiplication, and division and the corresponding assignment operators.
...
+      +-------------------------------+---------------------------------------+
+      | <, <=, >, >=                  | Comparison                            |
+      +-------------------------------+---------------------------------------+
+      | !=, ==                        | Equality                              |
+      +-------------------------------+---------------------------------------+
+      | &                             | Bitwise AND                           |
+      +-------------------------------+---------------------------------------+
+      | ^                             | Bitwise XOR                           |
+      +-------------------------------+---------------------------------------+
+      | |                             | Bitwise OR                            |
```
zachschoenfeld33, line 143: The `|` seems to not show up in the table? May need to write it as a special char (or maybe this is an error in markdown and it will appear in sphinx)

source/language/classical.rst
```diff
@@ -110,6 +111,43 @@ multiplication, and division and the corresponding assignment operators.
    angle[10] c;
    c = angle(a + b); // cast to angle[10]
 
+
+Evaluation order
+~~~~~~~~~~~~~~~~
+
+OpenQASM evaluates expressions from left to right.
+
+   .. table:: [operator-precedence] operator precedence in OpenQASM.
```
taalexander, line 120:
```suggestion
   .. table:: [operator-precedence] operator precedence in OpenQASM ordered from highest precedence to lowest precedence. Higher precedence operators will be evaluated first.
```

taalexander (changes requested):
Thank you @hhorii, these changes make sense.

zachschoenfeld33:
One other question before we merge this: I just realized that `||` is used for register concatenation and now also for logical or. While there shouldn't be ambiguity, should we consider removing this double usage?

Options:
- Use `and/or` for logical and/or as in Python
- Use `+` for concatenation as is standard in many other languages.

hhorii:
If concatenation (`||`) exists only in alias, a parser can identify `||` for concatenation or logical or, I believe.

BTW, I like `+` option.

taalexander (approved):

PR #161, opened 2021-03-11
Remove colons between types and identifiers in subroutine headers.

blakejohnson:

PR #190, opened 2021-05-10
Fix parameter/colon syntax and declaration of `qubit/bit` registers

zachschoenfeld33:
### Summary

- Removes colons from subroutine and `defcal` definitions to be in line w/ https://arxiv.org/pdf/2104.14722.pdf and https://github.com/Qiskit/openqasm/pull/161
- Fix declaration of `qubit/bit` registers (builds off https://github.com/Qiskit/openqasm/pull/189 by @jwoehr)
  - Based on OpenQASM3 paper, qubits/bits declared as: `qubit[n] q; bit[n] b;` and the legacy versions as `qreg q[n]; creg b[n];`

### DIFF
--- a/source/bibliography.bib
+++ b/source/bibliography.bib
@@ -6,6 +6,15 @@
     note = {Accessed: 2020-05-04}
 }
 
+@misc{cross2021openqasm,
+       title={OpenQASM 3: A broader and deeper quantum assembly language},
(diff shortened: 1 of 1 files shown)


zachschoenfeld33:
@jwoehr

jwoehr:
Yes, @zachschoenfeld33, this is more complete and where it overlaps what I submitted it's better antlr4. 💯

PR #191, opened 2021-05-11
Pyg c err

jwoehr:
In files where greek math characters are used, pygments throws an exception which makes sphinx spew errors for each `.. :code block:: c`
```
WARNING: Could not lex literal_block as "c". Highlighting skipped.
```
The blocks are translated without highlighting.

I added the `:force:` directive. The blocks are still translated without highlighting but the exceptions stop.

The only real fix for this is to PR the pygments project with a grammar for OpenQASM but this at least makes the build a little cleaner.


blakejohnson (approved):
LGTM

PR #192, opened 2021-05-11
fix a bug to use wrong variable names in adder.qasm

hhorii:
names of variables are wrong in adder.qasm. This PR fixes them.


zachschoenfeld33 (approved):
LGTM

Issue #179, opened 2021-04-17
[ANTLR] 'true' and 'false' are expressionTerminator


hhorii:
The spec defines `true` and `false` .

> There is a Boolean type bool name; that takes values `true` or `false`.
https://qiskit.github.io/openqasm/language/types.html#boolean-types

Currently, `true` and `false` are pased as `Identity`.

zachschoenfeld33:
@hhorii would you mind adding a PR fixing this when you can?

Issue #143, opened 2021-02-10
[ANTLR] Duplication of return types in kernel declaration 


hhorii:
Currently, `kernelDeclaration` is defined as follows [in the grammar](https://github.com/Qiskit/openqasm/blob/0b3fe89214dedd6831926ee56ca5cbf599f423f5/source/grammar/qasm3.g4#L233-L235).

```
kernelDeclaration
    : 'kernel' Identifier ( LPAREN classicalTypeList? RPAREN )? returnSignature?
    classicalType? SEMICOLON
```

This `returnSignature` is defined as follows, while including `classicalType`.

```
returnSignature
    : ARROW classicalType
    ;
```

On the other hand, `classicalType` is defined in `kernelDeclaration` after this `returnSignature`. Therefore, from the grammar, two `classicalType` can be listed in a kernel declaration after argument specification.

If both of `ARROW classicalType` and `classicalType` are allowed to specify a return type, the current grammar may be reasonable (though `(returnSignature? | classicalType?)` is better). If not, `classicalType?` is not necessary in `kernelDeclaration`.

Currently, this qasm can be parsed.
```
OPENQASM 3;
kernel test_kernel (int[64], int[64], int[64]) -> int[64] int[64];
```

zachschoenfeld33:
Good catch. Should only be a single return type. You can remove the additional `classicalType?` following `returnSignature?`. It should be done how subroutines do it, which are valid.

zachschoenfeld33:
Fixed in https://github.com/Qiskit/openqasm/pull/195

Issue #142, opened 2021-02-09
[ANTLR] gphase needs to take expression


hhorii:
Example of `gphase` [in the spec](https://qiskit.github.io/openqasm/language/gates.html?highlight=gphase
)  is `gphase(-tau/2);` and this `-tau/2` looks `expression` in the grammar. However, [current grammar](https://github.com/Qiskit/openqasm/blob/0b3fe89214dedd6831926ee56ca5cbf599f423f5/source/grammar/qasm3.g4#L233-L235
) takes `Identifier`.

```
quantumPhase
    : 'gphase' LPAREN Identifier RPAREN
    ;
```

zachschoenfeld33:
fixed in https://github.com/Qiskit/openqasm/pull/195.

PR #195, opened 2021-05-12
[ANTLR] Small fixes

zachschoenfeld33:
Minor ANTLR fixes to close https://github.com/Qiskit/openqasm/issues/179, https://github.com/Qiskit/openqasm/issues/143 and https://github.com/Qiskit/openqasm/issues/142.

These handle `true/false` literals, fix a return type issue and allow `expression` type to be passed to `gphase`.

`%` is also eliminated as a valid identifier as it intersects with modulo.

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -8,8 +8,7 @@ Types and Casting
 Generalities
 ------------
 
-Variable identifiers must begin with a letter [A-Za-z], an underscore, a
-percent sign, or an element from the Unicode character categories
+Variable identifiers must begin with a letter [A-Za-z], an underscore or an element from the Unicode character categories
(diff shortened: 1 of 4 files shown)


source/language/types.rst
```diff
@@ -8,8 +8,7 @@ Types and Casting
 Generalities
 ------------
 
-Variable identifiers must begin with a letter [A-Za-z], an underscore, a
-percent sign, or an element from the Unicode character categories
+Variable identifiers must begin with a letter [A-Za-z], an underscore or an element from the Unicode character categories
```
zachschoenfeld33, line 11: To be clear, percent is modulo which is why we disallow `%` for identifier.

taalexander (approved):

PR #199, opened 2021-05-20
Paper: Allow no args for `barrier`

zachschoenfeld33:
### Summary

According to the paper https://arxiv.org/pdf/2104.14722.pdf, `barrier` can be used with no arguments outside of a box. This PR updates the spec and grammar accordingly.


zachschoenfeld33:
Two questions:
1. Why is the no-argument form only allowed outside a box? This PR allows it inside a box as well. If we define the scoping of the barrier to be local to the box, it shouldn't be an issue.
2. Perhaps we should include a special instruction `barrier_all` instead? The current form lends itself to the simple mistake of forgetting to type the qubit/qreg id--it might be nice instead to see a compiler error in this case.

jwoehr:
>     2. Perhaps we should include a special instruction `barrier_all` instead? The current form lends itself to the simple mistake of forgetting to type the qubit/qreg id--it might be nice instead to see a compiler error in this case.

This is a cogent argument.

Parenthetically, neither `barrier` nor `measure` are allowed without arguments in OpenQASM 2.x although they certainly could have been.

Thus, the "forgetting to type" argument is only the stronger as (half-finished, half-forgotten) erroneous code from users' archives that would have errored out in OpenQASM2 would now execute, possibly incorrectly, without warning.

jwoehr:
Though on the other hand the proposal has its attractions.
Perhaps a translator switch or kwarg or two such as:

- `enforce_qasm2_semantics`
- `warn_on_zero_arity_metaops`

blakejohnson (approved):

blakejohnson (approved):

PR #201, opened 2021-05-21
Forbid identifier lists in measurement assignments

1ucian0:
After checking with @zachschoenfeld33, it seems like the `quantumMeasurementAssignment` grammar class needs to be adjusted a bit. Currently it looks like this:
```
quantumMeasurementAssignment
    : quantumMeasurement ( ARROW indexIdentifierList)?
    | indexIdentifierList EQUALS quantumMeasurement
    ;
```
In the first line `ARROW indexIdentifierList` should not be optional. In the second line, the idea of multi assignments was discarded and `indexIdentifierList` should be replaced by `indexIdentifier`. The final result is
```
quantumMeasurementAssignment
    : quantumMeasurement ARROW indexIdentifierList
    | indexIdentifier EQUALS quantumMeasurement
    ;
```

This PR adjust the class `quantumMeasurementAssignment` accordingly and adapt the related test.


source/grammar/qasm3.g4
```diff
@@ -237,8 +237,8 @@ quantumMeasurement
     ;
 
 quantumMeasurementAssignment
-    : quantumMeasurement ( ARROW indexIdentifierList)?
-    | indexIdentifierList EQUALS quantumMeasurement
+    : quantumMeasurement ARROW indexIdentifierList
```
zachschoenfeld33, line 240:
```suggestion
    : quantumMeasurement ARROW indexIdentifier
```
Single assignment only

source/grammar/qasm3.g4
```diff
@@ -237,8 +237,8 @@ quantumMeasurement
     ;
```
zachschoenfeld33, line 237: The line above `quantumMeasurement`, should only take `indexIdentifier`. That is,
```
quantumMeasurement    
    : 'measure' indexIdentifier
    ;
```

zachschoenfeld33:
@1ucian0 it could be beneficial if you could add a small test to the ANTLR grammar w/ the different measurement types (happy to help w this).

1ucian0:
> @1ucian0 it could be beneficial if you could add a small test to the ANTLR grammar w/ the different measurement types (happy to help w this).

Sure... I dont fully get the way this is being tested. Maybe in a follow up PR?

source/grammar/qasm3.g4
zachschoenfeld33: @1ucian0 pls update this. We do not allow for measuring multiple individual qubits at once. In order to measure multiple qubits in the same instruction, you'd need to use a qubit register. See https://qiskit.github.io/openqasm/language/insts.html#measurement

source/grammar/qasm3.g4
```diff
@@ -237,8 +237,8 @@ quantumMeasurement
...
 
 quantumMeasurementAssignment
-    : quantumMeasurement ( ARROW indexIdentifierList)?
-    | indexIdentifierList EQUALS quantumMeasurement
+    : quantumMeasurement ARROW indexIdentifier
```
zachschoenfeld33, line 240:
```suggestion
    : quantumMeasurement ( ARROW indexIdentifier )?
```
Need this to new optional. Something like `measure a`, without assignment, is valid.

zachschoenfeld33:
> > @1ucian0 it could be beneficial if you could add a small test to the ANTLR grammar w/ the different measurement types (happy to help w this).
> 
> Sure... I dont fully get the way this is being tested. Maybe in a follow up PR?

You can add a line here (https://github.com/Qiskit/openqasm/blob/3cc82855730b512bdab17c3c267a1f4ba3910e6a/source/grammar/tests/outputs/assignment.yaml#L9) that tests the `->` assignment measurement. And then verify that the correct ANTLR is produced.

blakejohnson:
It looks like a few more changes are needed here?

taalexander:
Bump on this @1ucian0?

source/grammar/qasm3.g4
1ucian0: done in 21ee0533ec3ac06bf93d969885ed34ea6e77306b

1ucian0:
> You can add a line here (
> https://github.com/Qiskit/openqasm/blob/3cc82855730b512bdab17c3c267a1f4ba3910e6a/source/grammar/tests/outputs/assignment.yaml#L9
> 
> ) that tests the `->` assignment measurement. And then verify that the correct ANTLR is produced.

done in 7db94a5a82262757391fc426bc38159c10a78586

1ucian0:
all change requests addressed.

taalexander (commented):
This looks good to me, would you mind taking a look as well @jakelishman?

jakelishman (approved):
Looks good to me too.  I've just pushed up some tests of invalid measurements to add to this, including ones that are explicitly forbidden by this PR, to catch regressions.

PR #209, opened 2021-05-22
Remove extra P at line 41

aspcompiler:
Not sure if this 'P' is needed.


taalexander (approved):
Thank you for catching and fixing this. Pending CI passing the fix looks good.

Issue #185, opened 2021-05-04
[ANTLR] delay instruction will be parsed to expressionStatement


godott:
1. Current parsing rule for `expressionStatement` is too general. Timing statement (which is also quantumStatement) like `delay[lengthof(a)] q[2], q[3];` is parsed as `expressionStatement`. `expressionStatement` has to be more specifically defined. 

2. At the same time, it means `delay[lengthof(a)]` is parsed as `indexIdentifier` as opposed to `TimingInstruction`. The solution can be defining `DELAY: 'delay'` so it will overwrite `identifier`.

godott:
From my observasion, the only expressionStatement is incrementors like `i++`.

hhorii:
I parsed `delay[lengthof(a)] q[2], q[3];` with the master 89e93e7, and then got following tree:
```
program
  header
  statement
    quantumStatement
      timingStatement
        timingInstruction
          timingInstructionName
            delay
          designator
            [
            expression
              expressionTerminator
                builtInCall
                  builtInMath
                    lengthof
                  (
                  expressionList
                    expression
                      expressionTerminator
                        a
                  )
            ]
          indexIdentifierList
            indexIdentifier
              q
              [
              expressionList
                expression
                  expressionTerminator
                    2
              ]
            ,
            indexIdentifier
              q
              [
              expressionList
                expression
                  expressionTerminator
                    3
              ]
        ;
```

is this issue resolved in your environment? > @godott

PR #217, opened 2021-05-31
replace subroutineCall and kernelCall with functionCall in ANTLR

hhorii:
Currently the parser distinguishes calls of subroutine and kernel.

```
kernelCall
    : Identifier LPAREN expressionList? RPAREN
    ;

subroutineCall
    : Identifier ( LPAREN expressionList? RPAREN )? indexIdentifierList
    ;
```

First, this `subroutineCall` cannot parse a subroutine call that does not take any qubit arguments (because `indexIdentifierList` has at least one element).

Second, if `indexIdentifierList` of `subroutineCall` is optional, no difference between `kernelCall` and `subroutineCall`.

This PR replaces above two definitions with a new definition `functionCall`:
```
functionCall
    : Identifier ( LPAREN expressionList? RPAREN )? indexIdentifierList?
    ;
```


zachschoenfeld33:
Looks good. Maybe you could call `functionCall` something like `kernSubCall`? So this way it relates to the vocabulary of OpenQASM3 (since we don't call them "functions").

It's also true that this could get mismatched w/ gate, but I think that's just an ambiguity of the language (which already existed prior to this PR).

stevenheidel:
We resolved this ambiguity in some last-minute discussions on the paper.

Subroutine calls take qubit arguments within the parenthesis. This is because the subroutine will not be broadcast across an array of qubits like is the case with gates.

See the example in section 3.3 of the paper:
` def segment(qubit[2] anc, qubit psi) -> bit[2] { `

stevenheidel:
Somewhat related: https://github.com/Qiskit/openqasm/issues/185

hhorii:
@stevenheidel thanks. I will close this PR and send another PR to support the paper's syntax.

source/grammar/qasm3.g4
```diff
@@ -447,27 +455,22 @@ kernelDeclaration
     : 'kernel' Identifier ( LPAREN classicalTypeList? RPAREN )? returnSignature? SEMICOLON
     ;
 
-// if have kernel w/ out args, is ambiguous; may get matched as identifier
-kernelCall
-    : Identifier LPAREN expressionList? RPAREN
+// if have function call w/ out args, is ambiguous; may get matched as identifier
+kernelOrSubroutineCall
+    : Identifier ( LPAREN expressionList? RPAREN )? indexIdentifierList?
```
zachschoenfeld33, line 460: I'm not sure this is correct. I think we need a list which allows either expression or index identifier and put that list into the parens. Something like `exprAndIndexIdList`

zachschoenfeld33: The test to add would be something like `sub_call(a, b, q[0] || q[1])` and see if that works...

source/grammar/tests/outputs/sub_and_kern_call.yaml
```diff
@@ -1,6 +1,6 @@
 # indent w/ 2 spaces
 source: |
-  bit x = sub_call(10, "01") q1[0], q2;
+  bit x = sub_call(10, "01", q1[0], q2);
```
zachschoenfeld33, line 3: Can you add index identifier specifics such as either `||` or `[:]` syntax to the test?

zachschoenfeld33 (commented):
Had one comment to fix...

source/grammar/qasm3.g4
hhorii: This was wrong... As @stevenheidel mentioned, `qubit` arguments must be in parenthesis in any calls. 

```
kernelOrSubroutineCall
    : Identifier ( LPAREN expressionList? RPAREN )?
    ;
```

Then, `sub_call(a, b, q[0] || q[1])` is parsed as

```
kernelOrSubroutineCall
  sub_call
  (
  expressionList
    expression
      expressionTerminator
        a
    ,
    expression
      expressionTerminator
        b
    ,
    expression
      expression
        expressionTerminator
          expressionTerminator
            q
          [
          expression
            expressionTerminator
              0
          ]
      ||
      logicalAndExpression
        bitOrExpression
          xOrExpression
            bitAndExpression
              equalityExpression
                comparisonExpression
                  bitShiftExpression
                    additiveExpression
                      multiplicativeExpression
                        expressionTerminator
                          expressionTerminator
                            q
                          [
                          expression
                            expressionTerminator
                              1
                          ]
```

zachschoenfeld33: Right--that's why I am suggesting a list that can either include expressions or index identifiers, which would be in the parenthesis. Something like:
```
exprOrIndexId : expression | indexIdentifier;
exprOrIndexIdList : (exprOrIndexId COMMA)* exprOrIndexId
```

zachschoenfeld33: I'm not sure we should add `||`, `:` to expression as they are specific expressions only allowed for registers. If we added to expression you could do `int x = [2:3]` which clearly doesn't make sense.

stevenheidel:
@hhorii @zachschoenfeld33 - are we okay with pulling in @aspcompiler 's fix?

zachschoenfeld33:
Yes works for me...

hhorii:
@aspcompiler Thank you for your help and I'm sorry that my response is very late. I cherry-picked your commit.

source/grammar/tests/outputs/sub_and_kern_call.yaml
hhorii: I'm not sure that such operators are allowed excepting alias statements.

blakejohnson:
@stevenheidel is this good to go, now?

source/grammar/qasm3.g4
```diff
@@ -464,7 +464,7 @@ externDeclaration
     ;
 
 // if have function call w/ out args, is ambiguous; may get matched as identifier
-kernelOrSubroutineCall
+externOrSubroutineCall
     : Identifier ( LPAREN expressionList? RPAREN )?
```
zachschoenfeld33, line 468: Are parentheses required in the case of no arguments? I would argue that they should be...

source/grammar/qasm3.g4
```diff
@@ -455,27 +463,22 @@ externDeclaration
...
+// if have function call w/ out args, is ambiguous; may get matched as identifier
+externOrSubroutineCall
+    : Identifier ( LPAREN expressionList? RPAREN )?
     ;
 
 /*** Subroutines ***/
 
 subroutineDefinition
-    : 'def' Identifier ( LPAREN classicalArgumentList? RPAREN )? quantumArgumentList?
+    : 'def' Identifier ( LPAREN anyTypeArgumentList? RPAREN )?
```
zachschoenfeld33, line 474: Same comment about whether parentheses should be required in the case of no arguments...

zachschoenfeld33 (commented):
One minor question: Should we require parentheses on `extern`'s and `subroutine`'s in the case of no arguments? I suggest that we do, as this will match classical function syntax and avoid any potential ambiguities...

source/grammar/qasm3.g4
aspcompiler: +1 for parentheses to remove the ambiguity with indexIdentifier: Identifier ( LBRACKET expressionList RBRACKET )?

blakejohnson: I would be okay making parentheses mandatory.

zachschoenfeld33 (approved):
LGTM

taalexander (approved):

PR #222, opened 2021-06-04
Change *List in Antlr grammar from left recursive to right recursive grammar

aspcompiler:
Currently, in qasm3.g4, we have something list grammar like:

```
quantumArgumentList
    : ( quantumArgument COMMA )* quantumArgument
    ;
```

Suggest to change to:
```
quantumArgumentList
    : quantumArgument ( COMMA quantumArgument )* 
    ;
```

This is because Antlr is an LL(*) parser. It performs better when we minimize number of look-aheads. In the old grammar, the parser does not know whether to match `( quantumArgument COMMA )*` or `quantumArgument` until it sees COMMA. In the new grammar, the parser will match `quantumArgument` first and then determine whether it needs to match `( COMMA quantumArgument )*` with a single look-ahead (LL(1)).

This applies to the follow lists:
- quantumArgumentList
- classicalArgumentList
- indexIdentifierList
- expressionList

### DIFF
--- a/source/grammar/qasm3.g4
+++ b/source/grammar/qasm3.g4
@@ -80,7 +80,7 @@ quantumArgument
     ;
 
 quantumArgumentList
-    : ( quantumArgument COMMA )* quantumArgument
+    : quantumArgument ( COMMA quantumArgument )* 
     ;
 
 /** Classical Types **/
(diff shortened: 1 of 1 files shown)


blakejohnson (approved):

jwoehr:
Whoever made the last edit on this added a trailing blank to each line of the 4 changes.
I know, I know, I'm just being picky :)

stevenheidel:
Thanks @aspcompiler !

PR #225, opened 2021-06-08
Adding Input/output grammar and live spec description

shiyunon:
In this PR, we added
* the section that describes input/output in the live specification.
* input/output grammar in qasm3.g4 which is properly tested.


blakejohnson (approved):
LGTM. Seems to accurately capture the context from the paper.

shiyunon:
The citation of VQE and 1st code example worked for me when I preview, somehow it's not rendering in the live spec.

PR #221, opened 2021-06-04
Add Working Group List

zachschoenfeld33:
This should be merged following #213.

Adds a list of current and former working groups (WG's) to the repository.

PR #232, opened 2021-06-22
Add note on ANTLR purpose

zachschoenfeld33:
### Summary

Add a note detailing the purpose and vision for the ANTLR grammar within OpenQASM3. This may evolve over time and details about new parsers will need to be added if they become reference implementations for OpenQASM3.


jwoehr:
Isn't the Qiskit Team, yourself, for instance, going to use precisely that grammar for Qasm 3 in Qiskit?

source/grammar/index.rst
```diff
@@ -3,6 +3,19 @@ OpenQasm 3.0 Grammar
 
 OpenQasm 3.0 Grammar specification based in ANTLR_ parser generator.
 
+The ANTLR grammar is intended to serve as the official reference grammar for OpenQASM3 and defines
+the set of syntactically valid statements in the language. ANTLR is used because it provides a
+human readable EBNF format which can evolve quickly. It further enables validation of the example
+files (see `openqasm/examples/`) via the built-in ANTLR parser. As an extension, it can be used
+to check whether a block of code is valid OpenQASM3.
```
taalexander, line 10:
```suggestion
The ANTLR grammar is intended to serve as the official reference grammar for OpenQASM3 and defines
the set of syntactically valid statements in the language. ANTLR is used because it provides a
human-readable EBNF format that is computer-validatable and is used to validate the example
files (see `openqasm/examples/`) using the generated ANTLR parser. As an extension, it can be used
to check whether a source file is parseable OpenQASM3 (note this does not ensure that the OpenQASM3 is semantically correct).
```

source/grammar/index.rst
```diff
@@ -3,6 +3,19 @@ OpenQasm 3.0 Grammar
...
+
+It should be noted, however, that the ANTLR grammar (and the associated parser) are not
+intended to be used in production. The ANTLR parser does not attempt to be performant or construct
+an AST. Furthermore, it does not conduct any semantic analysis.
+
+Thus, the ANTLR grammar should serve as a guide when developing compiler tooling for OpenQASM3, but
+should not itself be used as an OpenQASM parser.
```
taalexander, line 17:
```suggestion
Thus, the ANTLR grammar should serve as a guide when developing compiler tooling for OpenQASM3, but it
is not recommended to be used as the basis for a production-grade OpenQASM parser.
```

taalexander (changes requested):

zachschoenfeld33:
> Isn't the Qiskit Team, yourself, for instance, going to use precisely that grammar for Qasm 3 in Qiskit?

The grammar that ANTLR defines will be used, yes. But this is just an EBNF grammar. It defines the syntactic rules for the OpenQASM language. 

But the parser that comes along w/ ANTLR is not intended to be used in production. Production-grade parsers should be built using the ENBF grammar as defined in the `qasm3.g4` ANTLR file.

The ANTLR parser can of course be used to check if some OpenQASM code is parseable as is done in CI.

jwoehr:
> But the parser that comes along w/ ANTLR is not intended to be used in production. Production-grade parsers should be built using the ENBF grammar as defined in the `qasm3.g4` ANTLR file.

So what parser will OpenQASM 3 be using?
`pyparse` like OpenQASM 2?

taalexander:
> So what parser will OpenQASM 3 be using?
> pyparse like OpenQASM 2?

The backend software systems contain a closed source parser. Qiskit is working on its own parser, which as of now is using the ANTLR grammar and then transforming it into a more amenable AST. The scope of the responsibilities for OpenQASM3 need to be decided by the TSC, eg., will the language be only a specification (in which case the ANTLR grammar is just for validation/verification) or will it come with a reference implementation (in which case the ANTLR grammar or a replacement parser should be made more robust and practical).

jwoehr:
> will it come with a reference implementation (in which case the ANTLR grammar or a replacement parser should be made more robust and practical).

Sounds like a good idea.

> I think this proposal [#221 (comment)](https://github.com/Qiskit/openqasm/pull/221#issuecomment-862545336) is a great idea.

Yes, for instance, the expression stuff is too complicated. I have [a working OpenQASM 2 translator](https://github.com/Qiskit/qiskit-terra/pull/6657) based on a slightly hacked qasm3.g4 and I gave up on the expression parsing and parsed parameter expressions with a second grammar!

taalexander:
Do you mind if we merge the amendment in this PR and then revisit it as part of the grammar working group with a more long-term strategy?

1ucian0:
The TSC resolution about this PR is:
> Consensus seems to be that if people want to use the ANTLR grammar in a production setting, they are welcome to do so. We don’t need to claim otherwise in the spec.

This PR should be updated to reflect that.

taalexander (approved):

stevenheidel (approved):

PR #237, opened 2021-07-07
restoring source/grammar/tests/outputs/empty_gate.yaml

1ucian0:
One of the tests refers to the file `empty_gate.yaml`:

https://github.com/Qiskit/openqasm/blob/10c8f18ed5ded761ab926d5faede1bf32497ff02/source/grammar/tests/test_grammar.py#L123

However, this file does not exists, resulting in a test fail. i'm not fully sure when that file was removed and why. Adding it back.

### DIFF
--- a/source/grammar/tests/outputs/empty_gate.yaml
+++ b/source/grammar/tests/outputs/empty_gate.yaml
@@ -1,0 +1,17 @@
+# indent w/ 2 spaces
+source: |
+  gate g q {};
+reference: |
+  program
+    header
+    globalStatement
+      quantumGateDefinition
+        gate
+        quantumGateSignature
+          quantumGateName
+            g
+          identifierList
+            q
+        quantumBlock
+          {
+          }


aspcompiler:
The empty_gate.yrml was removed in [this commit](https://github.com/Qiskit/openqasm/commit/5a5789c3c5154d9d61cc8f550ac8ecd1da8fcc81) along with the reference to it in the test. The reference to it in the test came back after some merging. @zachschoenfeld33, should the test be removed or restored?

jakelishman:
In case this is merged later, I just wanted to point out that there's now a "logical" merge conflict even if not a real one; in #261 the reference YAML files moved into `tests/reference` rather than `tests/output` (since there are now more than one folder, having the more descriptive name was helpful).  So this file should be restored to `tests/reference/gate/empty_gate.yaml` now if we want to re-add it.  The new testing infrastructure will automatically discover it at pytest collection time; no need to add any lines to the test runner.

1ucian0:
The reference to the `empty_gate.yaml` was removed. Therefore, closing this PR

jakelishman:
Just for the future: the test in the old file was of a simple statement `gate g q {}` only.  The same statement is now part of a different reference test, so the behaviour is actually still tested:

https://github.com/Qiskit/openqasm/blob/135ec068007985bd2d26ff395e937d3c4d343c4e/source/grammar/tests/reference/gate/gate_modifiers.yaml#L3

PR #240, opened 2021-07-19
Add a vim plugin for syntax highlighting

jakelishman:
### Summary

Add a Vim plugin for syntax highlighting of OpenQASM 2.0 and 3.0 files.  For example:

Example syntax highlighting of an OpenQASM file with the solarized-dark color scheme Example syntax highlighting of an OpenQASM file with the solarized-light color scheme

### Details and comments

This is relatively basic for a plugin at the moment - it only does syntax highlighting, and not much else.  @levbishop suggested I put this here, but happy to make it its own repository if that would be better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ For previous version see: [2.0](https://github.com/Qiskit/openqasm/tree/OpenQASM
 
 ## About this project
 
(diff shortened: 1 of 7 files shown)


levbishop (approved):
Looks great. Thanks for this.

PR #251, opened 2021-08-12
Add CI tests for grammar via ANTLR parser

### LINKED ISSUE #250: Automated CI testing for components

Since the grammar has tests, we should get some automated CI running so at least we know we're not breaking them with changes.

jakelishman:
### Summary

This adds a CI run which builds the default ANTLR parser, and runs the
tests currently present in the directory, since it's easy to forget them
when making changes.  The pipeline is simple, and so is just run on the
free tier of GitHub Actions for now.

### Details and comments

This is currently just a very simple Action file, which prepares the environment and then runs the tests exactly as written.  I will likely follow up this PR with one separating out the reference parser from the tests of the it (but not changing it in any way, since the TSC are discussing what exactly will become the reference parser), and add some tests that _invalid_ QASM3 fails to parse, and things like this.

Note that the actual tests and reference parser will likely change drastically in the near future - this PR is just about getting CI up and running in the interim.

Fix #250.

### DIFF
--- a/.github/workflows/tests-grammar.yml
+++ b/.github/workflows/tests-grammar.yml
@@ -1,0 +1,32 @@
+name: Tests (Grammar)
+
+on:
+  [push, pull_request]
+
+jobs:
+  tests:
+    name: ANTLR grammar tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+      - name: Install ANTLR4
+        run: |
+          sudo apt install antlr4
+      - name: Install Python dependencies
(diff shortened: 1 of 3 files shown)


levbishop (approved):

taalexander:
It seems we now have some duplication with Travis - https://github.com/Qiskit/openqasm/blob/master/.travis.yml. We should choose one or the other.

jakelishman:
I missed that there was a Travis configuration file, because those tests aren't actually being run on CI - I assume this is related to travis-ci.org shutting down in mid-June.  This Action is equivalent to the Travis action, so we can just clear that out, now, since with Travis' new pricing model, it's easier for us to expand GH Actions workflows instead of Travis ones (unless we desperately need aarch64 machines, or so).

PR #252, opened 2021-08-12
Remove incrementors from language specification

### LINKED ISSUE #144: increment and decrement unary operator (++ and --) should be removed
Increment operator (`++` and `--`) is allowed for integer variables according to [the spec](https://qiskit.github.io/openqasm/language/classical.html#integers). But no description exists whether it is post-increment or pre-increment.

> Integer types support addition +, subtraction -, multiplication *, and integer division 2 /; the corresponding assignments +=, -=, *=, and /=; as well as increment ++ and decrement --.

In the grammer, only post-increment is defined as an expression.
```
expression
    : expression binaryOperator expression
    | unaryOperator expression
    | expression incrementor
   ...
```
Because it is an expression, `a=i++;` is accepted.

jakelishman:
### Summary

This removes the increment-and-return operators '++' and '--' from the
reference text and grammar.  The text did not specify whether these
should be pre- or post-incrementors, though only post-incrementors were
allowed by the ANTLR grammar.

The TSC conensus is that while these operators can easily be defined
unambiguously by parsers, in practice they are less readable by humans
than the equivalent `x += 1;` statement, and there is no significant
benefit to having them in OpenQASM 3.

### Details and comments

If we want to merge the CI PR (#251), I'll rebase this over it once it's in to show that the new tests pass.

Fix #144.

### DIFF
--- a/examples/msd.qasm
+++ b/examples/msd.qasm
@@ -160,10 +160,10 @@ distill_and_buffer(buffer_size) workspace, buffer;
 h q[0];
 cx q[0], q[1];
 Ty(address) q[0], buffer;
-address++;
+address += 1;
 cx q[0], q[1];
 Ty(address) q[1], buffer;
-address++;
+address += 1;
 
 // In principle each Ty gate can execute as soon as the magic
 // state is available at the address in the buffer register.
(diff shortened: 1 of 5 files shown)


levbishop (dismissed):

jakelishman:
Rebased over #251, so now the new tests will run and verify that the spec is updated, and the examples still successfully parse.  I want to add some tests of _failing_ behaviour as well across the language, so we can test that things which should be invalid at parse time really are forbidden by the grammar.  That's a different PR, though.

levbishop (approved):
Thanks

PR #261, opened 2021-08-20
Refactor tests and reference parser package

jakelishman:
### Summary

- Split out the "reference parser" logic from the "test" logic
- Make file tests use parametrised file discovery, rather than hard-coding file names.
- Add tests that _incorrect_ QASM programmes fail to parse.

### Details and comments

This PR refactors the `test_grammar.py` file, separating out the parser logic into a separate, buildable Python package.  This does not add any features to the "reference" parser, it just separates out the parsing logic from the testing logic to make things easier to handle, and wraps it up into a proper package so we can avoid all the hard-coded `sys.path` manipulations.  I have no intention to develop the reference parser further, since that's a much larger discussion.

I separated out the tests to be parametrised on the file name, and made these discover all tests in a directory, rather than relying on hard-coded values.  This means that all tests will now run, even if an early one fails - that helps with debugging if there are multiple problems, and `pytest` will clearly report the failing file name as well.

In addition to the tests of successful behaviour, I've also added some tests that various _invalid_ statements generate parse errors.  This should help in developing the grammar as well; in writing them, I already spotted a couple of places where the grammar allows things that it probably shouldn't (e.g. #260), and I imagine there are more lurking.


source/grammar/tests/test_grammar.py
```diff
@@ -1,173 +1,232 @@
...
+    "filename",
+    find_files(REPO_DIR / "examples", suffix=".qasm"),
+)
+def test_examples_parse(filename):
+    """Test that the example QASM3 files all parse without error."""
+    openqasm_reference_parser.pretty_tree(file=filename)
+
+
+class TestInvalidProgramsFailToParse:
+    valid_globals = [
```
taalexander, line 60: Would it be possible somehow to check these valid/invalid cases as code not tied to Python/pytest? I understand you did it this way to avoid creating `n` files, but this ties these test cases to our test framework?

Ways around this I can think of:
- We could create Yaml files with these cases
- We could create `n` files and just bite the bullet.
- We could use something like llvm-lit + FileCheck, this would add a significant amount of rework, but would enable a more flexible test suite in the longterm. I think this is too much work.
- Leave it as is 😄 .

taalexander (commented):
This looks great. A few non-blocking comments.

source/grammar/openqasm_reference_parser/tools.py
```diff
@@ -0,0 +1,43 @@
...
+
+from . import Qasm3ParserError
+from .qasm3Lexer import qasm3Lexer
+from .qasm3Parser import qasm3Parser
+
+__all__ = ["pretty_tree"]
+
+
+def pretty_tree(program: str = None, file: str = None) -> str:
+    if program is not None and file is not None:
```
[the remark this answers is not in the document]
jakelishman: oh yeah, duh, I forgot about those.  I'll add them.

source/grammar/tests/test_grammar.py
jakelishman: Perhaps I'll separate them out into separate files, but organised into the groups that I've done it here.  Then I'll have the parametrisation routine split the files by line, ignoring comments?

The reason for not having a separate file for every single line is because that would end up thrashing the disk IO, which is slow (though admittedly in our ~15s test suite, not a _huge_ deal), and also because it's hard to glance over what's being tested if there's a million little files.  I agree that it'd be best to separate them out into some easier files, though.

jakelishman:
@taalexander: latest commit has all the changes you suggested.  I think the way I've handled the parametrisation of all the single-statement cases (separate file for each group of tests, and split the lines into test cases) is a sensible trade-off between readability, ease of adding more cases, re-use in different suits, and not thrashing the disk IO, but that's the most major change, so best to have another look at it.

jakelishman:
Writing the "bad" cases files is pretty funny, because it (obviously) sends my syntax-highlighting plugin haywire!

taalexander (approved):
Thank you for the changes @jakelishman, this looks much better.

PR #267, opened 2021-08-30
Remove fixed type from openqasm 3

mbhealy:
### Summary
The Types and Casting Work Group ran into many problems when trying to define specific casting rules related to the fixed-point type and conversions to other types. We then stepped back and considered whether the fixed-point type `fixed` was even really necessary. We have yet to find an example of a significant problem that it solves that cannot be worked around with other language features.

### Details and comments

This PR removes the `fixed` type from the specification, grammar, tests, and examples.

### DIFF
--- a/examples/vqe.qasm
+++ b/examples/vqe.qasm
@@ -19,9 +19,9 @@ extern get_parameter(uint[prec], uint[prec]) -> angle[prec];
 extern get_npaulis() -> uint[prec];
 extern get_pauli(int[prec]) -> bit[2 * n];
 
-// The energy calculation uses fixed point division,
+// The energy calculation uses floating point division,
 // so we do that calculation in an extern function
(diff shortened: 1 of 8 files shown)


taalexander (commented):
If no major counter example can be found this is something I am very much in favour of.

jakelishman:
Aww, thanks for taking care of the syntax highlighter as well!

I don't know about anyone else, but I can't think of any realistic use-cases where you wouldn't be able to use a combination of `float`, `int` and (possibly) `duration` (`dt`, mostly) to achieve compatible results.

levbishop (dismissed):

blakejohnson (dismissed):

stevenheidel:
I believe this was approved by the TSC, pending @blakejohnson review

blakejohnson:
> I believe this was approved by the TSC, pending @blakejohnson review

I thought I had already approved it?

blakejohnson (approved):

PR #268, opened 2021-08-31
Specifics for casting classical types

mbhealy:
### Summary
This PR represents the first round of work from the Types and Casting WG. It contains specifics for the expected behavior when casting between classical types.

### Details and comments
Many mentions are made to slicing being required when there is precision mismatch in a casting operation. Slicing syntax is going to be discussed soon by the same WG, but has not been included in this PR.

Pursuant to PR #267 I have not included anything related to the `fixed` type.

I also modified the `duration` type to allow negative values, as the WG has discussed. I added the capability to divide durations by other durations, floats, and consts, this returns a float value. This is the alternative to allowing casting of durations, which we weren't able to come up with a well-defined behavior for.


source/language/delays.rst
```diff
@@ -125,8 +125,11 @@ these side effects. Also contrary to TeX, we prohibit overlapping gates.
 Operations on durations
 -----------------------
 
-We can add/subtract two durations, or multiply them by a constant, to get new
-duration. The result must be positive. These are compile time operations since ultimately all
+We can add/subtract two durations, or multiply them by a constant, to get a new
+duration. Division by compile time constants (either ``duration``, ``float``, or ``const``) results 
+in a machine-precision float (see :ref:`divideDuration`). Negative durations are allowed, however
```
jakelishman, line 130: I had thought it was that `duration / duration` was a float (since the dimensions are `time / time = 1`), but `duration / float` would be a `duration` (as `time / 1 = time`)?

source/language/types.rst
```diff
@@ -373,3 +375,152 @@ variables whose values may only be known at run time.
...
++--------------+------------+------------+------------+-------------+------------+------------+------------+------------+
+
+\*Note: ``duration`` values can be converted to ``float`` using the division operator. See :ref:`divideDuration`
+
+Casting from bool
+~~~~~~~~~~~~~~~~~
+
+``bool`` values cast from ``false`` to ``0.0`` and from ``true`` to ``1.0`` or 
+an equivalent representation. ``bool`` values can only be cast to ``bit[1]`` 
+(a single bit), so explicit slicing syntax must be given if the target ``bit``
```
jakelishman, line 458:
```suggestion
(a single bit), so explicit indexing syntax must be given if the target ``bit``
```

a slice would imply more than one bit, I think?

source/language/types.rst
```diff
@@ -373,3 +375,152 @@ variables whose values may only be known at run time.
...
+has more than 1 bit of precision.
+
+Casting from int/uint
+~~~~~~~~~~~~~~~~~~~~~
+
+``int[n]`` and ``uint[n]`` values cast to the standard types mimicking ``C``
+behavior. Casting to ``bit[m]`` is only allowed when ``m==n``. If the target 
+``bit`` has more or less precision, then explicit slicing syntax must be given.
```
jakelishman, line 466: Should we also bring the comment 2's-complement/LE into this section as well?

source/language/types.rst
```diff
@@ -373,3 +375,152 @@ variables whose values may only be known at run time.
...
+
+Casting from float
+~~~~~~~~~~~~~~~~~~
+
+``float[n]`` values cast to the standard types mimicking ``C`` behavior.
```
jakelishman, line 471: It might be better just to be explicit and say "discards the fractional part" (for integral types).  Did we decide on what happens if the integral part can't be represented by the new type?

Also:

```suggestion
``float[n]`` values cast to the standard types mimicking C behavior. 
```

source/language/types.rst
```diff
@@ -373,3 +375,152 @@ variables whose values may only be known at run time.
...
+
+Casting from float
+~~~~~~~~~~~~~~~~~~
+
+``float[n]`` values cast to the standard types mimicking ``C`` behavior. 
+Casting a ``float`` value to an ``angle[m]`` is accomplished by first 
+performing a modulo 2π operation on the float value. The resulting value 
+is then converted to the nearest ``angle[m]`` possible. In the event of a 
+tie between two possible nearest values the result is the one with an even 
+least significant bit (*i.e* round to nearest, ties to even).
```
jakelishman, line 476:
```suggestion
least significant bit (*i.e.* round to nearest, ties to even).
```

source/language/types.rst
```diff
@@ -373,3 +375,152 @@ variables whose values may only be known at run time.
...
+``n==m``, otherwise explicit slicing synatx must be provided. 
+
+When casting between angles of differing precisions (``n!=m``): if the target 
+has more significant bits, then the value is padded with ``m-n`` least 
+significant bits of ``0``; if the target has fewer significant bits, then 
+there are two acceptable behaviors that can be supported by compilers: 
+rounding and truncation. For rounding the value is rounded to the nearest 
+value, with ties going to the value with the even least significant bit.
+Trunction is likely to have more hardware support. This behavior can be
+controlled by the use of a ``#pragma``.
```
jakelishman, line 495: Should we talk with the implementation guys and try to decide on a roughlystandard syntax for the `#pragma` before merging this, or is it better to do later?  (While unknown pragmas can just be ignored, it may be better to try and nail down some of the basic ones first, even if just to decide on standard namespacing conventions.)

source/language/types.rst
```diff
@@ -373,3 +375,152 @@ variables whose values may only be known at run time.
...
+
+.. code-block:: c
+
+   const one_ns = 1e-9;
+   duration a = 500ns;
+   float a_in_ns = a / one_ns; // results in 500.0
+
+   float one = 1.0;
+   duration b = 1ms;
+   float b_float = b / one; // results in 1e-3
```
jakelishman, line 526: These examples seem wrong to me - I expected the first to result in `5e11ns` (or `500s`), and the second to result in `1ms`.  I thought what we'd said was just defining `duration / duration -> float`, with the idea that that avoided specifying a "natural" time unit (e.g. these ones treat the second as the natural unit, since `1ms / 1.0` goes to `1e-3`).

So I was expecting to have

```openqasm
duration one_ns = 1ns;
duration a = 500ns;
float a_in_ns = a / one_ns;  // 500.0

duration one_s = 1s;
float a_in_s = a / one_s; // 5e-7
```

and then `duration / float -> duration`, like "normal" division.

source/language/types.rst
```diff
@@ -373,3 +375,152 @@ variables whose values may only be known at run time.
...
+
+``bool`` values cast from ``false`` to ``0.0`` and from ``true`` to ``1.0`` or 
+an equivalent representation. ``bool`` values can only be cast to ``bit[1]`` 
+(a single bit), so explicit slicing syntax must be given if the target ``bit``
+has more than 1 bit of precision.
+
+Casting from int/uint
+~~~~~~~~~~~~~~~~~~~~~
+
+``int[n]`` and ``uint[n]`` values cast to the standard types mimicking ``C``
```
jakelishman, line 464:
```suggestion
``int[n]`` and ``uint[n]`` values cast to the standard types mimicking C
```

jakelishman (commented):
Thanks for writing it!  It looks good, and the table is helpful.  (Surprised we need jQuery to get the backgrounds, but that's a complete non-issue, really.)

I left a couple of very minor wording changes, and a few questions as to whether we should include a little more information in places, but those don't necessarily need action - they're just queries.  My main point is about the `duration` division thing - I had thought the result of the discussion was to define `duration / duration -> float` and leave `duration / float -> duration`.  Certainly division by float seems much more logical to me to produce a `duration` not a `float`, when looking at written-down examples - if it produces a float, it would be weird to me `duration * float` works "as expected" mathematically but `duration / float` changes the type.

source/language/types.rst
mbhealy: Fixed in https://github.com/Qiskit/openqasm/pull/268/commits/93e759e1cb14a14c5de6fdae55ffd33ae963ecbf

source/language/types.rst
mbhealy: Added in https://github.com/Qiskit/openqasm/pull/268/commits/93e759e1cb14a14c5de6fdae55ffd33ae963ecbf.

In line with C behavior, I believe it's the same behavior as casting from float to a large enough int, then casting from the large int to the target type.

source/language/types.rst
mbhealy: Fixed in https://github.com/Qiskit/openqasm/pull/268/commits/93e759e1cb14a14c5de6fdae55ffd33ae963ecbf

source/language/types.rst
mbhealy: Changed in https://github.com/Qiskit/openqasm/pull/268/commits/93e759e1cb14a14c5de6fdae55ffd33ae963ecbf

source/language/types.rst
mbhealy: This is a good point. I remember that part of the conversation now that you point it out. I've corrected the example to follow your suggestion above in https://github.com/Qiskit/openqasm/pull/268/commits/93e759e1cb14a14c5de6fdae55ffd33ae963ecbf

source/language/types.rst
mbhealy: Added in https://github.com/Qiskit/openqasm/pull/268/commits/93e759e1cb14a14c5de6fdae55ffd33ae963ecbf

source/language/types.rst
jakelishman: That shakes out as being modulo `UINTX_MAX + 1`, for unsigned, right?  That's sensible.  I think conversion to signed integer types in C is implementation-defined, though, if it doesn't fit - do we want to be prescriptive in that case?

mbhealy: You're right, it is implementation defined. I think it would make sense to apply the same approach to the OpenQASM 3 spec. I will add notes regarding this in the appropriate places.

mbhealy: Added in https://github.com/Qiskit/openqasm/pull/268/commits/fa1954277231cad03133dc4a1fc0f23b3ffe45c3

source/language/types.rst
ndebeaudrap: **N.B.** There is a detail that we fretted about, concerning _how_ to cast an ``angle`` to the nearest possible ``float``. There are two subtleties: the fact that this is discontinuous around the value 0 or 2π, and the issue of doing it quickly in practise given that ``angle`` types are fixed precision, in contrast to IEEE 754. We should figure out how/where to note these things.

(The first is not _really_ a big deal, but is something we should openly remark on, because it is an obvious pitfall for a programmer to be aware of. For the second, I'm now sure that there is a simple procedure for involving just a few machine instructions; it's a question of where we want to note that.)

mbhealy: Added in https://github.com/Qiskit/openqasm/pull/268/commits/8b7e89933298f4f3ffe4903e0d57743584aa9a1f

jakelishman: We left the question of `#pragma` til later and the grammar working group, so this doesn't need to be added here.

jwoehr:
PR reads:

> In general, for any cast between standard types that results in loss of 
precision, if the source value is larger than can be represented in the target 
type, the exact behavior is implementation specific.

Suggestion:

 > the exact behavior is implementation specific and must be documented by the vendor.

jwoehr:
> ``int[n]`` and ``uint[n]`` values cast to the standard types mimicking C
behavior.

This may need precise definition without waving at C, whose behavior in this regard has changed somewhat from time to time over 50 years.

jwoehr:
> ``float[n]`` values cast to the standard types mimicking C behavior (*i.e.* 
discarding the fractional part for integer-type targets).

Similarly, without waving at C. And [I believe _e.g._ is meant here](https://www.merriam-webster.com/words-at-play/ie-vs-eg-abbreviation-meaning-usage-difference) unless the only transformation in casting floats is the discarding of the fractional part, which does not appear to be the case here as floats can be cast to angle.

jwoehr:
Why can't an int or uint be cast to an angle, and back again?

mbhealy:
> Why can't an int or uint be cast to an angle, and back again?

Angles are interpreted as a stored value multiplied by 2pi, The stored value therefore exists in the range `[0,1.0)`, that value will always cast to 0 as an int/uint. Therefore, we decided that the cast made no sense.

mbhealy:
> > `float[n]` values cast to the standard types mimicking C behavior (_i.e._
> > discarding the fractional part for integer-type targets).
> 
> Similarly, without waving at C. And [I believe _e.g._ is meant here](https://www.merriam-webster.com/words-at-play/ie-vs-eg-abbreviation-meaning-usage-difference) unless the only transformation in casting floats is the discarding of the fractional part, which does not appear to be the case here as floats can be cast to angle.

Good point, I have updated this to specify the C99 standard behavior in https://github.com/Qiskit/openqasm/pull/268/commits/d4b1f820c6938236070489ccdf18535861ca5ff9

I would like to note that `angle` is an OpenQASM special type (along with `duration`, `bit`, and `qubit`), and not a 'standard' type, which consists of `bool`, `int`, `uint`, and `float`. Perhaps we should document this difference in the spec?

jwoehr:
> Good point, I have updated this to specify the C99 standard behavior in [d4b1f82](https://github.com/Qiskit/openqasm/commit/d4b1f820c6938236070489ccdf18535861ca5ff9)

Noted, and excellent.

> I would like to note that `angle` is an OpenQASM special type (along with `duration`, `bit`, and `qubit`), and not a 'standard' type, which consists of `bool`, `int`, `uint`, and `float`. Perhaps we should document this difference in the spec?

A concise note cannot hurt!

mbhealy:
> > the exact behavior is implementation specific and must be documented by the vendor.

Another good point, added in https://github.com/Qiskit/openqasm/pull/268/commits/0a839add6631bdb2307b482aee5b478a92f10400

mbhealy:
> > I would like to note that `angle` is an OpenQASM special type (along with `duration`, `bit`, and `qubit`), and not a 'standard' type, which consists of `bool`, `int`, `uint`, and `float`. Perhaps we should document this difference in the spec?
> 
> A concise note cannot hurt!

Added in https://github.com/Qiskit/openqasm/pull/268/commits/f11982d1cfae74968d321ecd47e8a923c7dab1ed

source/language/types.rst
```diff
@@ -290,9 +291,10 @@ Duration
 ~~~~~~~~
 
 We introduce a ``duration`` type to express timing.
-Durations are positive numbers with a unit of time. ``ns, μs, ms, s`` are used for SI time
+Durations are numbers with a unit of time. ``ns, μs, ms, s`` are used for SI time
```
stevenheidel, line 294: We should also allow "us" as a synonym for "μs"

source/language/types.rst
```diff
@@ -373,3 +375,173 @@ variables whose values may only be known at run time.
...
+
+``bool`` values cast from ``false`` to ``0.0`` and from ``true`` to ``1.0`` or 
+an equivalent representation. ``bool`` values can only be cast to ``bit[1]`` 
+(a single bit), so explicit index syntax must be given if the target ``bit``
+has more than 1 bit of precision.
+
+Casting from int/uint
+~~~~~~~~~~~~~~~~~~~~~
+
+``int[n]`` and ``uint[n]`` values cast to the standard types mimicking C99
```
stevenheidel, line 473: Can we be explicit about the int -> bool cast here? Similar to the other types

source/language/types.rst
```diff
@@ -17,6 +17,7 @@ override a reserved identifier.
 In addition to being assigned values within a program, all of the classical
 types can be initialized on declaration. Any classical variable or Boolean that is not explicitly
 initialized is undefined. Classical types can be mutually cast to one another using the typename.
+See :ref:`castingSpecifics` for more details on casting.
```
stevenheidel, line 20: Maybe I missed this, but is there a specific casting syntax? Or do you initialize a new variable with that type?

Similarly, do casts happen automatically when the wrong type is returned from a function?

mbhealy: Casting syntax is listed in the current spec, see the example at the end of [Integers](https://qiskit.github.io/openqasm/language/types.html#integers)
It looks like:
```
type1 identifier1 = initialization;
type2 identifier2 = type2(identifier1);
```

The `type()` syntax is the casting syntax

mbhealy: > Similarly, do casts happen automatically when the wrong type is returned from a function?

This counts as implicit casting, and we are working on defining rules for that now. It is explicitly disallowed in the current spec as far as I understand.

source/language/types.rst
mbhealy: Added in https://github.com/Qiskit/openqasm/pull/268/commits/6088c0267f31a9a56f6693e4fb58609f41ed46e2

source/language/types.rst
mbhealy: Added in https://github.com/Qiskit/openqasm/pull/268/commits/00bf4e712165aed383ad88687728f568fc2bf666

mbhealy:
I tried to address some inconsistencies in the `durationof()` syntax in https://github.com/Qiskit/openqasm/pull/268/commits/49949166d6281f7ff3321b9f8c999f18e0fd28c6, https://github.com/Qiskit/openqasm/pull/268/commits/324755c3e21b1b43f315eebb866b287e766391fd, and https://github.com/Qiskit/openqasm/pull/268/commits/fbc8b45bec04319bfc36486bb3d21f3606aada23

AbeerVaishnav13:
> > Why can't an int or uint be cast to an angle, and back again?
> 
> Angles are interpreted as a stored value multiplied by 2pi, The stored value therefore exists in the range `[0,1.0)`, that value will always cast to 0 as an int/uint. Therefore, we decided that the cast made no sense. We think it is better for intermediate types to be used to manipulate angle values using int/uint logic. For example, an angle value can be cast to a bit of the same width to get an exact bit-level copy, which can then be cast to an int/uint.
> 
> One could argue that the value in the range (0,2*pi] could be cast to an int in the range (0,6), but this seemed to have dubious utility to the WG.

Hi @mbhealy , This discussion got me interested and I was curious that if the `angle` type is stored as a "fraction-only" value in the implementation, then would the bit-wise operations be valid on an `angle` type variable?

For example, in the file `examples/ipe.qasm`, we use the the variable `angle[10] c;` in the following two contexts:
1. **As a bit string:** While storing the measurement result (`measure q -> c[0];`), and performing left_shift (`c  <<= 1;`)
2. **As an angle in range [0,2π):** While calling the `phase(c)` gate.

What I see in the above example is that the semantics and the value represented by the variable `c` seem different for different contextual usages in the code. So is this kind of "context-based" value of `angle` type correct to use? And if it is, then can't one also allow casting the `angle` type to other classical types based on the context in which it is used?

mbhealy:
> Hi @mbhealy, I was working on typing system for our project for the [Qiskit Advocate Mentorship Program 2021](https://github.com/qiskit-advocate/qamp-fall-21/issues/23), to write a reference `QASM3 -> QuantumCircuit` object translator. During working on the `complex` type, I noticed that the [specification](https://github.com/Qiskit/openqasm/blob/00bf4e712165aed383ad88687728f568fc2bf666/source/language/classical.rst#complex-numbers) doesn't allow power operator (`**`) on `complex` types. Is there a specific reason for not supporting that?
> 
> Thanks!

Hi @AbeerVaishnav13 
That's a good question. I suppose that the original motivation for the `complex` type was to represent samples in waveforms for the OpenPulse dialect. https://qiskit.github.io/openqasm/language/openpulse.html
For that use case I don't see a significant requirement for supporting the `**` operator. Additionally, I can imagine that the execution time of such statements could quickly become problematic if done at runtime (compile time should be fine assuming they are all compile-time resolved).

Do you have any examples of significant use cases we should consider that justify adding it to the spec?

jwoehr:
> I don't see a significant requirement for supporting the `**` operator. Additionally, I can imagine that the execution time of such statements could quickly become problematic if done at runtime (compile time should be fine assuming they are all compile-time resolved). 
> Do you have any examples of significant use cases we should consider that justify adding it to the spec?

Possibly the use case of "not encouraging programmers to perform undocumented calculations on a scratch pad to arrive at values for their written OpenQASM program". If it's a perfectly reasonable sort of calculation, it should not be excluded from the language where so many other math operations are being spooned in. A restriction to compile-time might well be reasonable.

source/language/types.rst
```diff
@@ -373,3 +375,174 @@ variables whose values may only be known at run time.
...
+Casting from angle
+~~~~~~~~~~~~~~~~~~
+
+``angle[n]`` values cast to ``bool`` using the convention ``val != 0.0``. 
+Casting to ``bit[m]`` values is only allowed when 
+``n==m``, otherwise explicit slicing synatx must be provided.
+Casting to ``float[m]`` is accomplished by first converting the angle value 
+to the nearest float, with ties going to the value with the even least 
+significant bit, and then multiplying the resulting value by 2π as a 
+``float[m]`` value. Note: the ring-like behavior of the angle type can
```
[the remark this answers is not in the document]
mbhealy: Can you suggest some syntax that would enable specifying what type of cast the user wants? Should we use a pragma? Can each cast be different, or is it a global setting?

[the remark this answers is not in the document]
ndebeaudrap: I would suggest syntax along the lines of ``angleToSignedFloat(angle a)`` and ``angleToUnsignedFloat(angle a)``, or possibly ``angleToFloatBelow(angle a, float u)`` to obtain a result ``x`` satisfying ``u - 2*π <= x < u``. (These names are perhaps not the best possible choices; I just mean to convey the meaning.) I have become wary of the idea of allowing the syntax of ``(float)a``, because it suggests a single standard approach to conversion, and I no longer believe that this is a good idea. Just conceivably one might want to allow a pragma to allow the programmer to fix the meaning of ``(float)a`` to some function as above using a pragma, but I am not even sure that this is a good idea.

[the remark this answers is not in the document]
ndebeaudrap: If syntax such as ``(signed)a`` or ``(unsigned)a`` were permissable, this might represent a good choice; though it suggests the existence of types which don't exist. ``signedf(a)`` or ``unsignedf(a)`` might be closer to a good choice of syntax.

source/language/delays.rst
```diff
@@ -125,14 +125,17 @@ these side effects. Also contrary to TeX, we prohibit overlapping gates.
 Operations on durations
 -----------------------
 
-We can add/subtract two durations, or multiply them by a constant, to get new
-duration. The result must be positive. These are compile time operations since ultimately all
+We can add/subtract two durations, or multiply or divide them by a constant, to get a new
+duration. Division of two durations results in a machine-precision float 
+(see :ref:`divideDuration`). Negative durations are allowed, however
+passing a negative duration to a ``gate[duration]`` or ``box[duration]`` expression will result in an error.
+All durations are compile time operations since ultimately all
```
blakejohnson, line 132:
```suggestion
All operations on durations happen at compile time since ultimately all
```

source/language/types.rst
```diff
@@ -373,3 +375,174 @@ variables whose values may only be known at run time.
...
+
+   <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.7.1/jquery.min.js"></script>
+   <script>
+     $(document).ready(function() {
+       $('.center').parent().addClass('center-parent');
+     });
+   </script>
+   <style>
+      .center-parent {text-align: center}
+   </style>
```
blakejohnson, line 436: What is all this for?

source/language/types.rst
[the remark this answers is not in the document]
blakejohnson: Interesting idea, @ndebeaudrap

blakejohnson (commented):
Aside from the mysterious jquery stuff that has snuck into this PR, I would be okay with merging this as-is. I think @ndebeaudrap has a good question about other angle -> float conversions which could be addressed in a follow-up PR.

source/language/types.rst
mbhealy: This was the only way I could find to add colors to the casting tables, which I felt was very important for readability.

source/language/delays.rst
mbhealy: Updated in https://github.com/Qiskit/openqasm/pull/268/commits/b3f3b65e875b47cb329e6c2af1fcb4b6601f8fcf

source/language/types.rst
mbhealy: The three styles this adds are red-background (rbg) green-background (gbg) and centered.

source/language/types.rst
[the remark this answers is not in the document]
ndebeaudrap: @blakejohnson We don't _have_ to define what ``float(a)`` means for ``angle a``, or at least not any more than we have to define what ``angle(b)`` means for ``bool b``. We can decide that it does not make sense, after all, to allow this as a typecast _as such_. What I advocate is to disallow the typecast, but instead provide functions (which may be evaluated in some instances at compile time) which take an ``angle`` as input and produce a ``float``, according to various specifications.

Having said that: your proposal allowing ``float(...)`` to take additional arguments is interesting. I would suggest that we should be extremely careful of allowing these arguments to be optional, in the case that the first argument is to be of type ``angle``. If we were to take this approach, I would advocate syntax along the lines of ``float(angle a, float ivlCenter)``, where the second argument is mandatory and indicates the center of a half-open interval of width 2π. For instance, ``float(a, 0)`` might produce a float in the range [-π,π), while ``float(a,π)`` might produce a float in the range [0,2π).

jwoehr:
I thought you could change colors in .rst files with css styles.

mbhealy:
> I thought you could change colors in .rst files with css styles.

I would appreciate if anyone can suggest an easy way to do this or a link to some resources. From what I can tell the Alabaster theme chose by the TSC does not have built-in support for changing colors in a way that seems useful to me. I could include other things that add support, but it's not clear to me that they are superior to the method I was able to get working.

jwoehr:
> > I thought you could change colors in .rst files with css styles.
> 
> I would appreciate if anyone can suggest an easy way to do this or a link to some resources. 

@mbhealy you may be right ... playing with this using some googled css recipes ... see what I can find :)

jwoehr:
@mbhealy I was able to change the words  **extended language** in the first sentence of `source/language/index.rst` to purple using css.

Probably the technique is applicable in general to any css task.

I added the file `purple.css` to the `source/_static directory`:

```
.purple {
    color: purple;
}
```

And here is the diff for everything else I changed:
```
diff --git a/source/conf.py b/source/conf.py
index fa6838b..de747c0 100644
--- a/source/conf.py
+++ b/source/conf.py
@@ -64,3 +64,5 @@ html_baseurl = os.getenv('HTML_BASEURL', '')
 numfig = True
 # Necessary setting for sphinxcontrib-bibtex >= 2.0.0
 bibtex_bibfiles = ['bibliography.bib']
+
+html_css_files = ['purple.css']
\ No newline at end of file
diff --git a/source/language/index.rst b/source/language/index.rst
index 9942021..5c289f6 100644
--- a/source/language/index.rst
+++ b/source/language/index.rst
@@ -1,9 +1,10 @@
 .. _sec:spec:
+.. role:: purple
 
 Language
 ========
 
-OpenQASM refers to the extended language we now describe, specifically
+OpenQASM refers to the :purple:`extended language` we now describe, specifically
 OpenQASM Version 3.0.
 
 The human-readable form of OpenQASM is a simple, case-sensitive textual language.
```

And then `make html`.

jwoehr:
@mbhealy the techniques I derived from the following two StackOverflow pages:

- https://stackoverflow.com/questions/4669689/how-to-use-color-in-text-with-restructured-text-rst2html-py-or-how-to-insert-h
- https://stackoverflow.com/questions/32079200/how-do-i-set-up-custom-styles-for-restructuredtext-sphinx-readthedocs-etc

mbhealy:
> @mbhealy I was able to change the words **extended language** in the first sentence of `source/language/index.rst` to purple using css.

@jwoehr Thank you very much for the pointers. I was able to use the css version you pointed out to color the text in the same manner. Unfortunately the centering doesn't seem to be working correctly, but it still looks pretty good.

jwoehr:
> Unfortunately the centering doesn't seem to be working correctly, but it still looks pretty good.

@mbhealy I see that the custom css

```
table.align-center {
    margin-left: auto;
    margin-right: auto;
}
```
was recommended in [the last response to the 2017 Sphinx issue 3942](https://github.com/sphinx-doc/sphinx/issues/3942#issuecomment-318371518)

mbhealy:
> > Unfortunately the centering doesn't seem to be working correctly, but it still looks pretty good.
> 
> @mbhealy I see that the custom css
> 
> ```
> table.align-center {
>     margin-left: auto;
>     margin-right: auto;
> }
> ```
> 
> was recommended in [the last response to the 2017 Sphinx issue 3942](https://github.com/sphinx-doc/sphinx/issues/3942#issuecomment-318371518)

Yes, I attempted that but it did not work.

blakejohnson:
@mbhealy the revised styling seems much less intrusive in the doc. Thanks.

blakejohnson:
So, all that's left now is to resolve this discussion: https://github.com/Qiskit/openqasm/pull/268/files#r732623768

source/language/types.rst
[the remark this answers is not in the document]
mbhealy: The discussion at the TG meeting resolved in us deciding to remove the possibility of casting directly from angle to float from the spec for now. We will try to address it more fully in a future PR.

mbhealy:
> So, all that's left now is to resolve this discussion: https://github.com/Qiskit/openqasm/pull/268/files#r732623768

The TG decided in today's meeting to simply remove the possibility of casting angles to floats. Changed in https://github.com/Qiskit/openqasm/pull/268/commits/0427362d83dc04b91bef32cafa31aab99ffbc162

source/language/types.rst
```diff
@@ -373,3 +375,126 @@ variables whose values may only be known at run time.
...
+(a single bit), so explicit index syntax must be given if the target ``bit``
+has more than 1 bit of precision.
+
+Casting from int/uint
+~~~~~~~~~~~~~~~~~~~~~
+
+``int[n]`` and ``uint[n]`` values cast to the standard types mimicking C99
+behavior. Casting to ``bool`` values follows the convention ``val != 0``.
+As noted above, if the value is too large to be represented in the
+target type the result is implementation-specific. Casting to ``bit[m]``
```
jakelishman, line 438: In C99, `signed[n] -> unsigned[n]` conversion is completely defined, but `unsigned[n] -> signed[n]` is implementation-defined if the `unsigned` is too big.  Given that the C99 spec gives implementations a choice between 1's, 2's and sign/magnitude ints, but we seem to be strongly assuming 2's complement, we could define both directional casts as "the bits are re-interpreted as {an unsigned, 2's complement} integer" as appropriate?  That's consistent with `signed[n] -> unsigned[n]` as C99 defines it, and has the "expected" behaviour that `int[n](uint[n](x))` is guaranteed to have the same bit pattern as `x` for all `int[n] x`, without being implementation-defined.

I'm also happy to just leave it as "implementation-defined" for now, though. I'm mostly thinking that this is one point where, if we're requiring 2's complement integers, then there really is only one sensible way to do things (that I can think of, at least!), so being explicit might be more clear.  I suppose it depends if we expect any hardware vendors to _not_ use little-endian 2's complement.

source/language/types.rst
```diff
@@ -373,3 +375,126 @@ variables whose values may only be known at run time.
...
+is only allowed when ``m==n``. If the target 
+``bit`` has more or less precision, then explicit slicing syntax must be given.
+As noted, the conversion is done assuming a little-endian 2's complement 
+representation.
+
+Casting from float
+~~~~~~~~~~~~~~~~~~
+
+``float[n]`` values cast to the standard types mimicking C99 behavior (*e.g.* 
+discarding the fractional part for integer-type targets). As noted above,
```
jakelishman, line 448: We don't mention the `val != 0.0` convention for Boolean conversion here.  Not certain it's necessary to say it again (we could just move _all_ instances of it up to the top paragraph), but an interesting consequence of this convention is that `bool(nan) == true`, because `nan` compares unequal to everything.  That's consistent with C and Python, for sure.

source/language/types.rst
```diff
@@ -373,3 +375,126 @@ variables whose values may only be known at run time.
...
+Casting a ``float`` value to an ``angle[m]`` is accomplished by first 
+performing a modulo 2π operation on the float value. The resulting value 
+is then converted to the nearest ``angle[m]`` possible. In the event of a 
+tie between two possible nearest values the result is the one with an even 
+least significant bit (*i.e.* round to nearest, ties to even).
+
+Casting from angle
+~~~~~~~~~~~~~~~~~~
+
+``angle[n]`` values cast to ``bool`` using the convention ``val != 0.0``.
```
jakelishman, line 460: I'm not 100% sure that this provides enough utility to justify a casting rule, when `val != 0` should always work.  Maybe it's just me being nervous about the arithmetic on a ring again, though, so I'm happy to leave it if others think it's worthwhile.  Mostly it's that I'm so conditioned to expect that `bool(2 * x) == bool(x)` for _all_ `x`, but `angle` breaks that.  It's not the fault of the casting rule, though - the `val != 0` form also breaks that equality, but it just feels more explicit.

Possibly best leave it since it was already discussed, and just let me keep being nervous about the `angle` type to myself.

source/language/types.rst
```diff
@@ -373,3 +375,126 @@ variables whose values may only be known at run time.
...
+is then converted to the nearest ``angle[m]`` possible. In the event of a 
+tie between two possible nearest values the result is the one with an even 
+least significant bit (*i.e.* round to nearest, ties to even).
+
+Casting from angle
+~~~~~~~~~~~~~~~~~~
+
+``angle[n]`` values cast to ``bool`` using the convention ``val != 0.0``. 
+Casting to ``bit[m]`` values is only allowed when 
+``n==m``, otherwise explicit slicing synatx must be provided.
```
jakelishman, line 462:
```suggestion
``n==m``, otherwise explicit slicing syntax must be provided.
```

source/language/types.rst
```diff
@@ -373,3 +375,126 @@ variables whose values may only be known at run time.
...
+Casting from bit
+~~~~~~~~~~~~~~~~
+
+``bit[n]`` values cast to ``bool`` using the convention ``val != 0``. Casting to 
+``int[m]`` or ``uint[m]`` is done assuming a little endian 2's complement 
+representation, and is only allowed when ``n==m``, otherwise explicit slicing 
+syntax must be given. Likewise, ``bit[n]`` can only be cast to ``angle[m]`` 
+when ``n==m``, in which case an exact per-bit copy is done using little-endian 
+bit order. Finally, casting between bits of differing precisions is not 
+allowed, explicit slicing syntax must be given.
```
jakelishman, line 482: We don't specify the size of the smallest addressable unit here, so when we talk about conversions assuming a little-endian ordering, I think we're opening ourselves up to a possibility that `int[n](bit_string)` produces different values on different hardware.  For example, if I have the bitstring `x = "01_10"`, then on a "2-bit address" machine the conversion `uint[4](x)` produces the value 9, but a "4-bit address" machine would read it as 6.

I'm guessing we wrote it like this with the idea that the bitstrings could be directly re-interpreted, but that seems unusual to me; in general high-level programming, I expect the bitstring `0b1_0000_0010` to be the number 258, but by these conventions, on a machine that addresses things in 8-bit bytes, I'm not even sure what it would become.  I guess it would actually interpret my 4-bit separators differently, and read it as `0b1000_0001_0[000_000]`, which it then converts to an unsigned as 129?

Maybe I'm completely confused here - it seems like I might be making a very simple mistake.

jakelishman (commented):
In general I'm pretty happy with this, but I read through it again and got myself in a right tizzy about what happens when we're assuming "little-endian ordering" in bit-string conversions.  Could someone check and potentially point out the trivial mistake I'm making?

I left a couple of other comments about places we _could_ tighten the language/spec, but I'm happy to accept the text with or without them for now.

source/language/types.rst
ndebeaudrap: I don't think you have to be nervous about ``angleVal != 0`` for angle values; this corresponds to something thoroughly unambiguous to me at least (albeit the ``angle`` value of 0 is not identical to the real number 0). However, I think that from a legibility perspective, it would be preferable for the programmer to write ``angleVal != 0`` rather than ``(bool)angleVal``; so for completely independent reasons I would be happy with disallowing casts from ``angle`` to ``bool``.

jakelishman: Yeah, I'm not nervous about the actual meaning of that cast - `val != 0` is definitely unambiguous, and it does produce the results I'd expect, given any one single angle.  I think it's just that all my tools for reasoning about computer descriptions of numeric types is broken when I have to think about arithmetic on a ring as a feature, not "undefined behaviour".

source/language/types.rst
mbhealy: I think for little endian ordering the entire string has to be interpreted as little endian, not just addressable unit by addressable unit. It is the responsibility of the compiler to produce the correct behavior on any given piece of hardware, whether that is an illusion supported by software or not.

source/language/types.rst
mbhealy: I feel like it's reasonable to rely on the C99 spec being generally well understood in this regard and also easily accessible to programmers, but I'm willing to add it if other agree that it's better.

source/language/types.rst
jakelishman: Michael and I talked privately, and the upshot is that I had the terminology backwards in my head when talking about bitstring literals.  So it was a very simple mistake.

jakelishman (dismissed):
I'm happy to accept this PR as-is, regardless of my other comments.  Thank you for all the effort, Michael!

source/language/types.rst
jakelishman: Resolved by 14a9f19.

source/language/types.rst
jakelishman: Yeah, agreed.  Mostly I was just curious about what happened with `nan`!  The `val != 0` seems to be entirely consistent with other languages, which also have `bool(nan)` is true.

source/language/types.rst
mbhealy: Fixed in https://github.com/Qiskit/openqasm/pull/268/commits/b60fcff30e7f2e1442eebc8376b5e948cef021f8

blakejohnson (dismissed):

source/language/delays.rst
```diff
@@ -125,14 +125,17 @@ these side effects. Also contrary to TeX, we prohibit overlapping gates.
 Operations on durations
 -----------------------
 
-We can add/subtract two durations, or multiply them by a constant, to get new
-duration. The result must be positive. These are compile time operations since ultimately all
+We can add/subtract two durations, or multiply or divide them by a constant, to get a new
+duration. Division of two durations results in a machine-precision float 
+(see :ref:`divideDuration`). Negative durations are allowed, however
+passing a negative duration to a ``gate[duration]`` or ``box[duration]`` expression will result in an error.
+All operations on durations happen at compile time operations since ultimately all
```
ajavadia, line 132: I think everything in this PR looks fine to me except the duration part. I have some questions:
- Is there a need for casting durations to floats? What is an example?
- What does it even mean for durations to be cast to floats? durations have units, floats do not. Dropping the units makes the conversion semantics non-preserving.
- Even if there is a need for casting, do we need duration here? I think the "multiply by constant" in the original spec should cover that already since you can e.g. do `.5 * duration`

jakelishman: There is no cast of duration to float for the reasons you gave.  You can do the binary operation `duration / duration -> float` though, which doesn't have the unit problem, because it dimensionally reduces to a scalar, rather than something with dimensions of time.  That's for cases like manual dynamic decoupling, where say you want to do
```
n_reps = dur1 / dur2;
long_operation[dur1]($0, $1);
for i in [1:n_reps] {
    dynamic_decouple[dur2]($2);
}
```

(this section isn't on the same page as the casting - it's in the bit about delays)

jakelishman (approved):

source/language/delays.rst
ajavadia: ok i think i understand now. being able to compute the ratio of two durations sounds good to me.

ajavadia (approved):

ndebeaudrap (dismissed):

stevenheidel (approved):

1ucian0 (approved):

Issue #246, opened 2021-07-28
Add a new token to match the body of defcal


aspcompiler:
### What is the expected enhancement?

TLDR: Match defcal body with a token so that defcal grammars can use their own lexer rules.

Details: Currently, the defcal body is parsed with a parser rule rather than a lexer rule. Since most of modern parsers separates the lexing and parsing phases,  defcal body is tokenized by the OpenQASM lexer rules. To allow other defcal grammar to implement their own lexer rules, it is desirable to match the defcal body as a single token and pass the resulting string to the parser of the defcal grammar.

Currently, the def grammar is written as:

```
calibrationDefinition
    : 'defcal' Identifier
    ( LPAREN calibrationArgumentList? RPAREN )? identifierList
    returnSignature? LBRACE .*? RBRACE  // for now, match anything inside body
    ;
```

`.*?` inside LBRACE and RBRACE will match any OpenQASM tokens non-greedily; this mean the body cannot contain RBRACE unless it is enclosed in a string or comment. White space and comments are skipped due to the OpenQASM lexer rule.

To preserve the body of defcal for other parsers, consider introducing a special lexer rule such as

```
ExternalGrammar
  : '{{' .*? '}}'
```

Then the defcal grammar would become:

```
calibrationDefinition
    : 'defcal' Identifier
    ( LPAREN calibrationArgumentList? RPAREN )? identifierList
    returnSignature? ExternalGrammar
    ;
```

jakelishman:
A possible alternative to this is to use ANTLR's "island grammar" rules to swap the _lexing_ mode when the `defcal` token is encountered.  We can then have the lexer emit an entirely different set of tokens when it encounters braces; we could potentially use a lexeme made of fragments, which will allow all text with balanced braces to occur.  We'd probably want some way to embed a literal unbalanced brace in the text, if not only for completeness (handle `\{`, `\}` and `\\` as separate tokens?), and then restore them later during parse-tree simplification, before passing the body on to a separate lexer for `defcal` blocks.

Without the escape characters, this looks something approximately like:
```antlr
// switch to defcal mode when the defcal lexeme is encountered
defcal: 'defcal' -> mode(DEFCAL);

mode DEFCAL;
// ... tokens which can handle the defcal signature ...
// Consume a balanced brace block, and then return lexing back to QASM mode.
DefcalgrammarBody: DefcalgrammarBlock -> mode(DEFAULT_MODE);
[...]

PR #269, opened 2021-08-31
Reference implementation of OpenQASM 3.0 AST in Python

aspcompiler:
### Summary

A reference implementation of OpenQASM 3.0 AST in Python. It consists of:

* ast.py: The AST nodes.
* parser/antlr: A parser based on Antlr grammar and reference parser also found in this repo.
It walks the ANTLR parse tree to generate the AST tree.
* ast_visitor: A base AST visitor and an example showing how to implement compiler passes
using the visitor.
* tests: A set of unit tests.
* README.md on details for developers to build and test.

### Details and comments

1. It is built on Antlr parser. It uses the generated qasm3Visitor so it is necessary to pass -visitor option to `antlr` command to generate the visitor.
1. It uses `black` to format the code, `pylint` to check the style and `pytest` to run unit test. The configuration for `black` and `pylint` are in open_node/pyproject.toml and open_node/.pylintrc respectively.
1. It uses .github/workflows/tests-ast.yml to run CI/CD.

A few things to consider:
1. Is package naming/directory structure desirable?
1. Shall we combine with requirements*.txt under grammar or maintain separate ones.
1. Shall we combine with .github/workflows/tests-grammar.yml or maintain separate ones.

### DIFF
--- a/source/open_node/ast.py
+++ b/source/open_node/ast.py
@@ -1,0 +1,1017 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import List, Optional, Union
+from enum import Enum, auto
+
+
+@dataclass
+class Span:
+    """
+    Start and end line/column in the source file
+    We use the Antlr convention. The starting line number is 1 and starting column number is 0.
+    """
+
+    start_line: int
+    start_column: int
+    end_line: int
+    end_column: int
+
+
+@dataclass
+class OpenNode:
+    """Base class for all OpenQASM 3 nodes"""
+
+    span: Optional[Span] = field(init=False, default=None, compare=False)
+    """
+    The span(location) of the node in the source code.
+    Because not all the nodes are generated from source, the span is optional.
+    To make it easier to write unit test, we exclude span from the generated __eq__().
+    """
+
+
+@dataclass
+class Program(OpenNode):
+    """
+    An entire OpenQASM 3 program represented by a list of top level statements
+    """
+
+    statements: List[Statement]
+    version: dict[str, int] = field(init=False, default_factory=dict, compare=False)
+    io_variables: List[IODeclaration] = field(init=False, default_factory=list)
+
+
+class Statement(OpenNode):
+    """A statement: anything that can appear on its own line"""
+
+
+@dataclass
+class AliasStatement(Statement):
+    """
+    Alias statement
+
+    Example::
+
+        let a = qubits[0];
+
+    """
+
+    target: Identifier
+    value: Union[IndexIdentifier, Identifier]
+
+
+@dataclass
+class ExpressionStatement(Statement):
+    """A statement that contains a single expression"""
+
+    expression: Expression
+
+
+class QubitDeclTypeName(Enum):
+    qubit = auto()  # OpenQASM 3 style.
+    qreg = auto()  # OpenQASM 2 style.
+
+
+@dataclass
+class QubitDeclaration(Statement):
+    """
+    Global qubit declaration
+
+    Example::
+
+        qubit q;
+        qubit[4] q;
+
+        q // <- qubit
+        [4] // <- designator
+
+    """
+
+    decl_type_name: QubitDeclTypeName
+    qubit: Qubit
+    designator: Optional[Expression]
+
+
+@dataclass
+class QuantumGateDefinition(Statement):
+    """
+    Define a new quantum gate
+
+    Example::
+
+        gate cx c, t {
+            ctrl @ unitary(pi, 0, pi) c, t;
+        }
+
+    """
+
+    name: str
+    arguments: List[ClassicalArgument]
+    qubits: List[Identifier]
+    body: List[QuantumStatement]
+
+
+class QuantumStatement(Statement):
+    """Statements that may appear inside a gate declaration"""
+
+
+@dataclass
+class QuantumForInLoop(Statement):
+    """For In loops that contain only quantum statements."""
+
+    loop_variable: Identifier
+    set_declaration: Union[RangeDefinition, List[Expression], Identifier]
+    block: List[QuantumStatement]
+
+
+@dataclass
+class QuantumWhileLoop(Statement):
+    """While loops that contain only quantum statements."""
+
+    while_condition: Expression
+    block: List[QuantumStatement]
+
+
+@dataclass
+class ExternDeclaration(Statement):
+    """
+    A extern declaration
+
+    Example::
+
+        extern get_pauli(int[prec]) -> bit[2 * n];
+
+        get_pauli  // <- name
+        int[prec]  // <- classical type
+        bit[2 * n] // <- return type
+
+    """
+
+    name: str
+    classical_types: List[ClassicalType]
+    return_type: Optional[ClassicalType]
+
+
+@dataclass
+class Qubit(OpenNode):
+    """
+    A qubit
+
+    Example::
+
+        qubit q;
+
+        q  // <- Qubit
+
+    """
+
+    name: str
+
+
+class Expression(OpenNode):
+    """An expression: anything that returns a value"""
+
+
+UnaryOperator = Enum("UnaryOperator", "~ ! -")
+
+
+@dataclass
+class UnaryExpression(Expression):
+    """
+    A unary expression
+
+    Example::
+
+        ~b
+        !bool
+        -i
+
+    """
+
+    op: UnaryOperator
+    expression: Expression
+
+
+BinaryOperator = Enum("BinaryOperator", "> < >= <= == != && || | ^ & << >> + - * / % **")
+
+
+@dataclass
+class BinaryExpression(Expression):
+    """
+    A binary expression
+
+    Example::
+
+        q1 || q2
+
+    """
+
+    op: BinaryOperator
+    lhs: Expression
+    rhs: Expression
+
+
+ConstantName = Enum("ConstantName", "pi π tau 𝜏 euler ℇ")
+
+
+@dataclass
+class Constant(Expression):
+    """
+    A constant expression
+
+    Example::
+
+        π
+        𝜏
+        ℇ
+    """
+
+    name: ConstantName
+
+
+@dataclass
+class Identifier(Expression):
+    """
+    An identifier
+
+    Example::
+
+        q1
+
+    """
+
+    name: str
+
+
+@dataclass
+class IntegerLiteral(Expression):
+    """
+    An integer literal
+
+    Example::
+
+        1
+
+    """
+
+    value: int
+
+
+@dataclass
+class RealLiteral(Expression):
+    """
+    An real number literal
+
+    Example::
+
+        1.1
+
+    """
+
+    value: float
+
+
+@dataclass
+class BooleanLiteral(Expression):
+    """
+    A boolean expression
+
+    Example::
+
+        true
+        false
+
+    """
+
+    value: bool
+
+
+@dataclass
+class StringLiteral(Expression):
+    """
+    A string literal expression
+
+    Example::
+
+        'Hadamard gate'
+
+    """
+
+    value: str
+
+
+class TimeUnit(Enum):
+    dt = auto()
+    ns = auto()
+    us = auto()
+    µs = auto()
+    ms = auto()
+    s = auto()
+
+
+@dataclass
+class DurationLiteral(Expression):
+    """
+    A duration literal
+
+    Example::
+
+        1.0ns
+
+    """
+
+    value: float
+    unit: TimeUnit
+
+
+@dataclass
+class FunctionCall(Expression):
+    """
+    A function call expression
+
+    Example::
+
+        foo(1)
+
+        foo // <- name
+
+    """
+
+    name: str
+    arguments: List[Expression]
+
+
+@dataclass
+class IndexExpression(Expression):
+    """
+    An index expression.
+
+    This is used to represent the following unlabeled line in the grammar:
+
+    expressionTerminator
+        : ...
+        | expressionTerminator LBRACKET expression RBRACKET
+
+    Example::
+
+        q[1]
+
+        q // <- expression
+        1 // <- index_expression
+    """
+
+    expression: Expression
+    index_expression: Expression
+
+
+class QuantumInstruction(QuantumStatement):
+    """
+    Baseclass for quantum instructions.
+    """
+
+
+@dataclass
+class QuantumGate(QuantumInstruction):
+    """
+    Invoking a quantum gate
+
+    Example::
+        cx 0, 1;
+
+        or
+
+        ctrl @ p(λ) a, b;
+
+        ctrl @ // <- quantumGateModifier
+        p // <- quantumGateName
+        λ // <- argument
+        a, b // <- qubit
+    """
+
+    modifiers: List[QuantumGateModifier]
+    name: str
+    arguments: List[Expression]
+    qubits: List[Union[IndexIdentifier, Identifier]]
+
+
+class GateModifierName(Enum):
+    inv = auto()
+    pow = auto()
+    ctrl = auto()
+
+
+@dataclass
+class QuantumGateModifier(OpenNode):
+    """
+    A quantum gate modifier
+
+    Attributes:
+        modifier: 'inv', 'pow', or 'ctrl'
+        expression: only pow modifier has expression.
+
+    Example::
+
+        inv @
+        pow(1/2)
+        ctrl
+    """
+
+    modifier: GateModifierName
+    argument: Optional[Expression]
+
+
+@dataclass
+class QuantumPhase(QuantumInstruction):
+    """
+    A quantum phase instruction
+
+    Example::
+
+        ctrl @ gphase(λ) a;
+
+        ctrl @ // <- quantumGateModifier
+        λ // <- argument
+        a // <- qubit
+
+    """
+
+    quantum_gate_modifiers: List[QuantumGateModifier]
+    argument: Expression
+    qubits: List[Union[IndexIdentifier, Identifier]]
+
+
+@dataclass
+class QuantumMeasurement(QuantumInstruction):
+    """
+    A quantum measurement instruction
+
+    Example::
+
+        measure q;
+    """
+
+    qubits: List[Union[IndexIdentifier, Identifier]]
+
+
+@dataclass
+class QuantumReset(QuantumInstruction):
+    """
+    A reset instruction.
+
+    Example::
+
+        reset q;
+    """
+
+    qubits: List[Union[IndexIdentifier, Identifier]]
+
+
+@dataclass
+class QuantumBarrier(QuantumInstruction):
+    """
+    A quantum barrier instruction
+
+    Example::
+
+        barrier q;
+    """
+
+    qubits: List[Union[IndexIdentifier, Identifier]]
+
+
+@dataclass
+class QuantumMeasurementAssignment(Statement):
+    """
+    A quantum measurement assignment statement
+
+    Example::
+
+        c = measure q;
+    """
+
+    lhs: List[Union[IndexIdentifier, Identifier]]
+    measure_instruction: QuantumMeasurement
+
+
+@dataclass
+class ClassicalArgument(OpenNode):
+    """
+    Classical argument for a gate or subroutine declaration
+    """
+
+    type_: ClassicalType
+    name: str
+
+
+@dataclass
+class ClassicalDeclaration(Statement):
+    """
+    Classical variable declaration
+
+    Example::
+
+        bit c;
+    """
+
+    type: ClassicalType
+    identifier: Identifier
+    init_expression: Optional[Expression]
+
+
+class IOIdentifierName(Enum):
+    output = auto()
+    input = auto()
+
+
+@dataclass
+class IODeclaration(ClassicalDeclaration):
+    """
+    Input/output variable declaration
+
+    Exampe::
+
+        input angle[16] theta;
+        output bit select;
+    """
+
+    io_identifier: IOIdentifierName
+
+
+@dataclass
+class ConstantDeclaration(Statement):
+    """
+    Constant declaration
+
+    Example::
+
+        const n = 10;
+    """
+
+    identifier: Identifier
+    init_expression: Optional[Expression]
+
+
+class ClassicalType(OpenNode):
+    """
+    Base class for classical type
+    """
+
+
+@dataclass
+class SingleDesignatorType(ClassicalType):
+    """
+    Type with one designator, e.g, int, uint, float, angle, bit array
+
+    Example::
+
+        int[32]
+        uint[32]
+        float[32]
+        angle[32]
+    """
+
+    type: SingleDesignatorTypeName
+    designator: Optional[Expression]
+
+
+class SingleDesignatorTypeName(Enum):
+    """
+    Single designator types
+    """
+
+    int = auto()
+    uint = auto()
+    float = auto()
+    angle = auto()
+
+
+@dataclass
+class BitType(ClassicalType):
+    """
+    Bit type
+
+    Example::
+
+        bit[8]
+        creg[8]
+    """
+
+    type: BitTypeName
+    designator: Optional[Expression]
+
+
+class BitTypeName(Enum):
+    """
+    Bit types
+    """
+
+    bit = auto()
+    creg = auto()
+
+
+@dataclass
+class DoubleDesignatorType(ClassicalType):
+    """
+    Type with two designators. Currently, fixed is the only type
+
+    Example::
+
+        fixed[8, 12]
+    """
+
+    type: DoubleDesignatorTypeName
+    designator1: Expression
+    designator2: Expression
+
+
+class DoubleDesignatorTypeName(Enum):
+    """
+    Double designator types
+    """
+
+    fixed = auto()
+
+
+@dataclass
+class NoDesignatorType(ClassicalType):
+    """
+    Type with no designators.
+
+    Example::
+
+        bool
+        duration
+        stretch
+    """
+
+    type: NoDesignatorTypeName
+
+
+class NoDesignatorTypeName(Enum):
+    """
+    No designator type names
+    """
+
+    bool = auto()
+    duration = auto()
+    stretch = auto()
+
+
+@dataclass
+class ComplexType(ClassicalType):
+    """
+    Complex Type. Its real and imaginary parts are based on other classical types.
+
+    Example::
+
+        complex[int[32]]
+        complex[float[32]]
+    """
+
+    base_type: Union[SingleDesignatorType, DoubleDesignatorType]
+
+
+class IndexIdentifier(OpenNode):
+    """
+    Quantum or classical identifier,
+    indexed or not indexed.
+
+    #TODO: we will have to update IndexExpression,
+    then IndexIdentifier will be subclass of Expression.
+
+    Example::
+
+        q
+        b
+        b[1]
+        b[3:5]
+        b || c
+    """
+
+
+@dataclass
+class Subscript(IndexIdentifier):
+    """
+    Indexed identifier with an integer
+    as subscript.
+
+    Example:
+
+        segments[0]
+        qubits[1]
+    """
+
+    name: str
+    index: Expression
+
+
+@dataclass
+class Selection(IndexIdentifier):
+    """
+    Indexed identifier with multiple integers
+    as subscript.
+
+    Example::
+
+        segments[1, 2]
+        qubits[1, 2, 3]
+    """
+
+    name: str
+    indices: List[Expression]
+
+
+@dataclass
+class Slice(IndexIdentifier):
+    """
+    Indexed identifier with a range
+    as subscript.
+
+    Example::
+
+        segments[1:2]
+        qubits[1:10:1]
+    """
+
+    name: str
+    range: RangeDefinition
+
+
+@dataclass
+class RangeDefinition(OpenNode):
+    """
+    Range definition.
+
+    Example::
+
+        [1:2]
+        [1:10:1]
+        [:]
+    """
+
+    start: Optional[Expression]
+    end: Optional[Expression]
+    step: Optional[Expression]
+
+
+@dataclass
+class Concatenation(IndexIdentifier):
+    """
+    Combination of two arrays.
+
+    Example::
+
+        segment1 || segment2
+        qubits[0:10] || qubits[15:20]
+    """
+
+    lhs: Union[IndexIdentifier, Identifier]
+    rhs: Union[IndexIdentifier, Identifier]
+
+
+@dataclass
+class CalibrationGrammarDeclaration(Statement):
+    """
+    Calibration grammar declaration
+
+    Example::
+
+        defcalgrammar "openpulse";
+    """
+
+    calibration_grammar: str
+
+
+@dataclass
+class CalibrationDefinition(Statement):
+    """
+    Calibration definition
+
+    Example::
+
+        defcal rz(angle[20] theta) $q {
+            shift_phase drive($q), -theta;
+        }
+    """
+
+    name: str
+    arguments: List[ClassicalArgument]
+    qubits: List[str]
+    return_type: Optional[ClassicalType]
+    body: str
+
+
+@dataclass
+class SubroutineDefinition(Statement):
+    """
+    Subroutine definition
+
+    Example::
+
+    def measure(qubit q) -> bit {
+        s q;
+        h q;
+        return measure q;
+    }
+    """
+
+    name: str
+    arguments: List[Union[ClassicalArgument, QuantumArgument]]
+    return_type: Optional[ClassicalType]
+    body: List[Statement]
+
+
+@dataclass
+class QuantumArgument(OpenNode):
+    """
+    Quantum argument in subroutine definition
+
+    Example::
+
+        qubit q
+
+        qubit[4] q
+
+    """
+
+    qubit: Qubit
+    designator: Optional[Expression]
+
+
+@dataclass
+class ReturnStatement(Statement):
+    """
+    Classical or quantum return statement
+
+    Example::
+
+        return measure q;
+
+        return a + b
+
+    """
+
+    expression: Optional[Union[Expression, QuantumMeasurement]]
+
+
+class ControlDirectiveStatement(Statement):
+    """
+    Base class for control directive statements
+    """
+
+
+class BreakStatement(ControlDirectiveStatement):
+    """
+    Break statement
+
+    Example::
+
+        break;
+
+    """
+
+
+class ContinueStatement(ControlDirectiveStatement):
+    """
+    Continue statement
+
+    Example::
+
+        continue;
+
+    """
+
+
+class EndStatement(ControlDirectiveStatement):
+    """
+    End statement
+
+    Example::
+
+        end;
+
+    """
+
+
+@dataclass
+class BranchingStatement(Statement):
+    """
+    Branch (if) statement
+
+    Example::
+
+        if(temp == 1) { ry(-pi / 2) scratch[0]; } else continue;
+
+    """
+
+    condition: Expression
+    if_block: List[Statement]
+    else_block: Optional[List[Statement]]
+
+
+@dataclass
+class WhileLoop(Statement):
+    """
+    While loop
+
+    Example::
+
+    while(~success) {
+        reset magic;
+        ry(pi / 4) magic;
+        success = distill(magic, scratch);
+    }
+
+    """
+
+    while_condition: Expression
+    block: List[Statement]
+
+
+@dataclass
+class ForInLoop(Statement):
+    """
+    For in loop
+
+    Example::
+
+    for i in [0: 2] { majority a[i], b[i + 1], a[i + 1]; }
+
+    """
+
+    loop_variable: Identifier
+    set_declaration: Union[RangeDefinition, List[Expression], Identifier]
+    block: List[Statement]
+
+
+class TimingStatement(QuantumStatement):
+    """
+    Base class for timing statement
+    """
+
+
+@dataclass
+class DelayInstruction(TimingStatement):
+    """
+    Delay instruction
+
+    Example::
+
+        delay[start_stretch] $0;
+
+    """
+
+    arguments: List[Expression]
+    duration: Expression
+    qubits: List[Union[IndexIdentifier, Identifier]]
+
+
+@dataclass
+class Box(TimingStatement):
+    """
+    Timing box
+
+    Example::
+
+        box [maxdur] {
+            delay[start_stretch] $0;
+            x $0;
+        }
+
+    """
+
+    duration: Optional[Expression]
+    body: List[QuantumStatement]
+
+
+@dataclass
+class DurationOf(OpenNode):
+    """
+    Duration Of
+
+    Example::
+
+        durationof({x $0;})
+
+    """
+
+    target: Union[Identifier, List[QuantumStatement]]
+
+
+AssignmentOperator = Enum("AssignmentOperator", "= += -= *= /= &= |= ~= ^= <<= >>= %= **=")
+
+
+@dataclass
+class ClassicalAssignment(Statement):
+    """
+    Classical assignment
+
+    Example::
+
+        a[0] = 1;
+
+    """
+
+    lvalue: Union[Identifier, Subscript]
+    op: AssignmentOperator
+    rvalue: Expression
--- a/source/open_node/ast_visitor/ast_visitor.py
+++ b/source/open_node/ast_visitor/ast_visitor.py
@@ -1,0 +1,74 @@
+from typing import Optional, TypeVar, Generic
+
+from open_node.ast import OpenNode
+
+T = TypeVar("T")
+
+
+class NodeVisitor(Generic[T]):
+    """
+    A node visitor base class that walks the abstract syntax tree and calls a
+    visitor function for every node found.  This function may return a value
+    which is forwarded by the `visit` method.
+
+    Modified from the implementation in ast.py in the Python standard library
+
+    The visit/generic_visit methods in subclasses do not have to have context argument.
+    """
+
+    def visit(self, node: OpenNode, context: Optional[T] = None):
+        """Visit a node."""
+        method = "visit_" + node.__class__.__name__
+        visitor = getattr(self, method, self.generic_visit)
+        # The visitor method may not have the context argument.
+        if context:
+            return visitor(node, context)
+        else:
+            return visitor(node)
+
+    def generic_visit(self, node: OpenNode, context: Optional[T] = None):
+        """Called if no explicit visitor function exists for a node."""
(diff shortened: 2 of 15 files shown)


jwoehr:
We have a [Fall mentorship project to build a reference implementation](https://github.com/qiskit-community/mentorship-program-fall-21/issues/23) and we will be looking at "stealing" @aspcompiler 's open source code for the qasm -> ast stage 👍🏾

aspcompiler:
@jakelishman, our parser, i.e., taking OpenQASM as input as generating AST tree as output, uses the parser generated by Antlr.  Therefore, you can see it references [openqasm_reference_parser](https://github.com/Qiskit/openqasm/pull/269/files#diff-4597fa30a23d5f2d5222072525e6d3d4e92c6194a1a1da1446248e67e0dccd8aR7). Do you suggest that we hide the Antlr generated code inside our package?

jakelishman:
@aspcompiler: If it were up to me, yeah, that's what I'd do - `openqasm_reference_parser` doesn't really supply any functionality other than a very basic tree-printer of the raw parse tree.  When we've got a proper implementation with a reference AST-generation step, it might be simpler for other users to understand the package if we _only_ have that - I don't see any benefit to the current `openqasm_reference_parser`, since we could just have the instructions and build processes generate the grammar into your tool (or whichever ANTLR-based reference) instead as part of its build system.  I can't speak for the others, but I'm certainly not opposed to completely deleting `openqasm_reference_parser` in favour of an _actual_ reference parser.

If we really wanted to start packaging it up correctly, we might be better with a proper CMake step to detect the ANTLR command and generate the Python files, so it can all be handled automatically.  I didn't do that with the super simple package that's in the repo right now because it seemed like overkill at the time, and I never had intentions that that parser would have any serious development.

ajavadia:
Just out of curiosity, can you paste some performance metrics for this? i.e. on circuits of progressively larger sizes, up to say 1000s of qubits and gates.

aspcompiler:
@ajavadia, we will run some benchmark. Could you point us to any previous benchmark work so we can take a look?

The time to parse code to AST is the sum of 1) the time taken by Antlr generated parser to parse to parse-tree and 2) the time taken to convert parse-tree to AST. The time taken by 2) is actually O(n) time so it is limited by the performance of Python language. The time from 1) depends on whether we optimize the Antlr grammar. Antlr is an LL(*) that allows arbitrary look-aheads. The grammar performs better when we minimize number of look-aheads. Therefore, the metrics depends very much on which part of the grammar is used. It is important to get good code examples to have good real-world estimate.

jakelishman:
Would it be possible/useful to add some more tests of _invalid_ QASM 3 statements, to ensure that the grammar and parser correctly reject them? I wrote several fairly simple examples in [this directory](https://github.com/Qiskit/openqasm/tree/20be033/source/grammar/tests/invalid/statements) that are run by these tests in `openqasm_reference_parser`: https://github.com/Qiskit/openqasm/blob/20be033b3fe28b87993cc53c05e60e48c11e5216/source/grammar/tests/test_grammar.py#L123-L148

They felt helpful to me for codifying some places where things _shouldn't_ be allowed, but not sure how others feel.

aspcompiler:
@jakelishman, invalid statement tests are definitely useful. I will take a look of your tests.

aspcompiler:
@ajavadia , I did a test using [pnnl/QASMBench Ising mode](https://github.com/pnnl/QASMBench/blob/master/large/ising_model_n1000/ising_model_n1000.qasm)l (1000 qubits 10994 gates). The average Antlr parsing time is 31.4 seconds. The average translation time from parse tree to AST is 0.54 seconds. This is from my 2.6 GHz Intel Core i7 laptop. There is definitely a need to optimize the Antlr grammar.

source/open_node/ast_visitor/ast_visitor.py
```diff
@@ -0,0 +1,74 @@
...
+                    if isinstance(item, OpenNode):
+                        if context:
+                            self.visit(item, context)
+                        else:
+                            self.visit(item)
+            elif isinstance(value, OpenNode):
+                if context:
+                    self.visit(value, context)
+                else:
+                    self.visit(value)
```
stevenheidel, line 43: Shorter to turn a single entry into a list before iterating:

```suggestion
            if not isinstance(value, list):
                value = [value]
            for item in value:
                if context:
                    self.visit(value, context)
                else:
                    self.visit(value)
```

source/open_node/ast_visitor/ast_visitor.py
```diff
@@ -0,0 +1,74 @@
...
+
+class NodeTransformer(NodeVisitor[T]):
+    """
+    A :class:`NodeVisitor` subclass that walks the abstract syntax tree and
+    allows modification of nodes.
+
+    Modified from the implementation in ast.py in the Python standard library
+    """
+
+    def generic_visit(self, node: OpenNode, context: Optional[T] = None):
```
stevenheidel, line 54: Add return type `-> OpenNode`

source/open_node/ast_visitor/ast_visitor_examples.py
```diff
@@ -0,0 +1,71 @@
...
+    QubitDeclaration,
+    QuantumGate,
+    QuantumGateDefinition,
+)
+from open_node.ast_visitor.ast_visitor import NodeVisitor, NodeTransformer
+
+
+class OpenNodePrinter(NodeVisitor):
+    """
+    Example visitor which "unparses" the Program back to text
```
stevenheidel, line 13: Should probably point out that this is a partial implementation

source/open_node/ast.py
```diff
@@ -0,0 +1,1017 @@
...
+
+@dataclass
+class Program(OpenNode):
+    """
+    An entire OpenQASM 3 program represented by a list of top level statements
+    """
+
+    statements: List[Statement]
+    version: dict[str, int] = field(init=False, default_factory=dict, compare=False)
+    io_variables: List[IODeclaration] = field(init=False, default_factory=list)
```
stevenheidel, line 41: Are the IODeclarations also part of the statements list above?

source/open_node/ast.py
```diff
@@ -0,0 +1,1017 @@
...
+        q1 || q2
+
+    """
+
+    op: BinaryOperator
+    lhs: Expression
+    rhs: Expression
+
+
+ConstantName = Enum("ConstantName", "pi π tau 𝜏 euler ℇ")
```
stevenheidel, line 214: From the AST's perspective I think `pi` and `π` are synonyms; different tokens that represent the same thing. Therefore this enum should just be `Enum("ConstantName", "pi tau euler")`

source/open_node/ast.py
```diff
@@ -0,0 +1,1017 @@
...
+        !bool
+        -i
+
+    """
+
+    op: UnaryOperator
+    expression: Expression
+
+
+BinaryOperator = Enum("BinaryOperator", "> < >= <= == != && || | ^ & << >> + - * / % **")
```
stevenheidel, line 195: How difficult is it to use these binary operator enums? Another option would be to map to a name for each like `eq` and `and`

source/open_node/ast_visitor/ast_visitor_examples.py
stevenheidel: @shiyunon iirc you had made more progress on a printer?

source/open_node/ast_visitor/ast_visitor.py
```diff
@@ -0,0 +1,74 @@
+from typing import Optional, TypeVar, Generic
+
+from open_node.ast import OpenNode
```
stevenheidel, line 3: We should consider changing the package name to `openqasm` and then securing that name on PyPi.

source/open_node/ast_visitor/ast_visitor_examples.py
shiyunon: Yes, I have most of it done. @aspcompiler and I discussed and decided to not include it in this initial PR for (1) easier review (2) making the decision of putting it in `ast.py` vs another file public and more open.

source/open_node/ast.py
shiyunon: In the current ANTLR grammar, IODeclarations are part of the header, not a statement.

source/open_node/ast.py
shiyunon: +1

source/open_node/ast.py
shiyunon: Given the number of operators we have, it might have a better user experience if we align the symbols in the AST with the symbols in the grammar (so people don't have to constantly look it up).

source/open_node/ast_visitor/ast_visitor.py
shiyunon: +1

source/open_node/ast_visitor/ast_visitor_examples.py
```diff
@@ -0,0 +1,71 @@
...
+            self._print(f"{self.visit(modifier)} @ ")
+        self._print(f"{node.name} {', '.join(q.name for q in node.qubits)};")
+
+    def visit_QuantumGateModifier(self, node: QuantumGateModifier):
+        self._print(node.modifier)
+        if node.argument:
+            self._print(f"({self.visit(node.argument)})")
+
+
+class GateResolver(NodeTransformer):
```
shiyunon, line 44: This example can be removed as well since it has not Contextual Analysis, meaning that it cannot correctly replace the logical qubit name when resolving a gate.

source/open_node/ast.py
```diff
@@ -0,0 +1,1017 @@
...
+    measure_instruction: QuantumMeasurement
+
+
+@dataclass
+class ClassicalArgument(OpenNode):
+    """
+    Classical argument for a gate or subroutine declaration
+    """
+
+    type_: ClassicalType
```
shiyunon, line 502: We can change `type_` to `type` to align with other `type` fields.

source/open_node/ast_visitor/ast_visitor.py
aspcompiler: Unfortunately, openqasm was [claimed](https://pypi.org/project/openqasm/) 2 years ago. openqasm_parser and openqasm_ast are still available. Any preference?

jwoehr: openqasm_ast +1

jwoehr: (My little mentorship team wants to leverage your work in our QIskit Advocates Mentorship Program project and write a code generator from your AST)

source/open_node/ast_visitor/ast_visitor_examples.py
aspcompiler: OK. Since ast_visitor_examples.py have 2 incomplete examples that you recommend to remove, I will remove the file for now and add back in future PRs when we complete.

source/open_node/ast.py
aspcompiler: If we remove π, 𝜏, ℇ from the enum, we will no know from the AST whether we have pi or π in the code. When we print the AST, both pi and π will be printed as pi.

If we want to preserve printing, we can add another str field to keep the information.

What do you think?

source/open_node/ast.py
aspcompiler: Currently, the symbols in the grammar are unnamed, l[ike this one](https://github.com/Qiskit/openqasm/blob/master/source/grammar/qasm3.g4#L445). If we split the combined grammar into lexer and parser grammar, we will have to named them in the lexer grammar.

source/open_node/ast_visitor/ast_visitor_examples.py
aspcompiler: I removed ast_visitor_examples.py as you suggested in this and one of the previous comments.

source/open_node/ast_visitor/ast_visitor.py
aspcompiler: Changed the package name to openqasm for now.

source/open_node/ast.py
stevenheidel: I don't think we should preserve this - similar to how we don't preserve comments, spacing, and other things that don't affect the semantics of the program.

jwoehr: One thought which may or may not be useful:
My (two) hacked-together ASTs for OpenQASM 2 [[1](https://github.com/jwoehr/nuqasm2),[2](https://github.com/Qiskit/qiskit-terra/pull/6657)] preserve source lines for "reasons". No harm in that, the AST can be dumped/saved  for debugging, etc., allowing the programmer recognizes their own code when it appears in the debugger or reconstruction/retreival, or discarded if not needed.

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,926 @@
...
+        else:
+            node.span = span
+            return node
+
+    return wrapped
+
+
+class OpenNodeVisitor(qasm3Visitor):
+    @span
+    def visitProgram(self, ctx: qasm3Parser.ProgramContext):
```
hhorii, line 130: The current implementation does not handle `include*`. The first version will not support it, right?

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,926 @@
...
+                        init_expression=None,
+                    )
+                )
+
+        statements = []
+        for globalStatement in ctx.globalStatement():
+            statements.append(self.visit(globalStatement))
+        for statement in ctx.statement():
+            statements.append(self.visit(statement))
+        program = Program(statements=statements)
```
hhorii, line 173: can we sort `statements` with `span.start_line` and `span.column` before constructing `Program`?

source/openqasm/parser/antlr/qasm_parser.py
godott: We imagine there will be a pre-processing step before this and all `include` will be expanded.

source/openqasm/parser/antlr/qasm_parser.py
aspcompiler: Good catch! Added sorting.

source/openqasm/parser/antlr/qasm_parser.py
hhorii: Parsing is necessary to expand `include` and I think this AST is useful to do it.

shiyunon: > Parsing is necessary to expand `include`

Agree. For parsing `include` though, I feel it's more efficient to just search for include and replace it with the file without constructing an AST.

nelimee:
Hi @aspcompiler,

We started to use your AST generator for the [Qiskit Advocate Mentorship Program project on OpenQASM3](https://github.com/qiskit-advocate/qamp-fall-21/issues/23) and found some things that might be considered as issues.

Here is a quick description of each, sadly I did not have the time to investigate the source of these issues yet:


```python
# Code to get the AST from one of the example files in https://github.com/aspcompiler/openqasm/tree/ast/examples
from openqasm.parser.antlr.qasm_parser import parse

file = "adder.qasm"
with open(file, "r") as f:
    qasm_str = f.read()

ast = parse(qasm_str)
```

### 1. Qubits in gate definition are grouped

In the AST of `adder.qasm`, the following piece of code
```py
ast.statements[0].qubits
```
returns a list with only one entry:
```python
[Identifier(span=Span(start_line=8, start_column=14, end_line=8, end_column=20), name='a,b,c')]
```
I think it is supposed to be a list with 3 entries, am I wrong?

### 2. `for` sequence blocks are not parsed correctly

Still in `adder.qasm`, there is the following line:
```
for i in [2: -1: 0] { unmaj a[i],b[i+1],a[i+1]; }
```
The `[2: -1: 0]` is either wrong in the example, or wrong in the AST as the following happens:
```py
decl = ast.statements[17].set_declaration
assert decl.start.value == 2, "OK"
assert decl.step.value == 0, "A step of 0 is suspicious"
# decl.end.expression evaluates to -1
```

### 3. Some files are not entirely parsed

The file `t1.qasm` contains a big `for`-loop that does not appear in the AST.

### 4. Support for `negctrl`?

In the [OpenQASM3 paper](https://arxiv.org/abs/2104.14722) some examples include a `negctrl` but the identifier `negctrl` is not present in the enumeration `GateModifierName`.

### 5. `ConstantDeclaration.identifier` is a `str` instance instead of `Identifier`

From `ipe.qasm`, the line
```
const n = 10;          // number of iterations
```
is parsed as 
```
ConstantDeclaration(
    span=Span(
        start_line=7,
        start_column=0,
        end_line=7,
        end_column=12,
    ),
    identifier='n',
    init_expression=IntegerLiteral(
        span=Span(
            start_line=7,
            start_column=10,
            end_line=7,
            end_column=10,
        ),
        value=10,
    ),
)
```
but the attribute `identifier` should be an instance of `Identifier` instead of a raw string according to the AST typing.

shiyunon:
Hi, @nelimee . Thanks for your feedback!

1. This seems to be a problem in the argument list construction. We will update the argument list construction.
2. This is because `adder.qasm` has a small mistake, the AST constructor will check if the step is 0 to avoid infinite loops. We will have to replace the range definition `[2: -1: 0] ` with `[2: 0: -1] ` then it will be correctly parsed.
3. This is not because of `for` loops, but related to the current way of handling `defcal` which is incorrectly parsed now. There is a new proposal for `defcal`, we will test if the new proposal solves this problem: https://github.com/Qiskit/openqasm/issues/246.
4. We will add it.
5. We will fix this. 
Again, thanks for the useful feedback!

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,893 @@
...
+    def visitQuantumInstruction(self, ctx: qasm3Parser.QuantumInstructionContext):
+        quantum_gate_call = ctx.quantumGateCall()
+        quantum_phase = ctx.quantumPhase()
+        quantum_measurement = ctx.quantumMeasurement()
+        quantum_reset = ctx.quantumReset()
+        quantum_barrier = ctx.quantumBarrier()
+        if quantum_gate_call:
+            return self.visit(quantum_gate_call)
+        elif quantum_phase:
+            return self.visit(quantum_measurement)
```
nelimee, line 267: This seems like a typo. `quantum_measurement` should be `quantum_phase` here.
```suggestion
            return self.visit(quantum_phase)
```

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,893 @@
...
+
+        gate = QuantumGate(modifiers=modifiers, name=gate_name, arguments=arguments, qubits=qubits)
+
+        return gate
+
+    @span
+    def visitQuantumPhase(self, ctx: qasm3Parser.QuantumPhaseContext):
+        modifiers = [self.visit(modifier) for modifier in ctx.quantumGateModifier()]
+        argument = self.visit(ctx.expression())
+        qubits = [self.visit(qubit) for qubit in ctx.indexIdentifierList().indexIdentifier()]
```
nelimee, line 298: This line fails on the following OpenQASM line:
```
gate rz(λ) a { gphase(2); U(0, 0, λ) a; }
```
because `ctx.indexIdentifierList()` is `None` on the `gphase(2);` part.
This OpenQASM 3 line is valid according to [https://qiskit.github.io/openqasm/language/gates.html](https://qiskit.github.io/openqasm/language/gates.html?highlight=gphase) and the ANTLR reference grammar. You can also find a more complex version of it in `stdgates.inc`.

shiyunon: Updated.

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,894 @@
...
+
+        gate = QuantumGate(modifiers=modifiers, name=gate_name, arguments=arguments, qubits=qubits)
+
+        return gate
+
+    @span
+    def visitQuantumPhase(self, ctx: qasm3Parser.QuantumPhaseContext):
+        modifiers = [self.visit(modifier) for modifier in ctx.quantumGateModifier()]
+        argument = self.visit(ctx.expression())
+        if ctx.indexIdentifierList():
```
nelimee, line 298: Missing
```suggestion
        qubits = []
        if ctx.indexIdentifierList():
```

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,894 @@
...
+        return program
+
+    @span
+    def visitGlobalStatement(self, ctx: qasm3Parser.GlobalStatementContext):
+        return self.visit(ctx.getChild(0))
+
+    @span
+    def visitQuantumGateDefinition(self, ctx: qasm3Parser.QuantumGateDefinitionContext):
+        gate_name = ctx.quantumGateSignature().quantumGateName().getText()
+        arguments = []
```
nelimee, line 186: Missing a proper value for the `argument` variable.

Examples such as
```
OPENQASM 3;
gate rz(λ) a { gphase(-λ/2); U(0, 0, λ) a; }
```
cannot be correctly used because the lambda is missing from the argument list.

source/openqasm/ast.py
```diff
@@ -0,0 +1,995 @@
...
+    """
+    Constant declaration
+
+    Example::
+
+        const n = 10;
+    """
+
+    identifier: Identifier
+    init_expression: Optional[Expression]
```
hhorii, line 552: I think that `init_expression` of `ConstantDeclaration` is not `Optional`.

source/openqasm/parser/antlr/qasm_parser.py
shiyunon: Updated; Added this example as a test case.

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,908 @@
...
+    @span
+    def visitUnaryExpression(self, ctx: qasm3Parser.UnaryExpressionContext):
+        return UnaryExpression(
+            UnaryOperator[ctx.unaryOperator().getText()],
+            self.visit(ctx.powerExpression()),
+        )
+
+    @span
+    def visitBuiltInCall(self, ctx: qasm3Parser.BuiltInCallContext):
+        name = ctx.builtInMath().getText() if ctx.builtInMath() else ctx.castOperator().getText()
```
hhorii, line 559: I think `castOperator` should not become `FunctionCall`.

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,908 @@
...
+    @span
+    def visitControlDirective(self, ctx: qasm3Parser.ControlDirectiveContext):
+        if ctx.returnStatement():
+            return self.visit(ctx.returnStatement())
+        elif ctx.getChild(0).getText() == "break":
+            return BreakStatement()
+        elif ctx.getChild(0).getText() == "continue":
+            return ContinueStatement()
+        elif ctx.getChild(0).getText() == "end":
+            return EndStatement()
```
hhorii, line 811: `def visitEndStatement(...)` is necessary because the grammar says
```
statement
    : expressionStatement
    ...
    | endStatement
    ;
```

source/openqasm/parser/antlr/qasm_parser.py
```diff
@@ -0,0 +1,908 @@
...
+
+        for child in ctx.getChildren():
+            if isinstance(child, qasm3Parser.ExpressionContext):
+                expression = self.visit(child)
+                if colons_seen == 0:
+                    start = expression
+                elif colons_seen == 1:
+                    end = expression
+                else:
+                    step = expression
```
hhorii, line 857: I believe that this line should be
```
step = end
end = expression
```
The spec says
> A range is written as a:b or a:c:b where a, b, and c are integers (signed or unsigned).

https://qiskit.github.io/openqasm/language/types.html#register-concatenation-and-slicing

aspcompiler: Good catch! Fixed.

source/openqasm/parser/antlr/qasm_parser.py
aspcompiler: Indeed.

source/openqasm/parser/antlr/qasm_parser.py
aspcompiler: Agree. Fixed.

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+class CalibrationGrammarDeclaration(Statement):
+    """
+    Calibration grammar declaration
+
+    Example::
+
+        defcalgrammar "openpulse";
+    """
+
+    calibration_grammar: str
```
jakelishman, line 775: should this `str` be `StringLiteral`?

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+    Example::
+
+        defcal rz(angle[20] theta) $q {
+            shift_phase drive($q), -theta;
+        }
+    """
+
+    name: str
+    arguments: List[ClassicalArgument]
+    qubits: List[str]
```
jakelishman, line 792: identifier and List[identifier]? or List[Qubit]?

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+    Example::
+
+    def measure(qubit q) -> bit {
+        s q;
+        h q;
+        return measure q;
+    }
+    """
+
+    name: str
```
jakelishman, line 811: identifier?

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+
+    Example::
+
+        if(temp == 1) { ry(-pi / 2) scratch[0]; } else continue;
+
+    """
+
+    condition: Expression
+    if_block: List[Statement]
+    else_block: Optional[List[Statement]]
```
jakelishman, line 902: Do we need to include the `Optional` in the `else_block`?  Obviously they are actually optional in the grammar, but storing it as an empty list conveys the same information (I think?), and doesn't imply that's there any behavioural difference between `else_block = None` and `else_block = []`.

AbeerVaishnav13:
Hi @aspcompiler, we were in the process of using your QASM3 to AST generator for the [Qiskit Advocate Mentorship Program project on OpenQASM3](https://github.com/qiskit-advocate/qamp-fall-21/issues/23) and noticed a few bugs/potential changes...

### 1. No restriction on re-assignment to `const` types
Presently there are no restrictions on re-assignment to `const` types. For example, if I write the following code:
```
const my_const = 10;        // parses as a `ConstantDeclaration` statement
my_const = 5;               // parses as a `ClassicalAssignment` statement - WRONG
```

Shouldn't the AST gen step throw an error for statement-2 i.e. re-assignment to an immutable identifier?

### 2. `'bool'` instead of `NoDesignatorTypeName.bool`
The newly added `Cast` statement does not use the `NoDesignatorTypeName` enum. For example, if I take the code:
```
uint[4] a = 1;
bool b = bool(a[0])
```
gives the output AST as:
```
Cast(
    span=Span(
        start_line=35,
        start_column=5,
        end_line=35,
        end_column=17,
    ),
    type=NoDesignatorType(
        span=Span(
            start_line=35,
            start_column=5,
            end_line=35,
            end_column=5,
        ),
        type='bool',
    ),

// Omitted irrelevant part
```

Here the `NoDesignatorType.type` should contain the value `NoDesignatorTypeName.bool` instead of `'bool'`.

source/openqasm/ast_visitor/ast_visitor.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def visit(self, node: OpenNode, context: Optional[T] = None):
+        """Visit a node."""
+        method = "visit_" + node.__class__.__name__
+        visitor = getattr(self, method, self.generic_visit)
+        # The visitor method may not have the context argument.
+        if context:
+            return visitor(node, context)
+        else:
+            return visitor(node)
```
jakelishman, line 30: If all `visit_` methods take `context: Optional[T] = None`, then you can just `return visitor(node, context)` without branching - the `None` will safely get passed down in the same manner anyway.

(This comment applies to a few places in this file.)

source/openqasm/ast_visitor/ast_visitor.py
```diff
@@ -0,0 +1,73 @@
...
+    We added the context argument to the visit method. It allows the visitor
+    to hold temporary state while visiting the nodes.
+
+    The optional context argument in visit/generic_visit methods can be used to hold temporary
+    information that we do not want to hold in either the AST or the visitor themselves.
+    """
+
+    def visit(self, node: OpenNode, context: Optional[T] = None):
+        """Visit a node."""
+        method = "visit_" + node.__class__.__name__
```
jakelishman, line 24: I think the lookup here is inconsistent with your naming convention?  Your visitors are all named things like `visitQuantumGateDefinition` (no underscore).

Actually I may just be misinterpreting the purpose of this file - I'm not really sure what it's for, since it's not used by anything else.

jakelishman (commented):
I've left some comments in line, most of which are pretty minor, or just discussion points for the future, rather than anything to block merging.  In order to avoid bike-shedding about things that can always be changed later, I've not really commented on the reference parser.

One recurring point I had in the AST was that there seems to be an inconsistency between some things which are logically identifiers being declared `str`, others `Identifier`, and some `Qubit`.  Obviously the latter conveys some additional meaning, except I can't tell what the convention is for when you use each of the three.

A couple of things in particular I'd like to talk about more, though they don't need to be blockers on this PR:

1. **What downstream uses do we expect from this AST, and from the visitor/parser?**  We would like to use a reference tree implementation in downstream packages to handle the "tree -> QASM" step, as well as the "QASM -> tree step".  In that sense, I'm in favour of there being some tree that's able to completely represent every possible QASM programme, such that the conversion "QASM -> tree -> QASM" is entirely lossless (so I'd like to see `Comment` and `Include` nodes), but that's not a description of an AST. The reference parser/visitor certainly doesn't need to support that lossless transformation (it can continue to throw away comments and includes).  Might we want a reference concrete syntax tree as well?

2. **Packaging: how are we going to package and distribute the various components in this PR?**  I'd probably like to see the AST released as basically an entirely stand-alone entity, and the reference parser/visitor in its own Python package, so downstream uses can depend on a _super_ light Python AST module (potentially also with a concrete syntax tree?), without any extraneous code.  It would also be good to release the reference parser and visitor as a public PyPI package, so downstream packages can depend on it for AST generation if they like.

Apologies for the number of comments (most of them are just "identifier?"!), and that I didn't really have time to do this sooner, with you guys wanting to get it merged soon.  I'm not too beaten up if we don't resolve all of them before merge, especially since several of them are about AST design decisions which might take a while to discuss.

jakelishman:
@AbeerVaishnav13: some quick answers, since I'm looking at this (but am not on the main author team).

1. **reassignment to const**: the AST can't throw an error on that, because it's about syntax, not semantics.  The first pass of AST generation (which the reference implementation here is) doesn't generally track which identifiers are being assigned, and what their types are - that would most likely come in a later, compilation stage.

2. **cast operator doesn't preserve enum**: that looks like a bug to me.

aspcompiler:
@AbeerVaishnav13:  @jakelishman just answered 1. I just pushed a fix for 2.

source/openqasm/ast_visitor/ast_visitor.py
aspcompiler: This file is the base class for the visitor of AST. We had some examples such as a pretty printer, but unfortunately did not add them to this commit due to incompleteness.

`visitQuantumGateDefinition` is from overriding Antlr generated base class. Antlr used Java naming convention.

We used `visit_` for our visitor methods to follow Python convention.

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+    designator: Optional[Expression]
+
+
+class BitTypeName(Enum):
+    """
+    Bit types
+    """
+
+    bit = auto()
+    creg = auto()
```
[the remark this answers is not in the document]
stevenheidel: The answer is "yes", we should fold this up in the same way as your other comments.

stevenheidel:
> What downstream uses do we expect from this AST, and from the visitor/parser? We would like to use a reference tree implementation in downstream packages to handle the "tree -> QASM" step, as well as the "QASM -> tree" step, and I'd like to have a concrete reference tree for tooling purposes. In that sense, I'm in favour of there being some tree that's able to completely represent every possible QASM programme, such that the conversion "QASM -> tree -> QASM" is entirely lossless, but what I'm describing for tooling is not an AST. The reference parser/visitor certainly doesn't need to support that lossless transformation (it can continue to throw away comments and includes). Might we want a reference concrete syntax tree as well?

I would love to see someone contribute a concrete syntax tree (CST) for the purposes of developer tools. This would be helpful for editor plugins and refactoring plugins. CSTs are overkill for most compiler passes though as you point out since you have to keep track of line numbers, parenthesis, tab levels, etc etc

> Packaging: how are we going to package and distribute the various components in this PR? I'd probably like to see the AST released as basically an entirely stand-alone entity, and the reference parser/visitor in its own Python package, so downstream uses can depend on a super light Python AST module (potentially also with a concrete syntax tree?), without any extraneous code. It would also be good to release the reference parser and visitor as a public PyPI package, so downstream packages can depend on it for AST generation if they like.

I'd like to avoid the packaging burden of having too many packages. The visitor and transformer classes are very little code so I feel should be in the same package as the AST. The main concern would be the "ANTLR -> AST" code, since this requires an additional dependency of "antlr4_python_runtime". Maybe we can use Python extras to avoid installing that code if the user wants a pure Python AST.

For instance,
`pip install openqasm3` would install a dependency-free package of just the AST and Listener/Transformer classes.
`pip install openqasm3[antlr]` would also install the the antlr parser as well, with the runtime dependency

source/openqasm/parser/antlr/qasm_parser.py
aspcompiler: I just added `Include` as an AST node. Since the parser can parse a program in a string, we don't want the parser to resolve `include` from file system. The visitor of the AST can resolve the file and run another parser to parse the include file.

source/open_node/ast_visitor/ast_visitor.py
```diff
@@ -0,0 +1,74 @@
...
+T = TypeVar("T")
+
+
+class NodeVisitor(Generic[T]):
+    """
+    A node visitor base class that walks the abstract syntax tree and calls a
+    visitor function for every node found.  This function may return a value
+    which is forwarded by the `visit` method.
+
+    Modified from the implementation in ast.py in the Python standard library
```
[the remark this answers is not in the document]
aspcompiler: Added comments.

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+    span: Optional[Span] = field(init=False, default=None, compare=False)
+    """
+    The span(location) of the node in the source code.
+    Because not all the nodes are generated from source, the span is optional.
+    To make it easier to write unit test, we exclude span from the generated __eq__().
+    """
+
+
+@dataclass
+class Program(OpenNode):
```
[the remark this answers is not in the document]
aspcompiler: Added

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+        qubit[4] q;
+
+        q // <- qubit
+        [4] // <- designator
+
+    """
+
+    decl_type_name: QubitDeclTypeName
+    qubit: Qubit
+    designator: Optional[Expression]
```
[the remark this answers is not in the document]
aspcompiler: We removed `decl_type_name` and `QubitDeclTypeName` to simplify.

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+
+
+@dataclass
+class Program(OpenNode):
+    """
+    An entire OpenQASM 3 program represented by a list of top level statements
+    """
+
+    statements: List[Statement]
+    version: dict[str, int] = field(init=False, default_factory=dict, compare=False)
```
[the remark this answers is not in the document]
aspcompiler: Simplified version to str.

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+
+    Example::
+
+        gate cx c, t {
+            ctrl @ unitary(pi, 0, pi) c, t;
+        }
+
+    """
+
+    name: str
```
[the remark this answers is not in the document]
aspcompiler: Changed to Identifier for consistency.

source/openqasm/ast.py
aspcompiler: Yes, List[Qubit]

source/openqasm/ast.py
aspcompiler: Yes.

source/openqasm/ast.py
aspcompiler: Good idea! Language like Kotlin intentional makes it hard to use null/None so one does not have to do null check. Changed.

source/openqasm/ast_visitor/ast_visitor.py
aspcompiler: One reason that this branch existed is that it gives users not to have the `context` parameter in visit_XXX methods if they do not need it. Once we add some concrete classes of the ast_visitor, it will much more clear.

jakelishman:
There's still a couple of outstanding comments in the AST (`us`/`µs` in `TimeUnit`, a naming oddness around `input` and `output`, and the possible split of type handling), but only type handling is more than trivial.  If we're keen to get this in in the next couple of days, and if we're happy that we're not going to make _any_ stability guarantees about the AST for now, I'm not opposed to merging without further discussion on the type handling yet.  At any rate, that's something we may want to also tweak in the grammar as well.

Thanks for doing all this Li and Yunong, and for the speed of changes you've been making after comments.  Thanks as well for putting the `Include` node in this tree despite the original design decision against it - that's going to help us a lot on the QASM generation side.

@stevenheidel: thanks for those responses - it's good to hear, and I think I align pretty well with everything you said.  On the point of packaging: I understand that concern against too many packages.  The idea of having the `antlr4_runtime` dependency gated behind an optional extra is good, and that would satisfy us for our downstream dependencies.  Similarly, it would be good to keep this pure Python with long-term Python version support (e.g. at least 3.7--3.10 on release), so we don't stop downstream packages depending on this because it's not available in the same environments.  This will probably already be the case, though, and while `dataclass` is a Python 3.7 thing, if we desperately want 3.6, we could swap to the external library `attrs` with only minor changes.

aspcompiler:
@jakelishman  Github collapsed some comments so I missed them; I will take care of them today. Thanks for the reminder!

I was not aware of the original decision decision against the include node. Could you point me to the discussion? What are alternatives?

jakelishman:
Oh sorry, I didn't mean anything formal, I just meant about how you had (seemingly deliberately) left it out originally, and Yunong had commented that handling it would be easier by a preprocessor pass (https://github.com/Qiskit/openqasm/pull/269/files#r710654212)

shiyunon:
> Oh sorry, I didn't mean anything formal, I just meant about how you had (seemingly deliberately) left it out originally, and Yunong had commented that handling it would be easier by a preprocessor pass (https://github.com/Qiskit/openqasm/pull/269/files#r710654212)

Yeah, sorry, the `include` part in the AST was edited by me, so Li might not be aware of it. With Hiroshi and follow up comments, I fully agree that we should have `include` node in the AST.

aspcompiler:
I see. I think `include node` provides flexibility whether to use a preprocessor. For those who preprocessed, the `include statements` were removed so the include node has no effect. For those who do not want to write a preprocessor, they can use this parser to parse each include files. The `include node` also allows the scenario that include file is at a different location, for example, the main qasm file is processed in one system, and then pass to another system that has calibration as `defcal` in an include file.

source/openqasm/ast.py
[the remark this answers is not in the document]
aspcompiler: Folded

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+    """
+    Complex Type. Its real and imaginary parts are based on other classical types.
+
+    Example::
+
+        complex[int[32]]
+        complex[float[32]]
+    """
+
+    base_type: Union[SingleDesignatorType, DoubleDesignatorType]
```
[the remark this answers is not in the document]
aspcompiler: Done.

source/openqasm/ast.py
aspcompiler: `StringLiteral` node is just a wrapper over a `str` value, but it gives us a span. I am in favor of StringLiteral.

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+    """
+
+    type: ClassicalType
+    identifier: Identifier
+    init_expression: Optional[Expression]
+
+
+class IOIdentifierName(Enum):
+    output = auto()
+    input = auto()
```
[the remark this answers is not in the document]
shiyunon: Makes sense, it's definitelly closer to a keyword. In this case, I would think `IOKeyword` might be a good name.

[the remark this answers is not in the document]
jakelishman: `IOKeyword` sounds fine to me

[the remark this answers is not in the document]
shiyunon: updated.

source/openqasm/ast.py
```diff
@@ -0,0 +1,1010 @@
...
+    """
+
+    value: str
+
+
+class TimeUnit(Enum):
+    dt = auto()
+    ns = auto()
+    us = auto()
+    µs = auto()
```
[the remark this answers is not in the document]
shiyunon: Hi, Jake. I updated to remove µs.

jakelishman:
I made a PR (aspcompiler#1) to the base branch to change the field name "designator" to "size" for the types (since that's what it is), if people are happy for that.

Other than that incredibly minor nitpick, I'm happy with the AST as it is now.

aspcompiler:
> I made a PR ([aspcompiler#1](https://github.com/aspcompiler/openqasm/pull/1)) to the base branch to change the field name "designator" to "size" for the types (since that's what it is), if people are happy for that.
> 
> Other than that incredibly minor nitpick, I'm happy with the AST as it is now.

Merged your PR. Thanks, @jakelishman

jakelishman (approved):
I like the AST in its current form, and I think it gives us a really solid base to build downstream packages on.  I think if we can get 4/6 TSC members to sign off on with their ticks, we can merge this without waiting for the meeting in two weeks' time.

After this, I suppose we'll need to sort out packaging and deployment to PyPI, but that probably _does_ need a little more discussion, which can happen after merge.

jakelishman:
@blakejohnson: oh perfect, thanks! I should have thought someone might have done a backport.

shiyunon:
Per discussion with @aspcompiler @stevenheidel , we removed `Qubit` node in the AST to be consistent with variables of other types. For variables with `Bit`, `Int` types, they are all `Identifier`. Thus, we make qubit variables `Identifier` as well. We leave the type checking to semantic analysis performed in the AST.

PR #270, opened 2021-09-02
footnote for affiliation

### LINKED ISSUE #64: Fix the contribution in the live documentation
Screen Shot 2020-11-03 at 6 18 13 PM

can we make this ^1 or * so it's simpler to read.

1ucian0:
Fixes #64

Fixes footnote for affiliation.

### DIFF
--- a/source/intro.rst
+++ b/source/intro.rst
@@ -54,9 +54,19 @@ Contributors
 
 The following individuals contributed to the OpenQASM version 3.0 specification:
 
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
Clearly correct from a markup perspective. The current behaviour doesn't register the link targets correctly:
Screenshot 2021-09-02 at 14 06 14
But with this change, it now looks like
Screenshot 2021-09-02 at 14 06 35

Sphinx makes some slightly strange choices with how it renders footnotes with multiple references, but arguing about that would just be an opportunity for us to bikeshed.

Issue #124, opened 2021-01-13
builtin math functions


hhorii:
[OpenQASM](https://arxiv.org/pdf/1707.03429.pdf) paper defines builtin math functions: `sin`, `cos`, `tan`, `exp`, `ln`, and `sqrt`. 
But an example ([rus.qasm](https://github.com/Qiskit/openqasm/blob/8c5d723aee624a8f044dbceb469d3ae483d36029/examples/rus.qasm#L38)) and [README](https://github.com/Qiskit/openqasm/blob/8c5d723aee624a8f044dbceb469d3ae483d36029/examples/rus.qasm#L38) uses `arccos`. 

I guess that OpenQASM3.0 extends builtin math functions and provide trigonometric-functions (and others). https://numpy.org/doc/stable/reference/routines.math.html#trigonometric-functions

blakejohnson:
I think I would rather go the other way and remove most of these from the language. We can always pull them back in through kernel functions.

Other opinions?

taalexander:
I think this makes sense for simplicity right now. We could define a `stdclmath.qasm` file with these kernels later on, or alternatively readd them to the language at that point?

zachschoenfeld33:
I think we should add all three inverse trigonometric functions. This would also address https://github.com/Qiskit/openqasm/issues/25.

jwoehr:
Since we're talking Specification here instead of Implementation, and since `arccos` and its pals are algebraically derivable from the base trig functions using operators already admitted in OpenQASM 3 expression syntax, we could punt as follows:
- `sin`, `cos`, `tan`, `exp`, `ln`, and `sqrt` are in.
- `arccos` and others _may_ be present, and the vendor will so document.
- If "they" (some selection of higher trig functions enumerated by the Spec) are absent, the vendor will document an equivalent supported expression.

stevenheidel:
We discussed this at the TSC meeting on September 3rd. Summarizing the discussion below:

There are two places where these functions can appear: in compile-time positions and run-time positions. Compile-time positions include gate definitions and consts. Run-time positions include classical statements that are based on the output of measurement operations.

It is important to have many of these functions built-in to the language, especially given their utility in defining gates. So for compile-time positions, we should include the inverse trig functions like `arccos`.

[...]

PR #275, opened 2021-09-18
Support For *Composition* Inverses 

NoahGWood:
### Summary
Added grammar support for built-in math functions:
-     asin
-     acos
-     atan

### Details and comments
Per issue: [124](https://github.com/Qiskit/openqasm/issues/124)
Files modified:   source/grammar/qasm3.g4
Lines Modified:   411-423

### DIFF
--- a/source/grammar/qasm3.g4
+++ b/source/grammar/qasm3.g4
@@ -409,7 +409,18 @@ builtInCall
     ;
 
 builtInMath
-    : 'sin' | 'cos' | 'tan' | 'exp' | 'ln' | 'sqrt' | 'rotl' | 'rotr' | 'popcount'
+    : 'sin'
+    | 'asin'
+    | 'cos'
+    | 'acos'
+    | 'tan'
+    | 'atan'
+    | 'exp'
+    | 'ln'
+    | 'sqrt'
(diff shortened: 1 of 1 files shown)


jakelishman:
Hello, and thanks for this!

The examples given in #124 use the inverse trig functions with the names `arccos`, etc, as opposed to `acos`, as in this PR.  We should decide which actually want to have.  I personally prefer `arccos`, but C, Rust and Python all use `acos`.

We'll also need to add some tests of these to the CI runs, probably in [the "builtin calls" file](https://github.com/Qiskit/openqasm/blob/f8b1a52/source/grammar/tests/reference/expression/built_in_call.yaml).

stevenheidel:
>  I personally prefer arccos

Same, OpenQASM tends to be biased towards spelling things out explicitly.

1ucian0 (changes requested):
once the changes requested in https://github.com/Qiskit/openqasm/pull/275#issuecomment-922748892 are addressed, this PR is ready to be merged.

NoahGWood:
Sorry, been a little tied up lately.
I've updated it to use arc instead, I've also added in tests for all the built-in math functions

Issue #210, opened 2021-05-24
Paper: Classical types w/ out width specification


zachschoenfeld33:
### Summary

As described in https://arxiv.org/pdf/2104.14722.pdf, the spec should be updated to allow for the classical types `int`, `uint`, `bit`, `float` to occur without requiring a bit width. That is, to declare an integer, one can do either
```
int x = 5;
int[3] x = 5;
```

In the first example, the bit-precision of `x` is defined by the particular hardware target. In the second example, `x` contains 3 bits. Only on types with the bit width explicitly defined is bit level access allowed. For instance, on the second example we could do  `bit a = x[1];` or `x[2] = 0;` This is not allowed in the first example.

We should also make this "bit-level" access explicit in the spec.

This flexibility allows high level programmers to abstract away hardware details while allowing low level applications to access constraints particular to a given controller. It is expected that any given hardware unit will only support certain hardware widths. 

### Questions
- Is `fixed` removed from the language?

zachschoenfeld33:
Add `complex` to this list when #175 is merged.

PR #276, opened 2021-09-19
allow no designator for singleDesignatorType

hhorii:
### Summary

`int`, `uint`, and `float` allow no designator.

### Details and comments

The paper allows `int`, `uint`, and `float` variable declaration without width:
```
int x = 5;
```

This PR will resolve a part of issues written in https://github.com/Qiskit/openqasm/issues/210.

### DIFF
--- a/source/grammar/qasm3Parser.g4
+++ b/source/grammar/qasm3Parser.g4
@@ -48,14 +48,14 @@ quantumDeclaration: QREG Identifier designator? | QUBIT designator? Identifier;
 quantumArgument: QREG Identifier designator? | QUBIT designator? Identifier;
 
 /** Classical Types **/
-bitType: BIT | CREG;
-singleDesignatorType: INT | UINT | FLOAT | ANGLE;
+bitType: (BIT | CREG) designator?;
+singleDesignatorType: (INT | UINT | FLOAT | ANGLE) designator?;
 noDesignatorType: BOOL | DURATION | STRETCH;
 
 nonArrayType
-    : singleDesignatorType designator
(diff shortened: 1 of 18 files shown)


jakelishman:
This is currently under discussion in the types WG as to what the semantics are of this, and what it really means for parsers and consumers of OpenQASM 3.  It might be better to leave this on hold for the time being.

mbhealy:
I propose that we update this PR to include language describing the meaning of the undesignated types in the spec. In the TAC WG we have settled upon an interpretation as follows:
Undesignated types assume a precision that is a default controlled by the compiler, likely contingent upon the target architecture being compiled for. Undesignated types are different from all designated types for the purposes of casting.

jakelishman:
I've updated this PR to solve the merge conflicts that had happened since its last commit, add the additional necessary text to the specification, and add the necessary structure to the reference AST/parser.  This is ready for final review now.

(It'll conflict with #278, but I'll solve those issues for whichever PR merges second.)

blakejohnson:
I committed to reviewing this at the TSC meeting today.

jakelishman:
@blakejohnson: the only parts that need TSC review are [the changes to the `rst` files](https://github.com/Qiskit/openqasm/pull/276/files?file-filters%5B%5D=.rst&show-viewed-files=true).  The rest is just updating surrounding code, and the consensus from the TSC meeting today seemed to be that we're happy to merge that without stringent review.

stevenheidel (approved):
I reviewed the spec changes and approve (trusting you on the rest)

PR #277, opened 2021-09-21
`rangeDefinition` has at least two expressions

hhorii:
### Summary

`rangeDefinition` must have at least two expressions

### Details and comments

`rangeDefinition` is defined in the [spec](https://qiskit.github.io/openqasm/language/types.html#register-concatenation-and-slicing):

> A range is written as `a:b` or `a:c:b` where `a`, `b`, and `c` are integers (signed or unsigned). The range corresponds to the set `{a, a+c, a+2c, ..., a+mc  }` where is the largest integer such that `a + mc <= b` if `c>0` and `a + mc >=b` if `c < 0`. If `a == b` then the range corresponds to `{a}`. Otherwise, the range is the empty set. If `c` is not given, it is assumed to be one, and `c` cannot be zero. Note the index sets can be defined by variables whose values may only be known at run time.

According to the above definition, the `rangeDefinition` has at least two expressions `a` and `b`. The current Antlr grammar allows no expression, such as `[:]`.

### DIFF
--- a/source/grammar/qasm3.g4
+++ b/source/grammar/qasm3.g4
@@ -192,7 +192,7 @@ indexIdentifierList
     ;
 
 rangeDefinition
-    : LBRACKET expression? COLON expression? ( COLON expression )? RBRACKET
+    : LBRACKET expression COLON expression ( COLON expression )? RBRACKET
     ;
 
 /*** Gates and Built-in Quantum Instructions ***/


hhorii:
I understand that this grammar will have a kind of relaxation by expecting another phase checks grammar strictly.
Moreover, in this case (unlike other cases), the live spec describes the length of arguments specifically.
I believe that we can close this issue.

PR #278, opened 2021-09-21
Split lexer and parser ANTLR grammars

### LINKED ISSUE #263: Problem with pong.qasm in examples
### Informations

- **OpenQASM version**: 3
- **Python version**: 3.x
- **Operating system**: Any

Grammar error with lines 24-26, 28 and 45.

If we comment out defcals in pong.qasm and run the tests, the tests will fail.

```
E           openqasm_reference_parser.exceptions.Qasm3ParserError: Parse tree build failed. Error:
E           line 24:8 mismatched input 'delay' expecting {'bit', 'creg', 'int', 'uint', 'float', 'angle', 'fixed', 'bool', 'complex', '~', '!', 'true', 'false', 'sin', 'cos', 'tan', 'exp', 'ln', 'sqrt', 'rotl', 'rotr', 'popcount', 'duration', 'stretch', 'durationof', '(', '-', ImagNumber, Constant, Integer, Identifier, RealNumber, TimingLiteral, StringLiteral}
E           line 25:8 mismatched input 'delay' expecting {'bit', 'creg', 'int', 'uint', 'float', 'angle', 'fixed', 'bool', 'complex', '~', '!', 'true', 'false', 'sin', 'cos', 'tan', 'exp', 'ln', 'sqrt', 'rotl', 'rotr', 'popcount', 'duration', 'stretch', 'durationof', '(', '-', ImagNumber, Constant, Integer, Identifier, RealNumber, TimingLiteral, StringLiteral}
E           line 26:8 mismatched input 'delay' expecting {'bit', 'creg', 'int', 'uint', 'float', 'angle', 'fixed', 'bool', 'complex', '~', '!', 'true', 'false', 'sin', 'cos', 'tan', 'exp', 'ln', 'sqrt', 'rotl', 'rotr', 'popcount', 'duration', 'stretch', 'durationof', '(', '-', ImagNumber, Constant, Integer, Identifier, RealNumber, TimingLiteral, StringLiteral}
E           line 28:19 missing ';' at '-'
E           line 28:34 missing ';' at '$0'
E           line 28:44 mismatched input ';' expecting '='
E           line 45:4 mismatched input 'if' expecting {'U', 'CX', 'gphase', 'reset', 'measure', 'barrier', 'inv', 'pow', 'ctrl', 'negctrl', 'for', 'while', 'box', 'delay', 'rotary', '}', Identifier}

```

The test should pass even if the defcals are not present. The issue was hidden by the `defcal` grammar:

```
calibrationDefinition
    : 'defcal' Identifier
    ( LPAREN calibrationArgumentList? RPAREN )? identifierList
    returnSignature? LBRACE .*? RBRACE  // for now, match anything inside body
    ;
```
Note the wildcard non-greedy match `LBRACE .*? RBRACE`. Since Antlr will try to match the entire file, if there is any invalid syntax after `defcal`, Antlr will match everything after `defcal` to the `defcal` body and thus hides the error.

### Suggested solutions

The current grammar cannot parse the following:

```
glue[1]-delay[0] $0;
```

Need to fix either grammar or the example.

### LINKED ISSUE #284: Redundant rules with ANTLR grammar

In the ANTLR grammar, 
- on line 86, quantumArgumentList is defined but never used.
- on line 296, logicalOperator is defined but never used.
Expected enhancement: remove above two lines.

jakelishman:
### Summary

This makes a clean split of the lexer and parser rules for the ANTLR
grammar, with no changes to the actual grammar of the language other
than those that were required to make the split.  The small extent of
these changes can be seen in the small diffs of the reference outputs.
The purpose of this split was to enable us to use lexer-specific modes
in the reference implementation, such as island grammars for parsing
pragmas and `defcal` blocks.

The main changes are just that keywords, symbols and builtin names are
now lexed as explicit tokens, rather than having implicit rules within
the parser.  In a couple of cases, moving the rules from unnecessary
parser rules (e.g.  booleanLiteral) into lexer rules caused a small part
of the parse tree to collapse, for example

    booleanLiteral
        true

became

    true

since the parser's `booleanLiteral` became a lexeme `BooleanLiteral`.

This should not have made any previously valid QASM 3 programmes
invalid, nor the other way around.

### Details and comments

This is the change that will enable #246, and let us provide a fix for #263 (and probably some other problems).  I'm making this split PR first, because there should be less that needs to be discussed in it, than in a full implementation of island grammars for pragmas and defcal blocks - I don't want to block #269 while we're discussing that.

In general, I kept as much of the original files as I could, so they should be very similar (albeit with whitespace changes).  Git won't recognise them as "sort-of renamed" though.

This will cause conflicts with several current PRs due to the split in the grammar files (i.e. #277, #276, #275, #269 and #201), but with the exception of #269, they should all be easy to bring up-to-date.

With commit 5c1f6057, this fixes #284.


jakelishman:
Now rebased over #201.

source/grammar/qasm3.g4
```diff
@@ -1,610 +0,0 @@
...
-    | pragma
-    ;
-
-statement
-    : expressionStatement
-    | assignmentStatement
-    | classicalDeclarationStatement
-    | branchingStatement
-    | loopStatement
-    | endStatement
```
shiyunon, line 46: Just want to have a clarification, what is the purpose of endStatement in the grammar? Is it similar to Fortran's `end`?

jakelishman: Yeah, it does the same thing: ends the programme.  It's buried in a slightly arcane position in the spec, just before the last example of the looping and branching section: https://qiskit.github.io/openqasm/language/classical.html#looping-and-branching.

source/grammar/qasm3Parser.g4
```diff
@@ -0,0 +1,361 @@
...
+    | pragma
+    ;
+
+statement
+    : expressionStatement
+    | assignmentStatement
+    | classicalDeclarationStatement
+    | branchingStatement
+    | loopStatement
+    | endStatement
```
shiyunon, line 30: Just want to have a clarification, what is the purpose of endStatement in the grammar? Is it similar to Fortran's `end`?

jakelishman: Yeah, it does the same thing: ends the programme. It's buried in a slightly arcane position in the spec, just before the last example of the looping and branching section: https://qiskit.github.io/openqasm/language/classical.html#looping-and-branching.

source/grammar/qasm3Parser.g4
```diff
@@ -0,0 +1,361 @@
...
+
+quantumReset
+    : RESET indexIdentifierList
+    ;
+
+quantumMeasurement
+    : MEASURE indexIdentifier
+    ;
+
+quantumMeasurementAssignment
```
shiyunon, line 169: On line 149, in `quantumInstruction` rule, we have `quantumMeasurement` but it cannot be a `quantumMeasurementAssignment`. Is there a reason to not include assignment there? (I am also wondering what's the expected compiler behavior if the measurement results are not stored)

If indeed we don't want to include "assignment" in `quantumInstruction` for some reason, I suggest we can make the `(ARROW indexIdentifier)?` part on line 170 not optional. It will not change the grammar. Because in `statement`, we have both `quantumStatement` (thus `quantumInstruction`) and `assignmentStatement` (thus `quantumMeasurementAssignement`. This will bring more clarity to the parser.

jakelishman: I didn't change any of these parser rules in the split (as far as I remember!).  In general, I'd like to keep it that way, so this PR is just a (mostly) clean application of the split, then we can do a bunch of follow-ups to fix all these little grammatical niggles.

On the actual grammar issue, though, yeah, I agree.

source/grammar/qasm3Parser.g4
```diff
@@ -0,0 +1,361 @@
...
+    | RealNumber
+    | ImagNumber
+    | BooleanLiteral
+    | Identifier
+    | StringLiteral
+    | builtInCall
+    | externOrSubroutineCall
+    | timingIdentifier
+    | LPAREN expression RPAREN
+    | expressionTerminator LBRACKET expression RBRACKET
```
shiyunon, line 296: I suggest we should refine this rule a bit. Currently, there's an overlap between this line and `IndexIdentifier`, thus, sometimes the same string, say `a[10]`, can be parsed to either this or `IndexIdentifier`, depending on the location in the program.

Also, this line excludes expressions like `q[1:3]`, `q[1, 2, 3]` (since rangeDefinition is not an expresssion) but `IndexIdentifier` includes them. I am not sure if it's intended. It seems to me that those are useful to be in `expression`.

Looks like the only `expressionTerminator` can show up in this line are `Identifier` and `externOrSubroutineCall`. If we can make these 2 rules explict and potentially include `IndexIdentifier` into `expression` (depending on if we want things `q[1:3]` to be `expression`), it will make the grammar more clear.

source/grammar/qasm3Parser.g4
```diff
@@ -0,0 +1,361 @@
...
+    | WHILE LPAREN expression RPAREN
+    ;
+
+loopStatement: loopSignature programBlock;
+endStatement: END SEMICOLON;
+returnStatement: RETURN (expression | quantumMeasurement)? SEMICOLON;
+
+controlDirective
+    : (BREAK| CONTINUE) SEMICOLON
+    | endStatement
```
shiyunon, line 333: We have `endStatment` both here and in `statement` on line 30. It might confuse the parser, I am wondering if we should just put it in one place.

jakelishman:
I'm pretty happy with all your grammar suggestions, @shiyunon, just (if it's ok with you), I'd rather do them in follow-up PRs, so it'll be easier to trace discussion of them in the future.  Perhaps it's just my way of working - when I try and discover why a change was made, I tend to find it in the `git log`, then read the discussion in the PR that included that commit, and it's easier to trace that all if the PRs have a bit more separation of concerns.

jakelishman:
I've (finally) updated this PR to be in-line with the current specification.  I'd like to merge this relatively soon to avoid it getting out-of-sync again, and because we need it to solve some existing issues with parsing of `defcal` blocks, and the pragmas group will need this to more easily write their proposed additions in the reference grammar (I'm happy to help when it comes time to write that).

On merge, I'll update any existing PRs that have merge conflicts caused because of this.

source/grammar/qasm3Parser.g4
jakelishman: This comment chain was obsoleted by changes to how indexing is handled during the arrays PR (#283) - `indexExpression` became a complete part of the expression tree (with very tight priority) because we needed to allow the grammatical possibility of multiple indexers on the same identifier, and a specific rule was added for indexed identifiers in l-value positions.

source/grammar/qasm3Parser.g4
jakelishman: Yeah, I guess it was probably an oversight in the original grammar that caused this.  It won't confuse ANTLR - it has a defined priority for which rule it tries to construct, so it'll reliably always pick the same production (though I don't remember exactly which it'll be).

Similar to other bits that actually change the grammar, we can fix it up in a follow-on.

jwoehr:
Since this effort is blessed by the TSC I'll approve and merge if you say it's good, @jakelishman 
Just post back and confirm, please.

jakelishman:
From my perspective I'm fine for it to be merged, though the most recent changes haven't officially been reviewed yet.

jwoehr:
Shall we wait then for @hhorii and/or @shiyunon to make final comments?

jakelishman:
Possibly @shiyunon or @aspcompiler, since I made changes to the reference AST and Python parser that they were the original authors of.  The grammar changes should (with the exception of one bit to do with `aliasInitializer` that was being mildly awkward) be an entirely mechanical split of the tokenisation from the parsing.

stevenheidel:
Checked with @shiyunon offline and no concerns

stevenheidel (approved):

Issue #218, opened 2021-06-02
clarification on singleton registers and single bits equivalence


1ucian0:
When declare (qu)bits, is `bit foo;` equivalent to `bit[1] foo;`? In other words, singleton registers and single bits equivalent?

zachschoenfeld33:
This is something that should be clarified by a prospective Types working group if and when it is formed...

1ucian0:
if single qubits are also singleton registers `let concatenated = q[0] || q[3];` is allowed

Issue #137, opened 2021-02-05
Typing issues


blakejohnson:
This is meant to be kind of a meta-issue describing the set of issues that need to be resolved w.r.t. types in QASM3.

1. typed variables in `gate`s and `def`s
At some point we already removed the type identifiers from `gate`s, arguing that the compiler did not need this information to satisfy the need of having a mathematical representation of the meaning of `gate`. Meanwhile, we have kept them for `def`s. I would argue that the latter is important for consistency with `kernel` and `defcal` declarations. @ndebeaudrap, meanwhile, argued for dropping the types for `def` in his write-up attached to issue #130.

2. order of type and variable identifier
See issue #117.

3. array types
e.g. `qreg q[4]` vs `qubit[4] q` vs `qubit[4] q[10]`
or `bit[4] a` vs `bit a[4]`. This latter example relates to some of the discussion in @ndebeaudrap's write up where he suggested making `creq q[4]` and `bit[4]` synonyms for each other.

ndebeaudrap:
The issues raised by @steleman in #130 regarding conversions between different real-valued types (even between, say, `angle[16]` and `angle[8]`) may also need to be addressed. It seems to me that those remarks apply whether or not there is 'implicit' conversions, as they touch on the question of whether or not there is a well-considered (as opposed to just well-defined) way to go about doing any sort of typecasting between these types.

hhorii:
Regarding 3: `bit[4] a` vs `bit a[4]`

In #130, necessity of "return type" in subroutine `def` is discussed. I believe that a subroutine can return an array. If we keep "return type" and this "return type" includes its size, return type for array will be `-> bit[4]`.

zachschoenfeld33:
This is a good point @hhorii. If we use `type[BITS] id[ARRAY_SIZE]`, how do we denoted return types w/ `->`. We could do `-> type[BITS][ARRAY_SIZE]`, where the latter entry is implicit. 

Also, why would we allow `BITS` for qubit/qreg/bit/creg. It seems like they should always have a single bit.

blakejohnson:
Then how about an explicit array type? e.g. `array[bit[4]] a`

zachschoenfeld33:
That could work, although if this language should be C-like, we could define standard bit lengths for the main types and then use the simple array syntax `type name[SIZE]`. I am not as familiar with the importance of defining precision lengths for different types, so maybe that can't be done without. 

[...]

PR #283, opened 2021-09-29
Add arrays to the specification

mbhealy:
### Summary
Adds arrays to the language following examples from #137 and #218

### Details and comments
Includes examples of
- Initialization
- Multi-dimensional arrays
- Individual element assignment
- Access using index sets
- Bit-level access to additional classical types (int, uint, angle) using index sets
- Special optional syntax (#) for bit access

Still needs more discussion:
- Copying vs taking references
   - In general (let -> references, otherwise copying?)
   - For subroutine and extern arguments
   - Concatenation


steleman:
I am interested in learning the semantics of _multi-dimensional arrays_ in this context.

Does it imply  a maximum of two-dimensional arrays, or does it mean N-dimensional?

If it means N-dimensional -- i.e. more than two dimensions, there are some things to consider that I suspect have not been.

mbhealy:
> I am interested in learning the semantics of _multi-dimensional arrays_ in this context.
> 
> Does it imply a maximum of two-dimensional arrays, or does it mean N-dimensional?

Yes, we intend to support N-dimensional arrays if possible. Please let us know your concerns. I am sure there are many things we have not considered yet.

steleman:
> > I am interested in learning the semantics of _multi-dimensional arrays_ in this context.
> > Does it imply a maximum of two-dimensional arrays, or does it mean N-dimensional?
> 
> Yes, we intend to support N-dimensional arrays if possible. Please let us know your concerns. I am sure there are many things we have not considered yet.

I have been thinking about the AST representation of N-dimensional arrays within the current given constraints of the OpenQASM3 Language:

- Arrays are statically allocated
- There is no memory model
- There are no address spaces
- There are no pointers
- There is no offset-based addressing

One-dimensional and two-dimensional arrays can be represented within these constraints. In this case, a two-dimensional array is, in fact, a matrix.

I think this model breaks down at 3-dimensional arrays and higher.

What happens starting at 3 dimensions and higher is that, for every additional array index dereference [N] past the second one, the compiler must create a new _Type_. Since OQ3 arrays are value-based (as opposed as base + offset memory addressing as in C), each array index dereference must obey value semantics. The implication is that the compiler must somehow create N number of types, to accommodate for N-dimensional arrays whose elements can be dereferenced at any point of the dimension chain.

For example:

// 1. Assuming we have declared a 3-dimensional array 'a' of ints:
int i = a[3][5][7];

// 2. Assuming we have declared a 5-dimensional array 'b' of ints:
int j = b[3][5][7][9][11];

Question: In value-based semantics (not base + offset pointer semantics) is the _Type_ of a[3][5][7] of the same type as the _Type_ of b[3][5][7]?

C solved this problem by employing base + offset memory addressing of array elements. In that implementation, **only one** _Type_ is required to reference any element of an N-dimensional array, since all accesses are base + offset memory addressing based. Namely, this _Type_ is the pointer. A consequence of that design decision is that, in fact, N-dimensional arrays in C are, in fact, one-dimensional, and everything has the same type, and everything is accessed via base + offset addressing.

jwoehr:
> I think this model breaks down at 3-dimensional arrays and higher.

No point getting wrapped around the tools and losing sight of the goal.
If _n_D arrays are wanted, it can be made to work.

jakelishman:
I think Stefan's point is something we definitely want to have in mind when considering array interactions in general; we can for sure make them work, but we don't want to mistakenly specify them in a way that makes things like passing them to subroutines have major memory overhead or performance costs at runtime.

This was mentioned briefly in the TSC meeting today, and is going back to the types working group.  One method mentioned was to specify that n-dimensional arrays are implemented on hardware similarly to C (i.e. they decay to a pointer to the first element), but with a little bit of additional structure around them to make passing them to subroutines more convenient.  Essentially, all arrays would be represented by something (very approximately) like this C struct:
```c
struct ndarray {
    type_info type;
    size_t ndims;
    size_t *dimensions;
    size_t *strides;
    void *data;
}
```
where the `type_info type` entry represents information about the base type (I suspect that may be cleaner with C++ templates?). We could allow hardware implementors to have a maximum number of dimensions supported, if it would be easier for them to have the dynamic `size_t *` pointers become fixed length `size_t [MAX_DIMENSIONS]` items on compilation to those particular backends.  I _think_ that if we don't introduce pointers as a type in QASM itself, then we'd be able to statically know the dimensions of all arrays at all times, in which case it would be easy for tools to reject code that nests dimensions too high.

Is this something that could reasonably implemented by compilers?  I know that what I've written isn't a full spec, and we'd have to iron out several details, such as how arrays could be re-interpreted with different numbers of dimensions with explicit (or implicit, maybe?) casting.  But as a starting point, is it a model that could work in existing compiler infrastructure?

steleman:
> Is this something that could reasonably implemented by compilers? I know that what I've written isn't a full spec, and we'd have to iron out several details, such as how arrays could be re-interpreted with different numbers of dimensions with explicit (or implicit, maybe?) casting. But as a starting point, is it a model that could work in existing compiler infrastructure?

I had a brief talk with Michael earlier this morning and, at least in my view, things become much easier to handle if we allow reference and reference semantics.

The internal representation of arrays would be `base address + offset`, and arrays could be passed by reference to functions with side-effects, either as a `const` or a `non-const` reference.

Without the concept of type-agnostic references or pointers, please consider the following example:

```
array[array[int, 16], 16] ai;

array[int, 16] ki;

ai[9] = ki;
```
Line 3 of this example wants to modify the two-dimensional array named `ai` declared on line 1.

This small example above reduces to conversion from rvalue reference to lvalue reference, which then can be assigned to.

If we only allow value semantics, and disallow reference (or pointers) semantics, the assignment statement on line 3 will **not** have the semantics everyone expects: `ai[9]` is a rvalue reference which has been implicitly converted to a lvalue reference that can, and is, subsequently, assigned to.

Instead, `ai[9]` is a value-semantics based **copy** of the 10th element from `ai` that gets `ki` assigned to it, and at this point I can no longer wrap my head around what the semantics of this assignment expression even are. It certainly doesn't do what I suspect everyone would like it to do, which is to change the 10th element of the array `ai`. It operates on a value-based copy, and not on a reference. Therefore, `ai` is not modified. I don't think that's what we want.

blakejohnson:
@steleman is it not sufficient to treat any array access on the left side of an equals sign as a reference? It think you already need this even for just 1-d arrays in order for `ki[9] = 1` to have meaning.

blakejohnson:
Something else to consider for multi-dimensional arrays are the type reduction rules on slices. Can I assign to a slice (my answer: yes)? What if that slice isn't over the last dimension? Do the shapes need to match in the assignment, or is it sufficient to have the correct total number of elements?

steleman:
> @steleman is it not sufficient to treat any array access on the left side of an equals sign as a reference? It think you already need this even for just 1-d arrays in order for `ki[9] = 1` to have meaning.

In the case where only 1-dimensional arrays are supported, the semantics are very simple: `ki` is the array ,`ki[9]` is a  reference.

In the case of multi-dimensional arrays, there are multiple references:

- `ki[9]`
- `ki[9][3]`
- `ki[9][3][1]`

etc etc etc.

This reduces to my original post: the compiler would need to invent N number of references, and keep track of them. 

Even with this artifact, I still do not understand the semantics of assigning to a by-value copy of an array slice, and expecting the assignment to modify the original. That kind of semantics only exists with pointers.

References - aside from syntax sugar - are, in fact, pointers with constraints:

- References (unlike pointers) must alias a lvalue. Consequently, references are constrained by their type.
- References **cannot** perform _reference arithmetic_.
- References cannot be re-assigned.
- References cannot be re-typed.
- There is no universal catch-all **void&** reference.

Only one type of addressing is needed here, and only one type of addressing is required here: _the pointer_, as in memory address.

I do not see the benefit in making this orders of magnitude more complicated than it needs to be.

> Can I assign to a slice (my answer: yes)?

You can assign to a slice as long as the slice can be represented by a starting pointer. If it's not a pointer, it's a by-value copy, and you are assigning to the copy, not to the original, which is what I presume was the original intent.

> What if that slice isn't over the last dimension?

What would the types of the resulting ranges be?

For a multidimensional array such as:

```
array[array[array[array[int, 16], 16], 16], 16] ki;
```

The last slice are ints. The last before last slice are arrays of ints. The slice before that are arrays of arrays of ints.

Three different reference types already.

In reality, these are all addresses and address offsets. We need pointers for this. Pointers solve the addressing problem. They do not solve the memory layout problem.

> Do the shapes need to match in the assignment

I don't quite follow what _shape_ means in this context. I will assume it refers to the layout of one of the internal arrays. In that case:

```
array[array[array[array[int, 16], 16], 16], 16] ki;

array[array[int, 32], 64] zi;

ki[2][3][4] = zi;
```

will not work, because there is no room for assigning the double-array `zi` which is of a different size. It can be done if we have malloc/realloc/free semantics. And then, we will be implementing our own OpenQASM3 virtual machine. I am not saying this in jest by any stretch.

> or is it sufficient to have the correct total number of elements?

Having the correct total number of elements is not only sufficient, it's also necessary.

Currently, in the OpenQASM3 Spec, arrays are statically typed and statically allocated. That means their sizes are immutable.

If you would like to be able to change the memory layout of a multi-dimensional array on-the-fly, we need malloc/realloc/free semantics. And array sizes can no longer be immutable. Which, in turn, implies two things:

- An array always decays to a pointer to its first element. (OK, that's a requirement anyway at this point).
- There must be two types of memory allocation: static and dynamic.

If you decided to shrink the size of one of the internal arrays (via assignment), there is no practical way of representing the memory slots that have become unused. OQ3 - conceptually - does not have a runtime, it does not have a stack, it does not have a heap, and it does not have any of the other memory segments that construct the runtime memory image of a process.

There is no feasible way of _poisoning_ an integer value (pointers are in fact unsigned integers), unless there exists a runtime, a stack and a heap at a minimum.

Producing reliable _poisoned_ values requires all this machinery. We need to be able to point a poisoned address to an address that cannot possibly be valid, under any circumstances. OQ3 does not even have the concept of valid or invalid addresses.

Having an array descriptor - as it has been suggested above - is helpful, but it does not fully solve the underlying problem. Inside the array descriptor, aside from all the internal housekeeping metadata, there is / there will be a **void***. That's a memory address, and it represents the starting address of the array, regardless of its number of dimensions. Which, to me, indicates that the assumption that an array - regardless of its dimensions - implicitly decays to a pointer to its first element has already been made.

jakelishman:
Thanks for all the detailed explanation, Stefan - it's super helpful to have this laid out clearly, and I know I'll still need to go and think things through rather more to get a proper handle on them.

I think your comments make it fairly clear that we will need some sort of reference type in implementations of OpenQASM 3, and discussions in the types WG about the semantics of passing arrays to functions has been circling around the same idea as well (that's why I had a `void *` in my array descriptor - as you say, I'd assumed it would have to decay to a reference at some stage).

The other side of the coin is that I think people are generally keen not to introduce a user-facing reference operator (like C++'s `&`) into the language - the language is more for mathematical calculation than classical systems programming, and we'd like to hide those details at the user level if it's at all reasonable to do so.  At the moment we're trying to square those two things, if they're possible in a way that doesn't produce unexpected behaviour.  I need to read more about all the complexities oft this before I can have proper opinions, though.

blakejohnson:
> The other side of the coin is that I think people are generally keen not to introduce a user-facing reference operator (like C++'s `&`) into the language - the language is more for mathematical calculation than classical systems programming, and we'd like to hide those details at the user level if it's at all reasonable to do so.

Right. I think what we are after is closer to Fortran-style arrays in which the compiler has to internally handle memory address arithmetic but the concept of a _pointer_ is never surfaced to the programmer.

> This reduces to my original post: the compiler would need to invent N number of references, and keep track of them.

Yes, the compiler would need an extensible type for references that tracked all of the "trailing" dimensions of any particular reference.

> Even with this artifact, I still do not understand the semantics of assigning to a by-value copy of an array slice, and expecting the assignment to modify the original. That kind of semantics only exists with pointers.

If it is on the left hand side of an equals sign it wouldn't be copy, it would be a reference to a section of memory. As I think you've pointed out, you already need this even for `ki[9] = 1` to update `ki` in your example.

> References - aside from syntax sugar - are, in fact, pointers with constraints:

100% agree with this and all your following bullets.

> I don't quite follow what shape means in this context. 

Just an ordered list of the dimensions, i.e. for your examples 

`array[array[array[array[int, 16], 16], 16], 16] ki;` has a shape of 16 x 16 x 16 x 16.
`array[array[int, 32], 64] zi;` has a shape of 32 x 64

I think what I was arguing for was that assignments must have the correct number of elements on the right hand side. i.e. I suggest we don't use MATLAB/Numpy/Julia semantics of automatic broadcasting of assignments where you can do something like

```
array[int, 16] foo;
foo[0:15] = 1;
```
which in a language with broadcasting assignment would treat that last statement as if it were:
```
for i in [0:15] {
   foo[i] = 1;
}
```

Instead, in OpenQASM3 I suggest that this produce an error and the programmer has to provide the correct number of elements for assignment.

mbhealy:
> Instead, in OpenQASM3 I suggest that this produce an error and the programmer has to provide the correct number of elements for assignment.

100% agree with this statement.

steleman:
> I think what we are after is closer to Fortran-style arrays

I am the bearer of bad news.

- Fortran multi-dimensional arrays are limited to 7 dimensions. See my earlier comment about N number of reference types.
- Fortran has pointers, just like C (Fortran 90).
- It follows that Fortran has a memory model. We don't.
- Fortran has a runtime. A Fortran program loads in memory, and creates a process memory image with memory segments. We don't.

`array[array[array[array[int, 16], 16], 16], 16] ki; has a shape of 16 x 16 x 16 x 16.`
`array[array[int, 32], 64] zi; has a shape of 32 x 64.`

In C - which has **real** multi-dimensional arrays - the _shape_ of a one-dimensional array is no different than the _shape_ of a 9-dimensional array, for example. In C, they all have the same _shape_: **all** of them are **one-dimensional**.

You can think of C multi-dimensional arrays as the projection of a N-dimensional geometric object onto a two-dimensional plane.

The multi-dimensionality in C (and friends) is created by pointer arithmetic, and by the fact that a **void****** can implicitly convert to a **void***** who can implicitly convert to a **void**** who then implicitly converts to a **void***. The rest is done with pointer arithmetic.

Matlab/Numpy run inside a virtual machine. Numpy runs inside the Python virtual machine, and the Numpy multi-dimensional arrays are really Python wrappers around C.

Matlab emits C which can then be compiled to an ELF or PE/COFF (Windows).

We can choose not to expose pointer semantics to the user. But we still need pointer semantics internally, and pointer semantics imply a memory model.

blakejohnson:
> You can think of C multi-dimensional arrays as the projection of a N-dimensional geometric object onto a two-dimensional plane.

It's really just a one-dimension space: memory addresses. Everything beyond the first index simply refers to some address arithmetic given by an address stride.

mbhealy:
> > I guess I am thinking of these things as effectively for loops.
> > `twoD[1:2][0] = anotherTwoD[0:1][1] // does this make sense? 2x1 = 2x1?`
> 
> I _think_ this is an assignment to a scalar - namely to one of the **ints** inside the inner array. So this should be allowed.

I don't think this could be seen as assignment to a scalar. If `twoD[M:N]` is a 2D array, then `twoD[M:N][K]` should be a 2D array where the nested array is of type `array[int, 1]`.

 `array[int, 1]` must be a different type than `int`.

steleman:
> `twoD[M:N]` should be a 2D array of its own [ ... ]

What are the semantics of the slice then?

Is it a reference to the original array, or is it a by-value copy?

I think this has implications for the example you provided.

If `twoD[M:N][K] = <something>;` is expected to modify the original, that's reference semantics.

On the other hand, if `twoD[M:N][K] = <something>;` creates a by-value copy on-the-fly in the LHS, I don't get what it accomplishes. The by-value copy will go out of scope the moment the assignment statement terminates.

steleman:
> when we get a slice/index set inside the [] operator what we get does not have one of the dimensions removed.

So the expression `twoD[M:N]` evaluates to a select subset of `twoD`?

In that case, the assignment

```
twoD[M:N][K] = 5;
```

refers to a select subset of inner arrays of `twoD`, not to a scalar. In that case I can only assume it has reference semantics, for reasons already explained.

mbhealy:
> > when we get a slice/index set inside the [] operator what we get does not have one of the dimensions removed.
> 
> So the expression `twoD[M:N]` evaluates to a select subset of `twoD`?
> 
> In that case, the assignment
> 
> ```
> twoD[M:N][K] = 5;
> ```
> 
> is impossible, because it refers to a select subset of inner arrays of `twoD`, not to a scalar. In that case I can only assume it has reference semantics, for reasons already explained.

Correct and correct. That assignment is impossible. So then back to my original question. Does it make sense to assign to a slice that is _not_ the last dimension?
```
twoD[M:N] = anotherTwoD[O:P];
twoD[M:N][K] = anotherTwoD[O:P][L];
```
I feel like the answer should be yes as long as the shapes match. Not just the number of elements, but the shapes. Implementation will be ugly, but seems logically consistent to me.

steleman:
> > > when we get a slice/index set inside the [] operator what we get does not have one of the dimensions removed.
> > 
> > 
> > So the expression `twoD[M:N]` evaluates to a select subset of `twoD`?
> > In that case, the assignment
> > ```
> > twoD[M:N][K] = 5;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is impossible, because it refers to a select subset of inner arrays of `twoD`, not to a scalar. In that case I can only assume it has reference semantics, for reasons already explained.
> 
> Correct and correct. That assignment is impossible. So then back to my original question. Does it make sense to assign to a slice that is _not_ the last dimension?

Nope. It will create a giant amount of confusion and I am not sold on its benefits.

I am not sold on the concept of slices that change the semantics of the subscript operator either.

steleman:
> Discussion of Numpy and Python is very relevant;

It's not, for reasons already explained.

You are discussing the semantics of slicing multi-dimensional arrays when we don't even have a theoretical model for OpenQASM3 multi-dimensional arrays in the first place. Ant that is because we do not have a model for an OpenQASM3 Virtual Machine, and we don't have a Memory Model for OpenQASM3, either.

We need one of those two. This is not a nice-to-have, or a pet project of mine. This is a requirement.

Without one of those two, multi-dimensional arrays cannot be implemented. It's really that simple.

Can you point us to a translation unit in the C numpy implementation in support of your earlier statement that _numpy arrays do NOT allocate memory_?

I've already provided you the links to the numpy source code, and the documentation describing numpy's Memory Model.

What matters here to me is _what happens inside Python_, **not** what Python decides to print in the interactive shell, or what Python looks like to someone writing Python scripts.

The numpy C code tells me what happens inside Python and numpy arrays. They allocate dynamic memory, inside the Python Virtual Machine.

steleman:
> Going back to @jakelishman's proposal, I don't follow this restriction:
> 
> > all arrays must be declared statically, in the global scope of the programme, and arrays cannot be declared within subroutines; there is no dynamic memory allocation.
> 
> Certainly I see the problem if an array within a subroutine has a size determined by a runtime value, but why is this an issue if the size is a compile-time constant? Wouldn't that just become stack allocation?

The problem is that there is no stack, just like there is no heap. Currently, there is no Memory Model in OQ3.

Also: the mechanism model for the execution of a subroutine is currently unspecified in the OQ3 Spec. There is no calling convention, and it is unspecified if OQ3's calling convention is callee-saved or caller-saved.

It is also left unspecified in the OQ3 Spec if arrays passed as arguments to a subroutine are passed by-value -- i.e. copied onto the subroutine's hypothetical _stack_ -- or they are passed by pointer decay to their first element. Based on my current reading of the Spec, it appears that all subroutine arguments are expected to be passed by value, i.e. copied. That includes arrays.

Also, if we assume that everything is declared on, and copied onto, this hypothetical _stack_, the maximum size of a stack frame is always an implementation-dependent limit. Depending on the compiler, and depending on the hardware the program is compiled for, this maximum stack frame size limit can, in certain cases, be overridden at compile-time. Other times it cannot. There is no guarantee that the maximum size of a stack frame can be changed on any ISA. Some ISA's hardcode their maximum stack frame size.

Which is why Fortran compilers always implement an option equivalent to GFortran's **-fno-stack-arrays** or **-fmax-stack-var-size=\<size\>** or Intel Fortran Compiler's **-heap-arrays**. These options are countermeasures to avoid corrupting the stack when declaring on stack multi-dimensional arrays whose size exceeds the maximum stack frame size.

If we do not want to have to deal with these hardware-imposed limits on the maximum size of a stack frame, we must use a Virtual Machine, which has the advantage of hiding all these hardware-dependent details. VM's perform their own stack frame allocations and deallocations inside dynamically allocated managed memory. At the expense of significantly increasing the implementation's complexity. 

I am not advocating for one type of model over the other (VM vs. stack/heap model). Both are very difficult to implement correctly. I am only trying to explain some of the numerous complexities involved in implementing multi-dimensional arrays and subroutine calls.

source/language/types.rst
```diff
@@ -359,3 +410,94 @@ variables whose values may only be known at run time.
...
+.. _array-slicing:
+
+Array concatenation and slicing
+-------------------------------
+
+Two or more classical arrays can be
+concatenated to form an array of the same type whose size is the
+sum of the sizes of the individual arrays. Unlike with qubit registers, this operation
+copies the contents of the input arrays to form the new (larger) array. This means that
+arrays *can* be concatenated with themselves.
```
jakelishman, line 457: If we allow arrays of arbitrary length to be passed to subroutines (e.g. `def f(array[int, dim(1)] v)`), then we need to forbid the concatenation operator from being used in a subroutine parameter list.

mbhealy: > we need to forbid the concatenation operator from being used in a subroutine parameter list.

This is confusing to me because I was not 100% clear on the language here until I looked it up. Are you referring to the parameter list associated with the subroutine definition, or the argument list associated with the subroutine call?

mbhealy: Based on our discussion during the last WG meeting I'm assuming you mean the argument list associated with the call.

jakelishman: Yeah, in the call, sorry.  I mean that
```text
def f(const array[int, ...] x) { }

f(a || b)
```
is an invalid call, because `a || b` doesn't resolve to a memory location.  My whole comment wasn't very well written, sorry.  This _can_ be forbidden grammatically, but it doesn't necessarily need to be; it can also just fail during semantic analysis.  The main point is that it's a caveat on passing arrays to function bodies.

mbhealy: Added in https://github.com/Qiskit/openqasm/pull/283/commits/b05dc4bbe35ba6dee930c10e5d2f7af46860edcc

1ucian0:
This PR is ready for review and voting by the TSC.

source/language/types.rst
```diff
@@ -268,6 +268,56 @@ are all of type ``float[64]``.
...
+   multiDim[0][0] = 0.0; // multiDim == {{0.0, 1.2}, {2.1, 2.2}, {3.1, 3.2}}
+   multiDim[-1][1] = 0.0; // multiDim == {{0.0, 1.2}, {2.1, 2.2}, {3.1, 0.0}}
+
+Arrays *cannot* be declared inside the body of a function or gate. All arrays
+*must* be declared within the global scope of the program.
+Indexing of arrays is n-based *i.e.*, negative indices are allowed.
+Multi-dimensional arrays (as in the example above) are allowed, with a maximum
+of 7 total dimensions.
+
+For backwards compatability, the standard
```
ndebeaudrap, line 299: For 'interoperability' with quantum registers, etc.? ("Backwards compatibility" suggests that this syntax was present in OpenQASM2, which it wasn't.)

ndebeaudrap (dismissed):

source/language/types.rst
```diff
@@ -268,6 +268,56 @@ are all of type ``float[64]``.
...
+   float[32] lastLastElem = multiDim[2][1]; // 3.2
+   float[32] alsoLastLastElem = multiDim[-1][-1]; // 3.2
+
+   myArray[4] = 10; // myArray == {0, 1, 2, 3, 10}
+   multiDim[0][0] = 0.0; // multiDim == {{0.0, 1.2}, {2.1, 2.2}, {3.1, 3.2}}
+   multiDim[-1][1] = 0.0; // multiDim == {{0.0, 1.2}, {2.1, 2.2}, {3.1, 0.0}}
+
+Arrays *cannot* be declared inside the body of a function or gate. All arrays
+*must* be declared within the global scope of the program.
+Indexing of arrays is n-based *i.e.*, negative indices are allowed.
```
1ucian0, line 295: clarification on the meaning?

source/language/types.rst
mbhealy: Changed in https://github.com/Qiskit/openqasm/pull/283/commits/7908b8aa4ebebabb4a58e256b4c5216951a6f2e7

source/language/types.rst
mbhealy: Addressed in https://github.com/Qiskit/openqasm/pull/283/commits/abd29161b70d7969e00476f828ceb584d38c41d1

mbhealy:
The grammar is ready and merged with this PR, so it is fully ready for final review and merge.

examples/arrays.qasm
```diff
@@ -0,0 +1,101 @@
...
+
+// Initialise an array using a row of another array.  Note that this copies the
+// data; changes to the values in `my_defined_float_row` will not affect
+// `my_defined_floats`.
+array[float[32], 2] my_defined_float_row = my_defined_floats[0];
+
+// Arrays can be defined in terms of compile-time constant dimension sizes, and
+// can use expressions in the slots of their initialisers.
+const int[8] DIM_SIZE = 2;
+array[int[8], DIM_SIZE, DIM_SIZE] all_ones = {{2+3, 4-1}, {3+8, 12-4}};
```
blakejohnson, line 34: Does this imply an implicit cast of `DIM_SIZE` from `int[8]` to `int`?

jakelishman: Do you mean when it's used as the second and third items in the array type declaration?  It's a good question - it's the same problem as asking what the type of `<SIZE>` must be in declarations like `qreg q[<SIZE>];`.  I think we haven't answered it in the types group yet.  I can change it to a generic sized `int` (actually `uint` - `int` can't be right!), but perhaps we'll allow any width in those locations; it's not something that should be an issue at runtime.

It's also on my list of examples of "stuff that needs to work for OpenQASM 2 compatibility" - you need to be able to use an integer literal with no suffix.

source/grammar/tests/invalid/statements/declarations.qasm
```diff
@@ -31,6 +31,12 @@ complex[creg] myvar;
 complex[qreg[8]] myvar;
 complex[creg[8]] myvar;
 
+// Bad array specifiers.
+array myvar;
+array[8] myvar;
+array[int, 8] myvar;  // (no width for the integer)
```
blakejohnson, line 37: Why does this fail? Seems like this should work taking the compiler's definition of `int`.

source/language/subroutines.rst
```diff
@@ -70,4 +70,65 @@ this
...
+dynamically allocated the memory associated with the array stays intact after
+subroutine exit. Additionally, the OpenQASM3 language is not anticipated to
+support explicit user-controlled creation of pointers and references outside
+of the specific context of passing arrays to subroutines.
+
+The dimensions of arrays may be queried inside of subroutines using the built-in
+``sizeof()`` function, which takes two parameters: the array being queried, and
+the zero-based dimension number requested. If the second parameter is omitted,
+then it defaults to ``0``, *i.e.* ``sizeof(arr) == sizeof(arr, 0)``.
+``sizeof()`` returns a ``const int[32]`` representing the length of the
```
blakejohnson, line 118: Wouldn't returning an unsigned type from `sizeof()` be more natural? i.e. `const uint[32]` or even `const uint`?

source/grammar/tests/invalid/statements/declarations.qasm
jakelishman: At the moment the ANTLR grammar doesn't permit the statement `int x;` with no width, so this is just consistent with that.  I don't remember whether the idea of "generic/default int" is present in the spec at the moment, but if it is, then we should update this in a separate PR - there's a bunch more of the grammar that would need updating too.  I think it's not there yet, though it's something we've been discussing.

blakejohnson (commented):
Looks good. My only open issues are:
* Should the return type of `sizeof()` be an unsigned type?
* Seems like declaring arrays with compiler-default value widths should be allowed, e.g. `array[int, 8]` should be valid

source/grammar/tests/invalid/statements/declarations.qasm
blakejohnson: This seems to be something we neglected to translate over from the paper (see section 4.4 of https://arxiv.org/abs/2104.14722)

examples/arrays.qasm
blakejohnson: Yes, I was referring to the second and third items in the array type declaration. Let's just change the `DIM_SIZE` definition to `uint` and we can address widths later.

jakelishman:
@blakejohnson: with 967a29d, which changes the signed integer used in a dimension constant and removes the questionable "invalid" tests, I think we've addressed everything we said we would in the types group meeting.

For others: the point about the spec and the ANTLR grammar not currently allowing types of arbitrary width (e.g. `int a;`) will be the subject of a different PR.  We strongly expect they will be allowed - they were in the original QASM 3 paper - but the types group is discussing exactly what the semantics of such a type would be before making a PR (#276 exists, but is incomplete).

levbishop (dismissed):
I have a couple small open questions, but this PR has become unwieldy and we have consensus on the broad approach, so I approve with the caveat to revisit some aspects post merge.

levbishop (approved):

PR #285, opened 2021-10-06
Correct spec examples of subroutine definitions

jakelishman:
### Summary

The specification still had the old form of subroutine definitions
listed, where qubit arguments were specified separately to the classical
ones.  This was outdated by commit cc9bbc7, and this commit brings the
spec in line with both the grammar and the paper.

### Details and comments

Judging by Steven's comment in #217 (https://github.com/Qiskit/openqasm/pull/217#issuecomment-852476827), this change happened last-minute as the paper went out the door, and that's presumably how the discrepancy arose.

Related to #284.

### DIFF
--- a/source/language/subroutines.rst
+++ b/source/language/subroutines.rst
@@ -1,32 +1,30 @@
 Subroutines
 ===========
 
-Subroutines are declared using the statement ``def name(parameters) qargs -> output { body }``.
+Subroutines are declared using the statement ``def name(parameters) -> output_type { body }``.
 Zero or more quantum bits
-and registers are passed to the subroutine by reference or name in ``qargs``.
-Classical types are passed by value in ``parameters``. The parentheses may be omitted if no
-``parameters`` are passed. The subroutines return up to one value of classical type, signified by the
+and classical values are passed to the subroutine by reference or name in ``qargs``.
+Classical types are passed by value in ``parameters``.
+The subroutines return up to one value of classical type, signified by the
 ``return`` keyword. If there is no return type, the empty ``return``
 keyword may be used to immediately exit from the subroutine. All arguments are declared together
 with their type, for example ``qubit ancilla`` would define a quantum bit argument named ``ancilla``.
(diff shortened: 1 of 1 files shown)


jakelishman:
This PR itself isn't a change to the language (it's just bringing the documentation up-to-date with the paper and the reference grammar), but I think originally it was changed to make calls to `def`- and `extern`-defined functions look the same.  The justification for treating qubits differently here compared to in gates seems to be because in `def` subroutines, the broadcast rules of `gate` don't apply.

I can't give a definitive answer though, sorry - the real change happened before I joined IBM.  Perhaps @stevenheidel remembers the full context?

steleman:
> This PR itself isn't a change to the language [ ... ]

When you are making a change to the Language Grammar, you are making a change to the Language itself.

The rationale behind placing QArgs outside parameters was - at the time - that _Function Declaration Syntax should be consistent with Gate Declaration Syntax_.

Formally, the rationale for this particular form of syntax was the avoidance of context-dependent type analysis. "Classical" type parameters were passed inside parens (if any), and "Quantum" type arguments were passed outside parens (QArgs). This specific placement of the two types of formal parameters made context-dependent type analysis unnecessary.

Now that both types of formal parameters are passed inside parens, we need to do context-dependent type analysis. Which is a fancy way of saying "the compiler has to guess what was meant here".

Are there any ordering rules that apply to the placement of parameter types inside parens? I.e. "Classical" parameter types first, followed by "Quantum" types? Or can they be freely mixed?

bettinaheim:
@steleman I am not entirely sure I understand that statement regarding the analysis becoming context dependent; implementation details aside, if the type system is expressive enough to clearly distinguish which types are classical and which ones are quantum, I don't believe any context information would be required to make the split automatically. To me, the syntax distinction to me feels like moving something part of what I would usually expect to be reflected in the type system into the grammar. Am I misunderstanding your comment, or am I missing something about the implementation or future evolution of the type system?

PR #288, opened 2021-10-11
Create Python package out of reference AST

jakelishman:
* Add packaging structure to reference AST

  This creates a full Python package (called `openqasm3`) out of the reference AST, parser and visitor, so it can be installed through `pip` in the normal manner, and distributed through PyPI.

  The `test_openqasm_tests.py` file is not currently included when generating the sdist, because it relies on files that are elsewhere in the repository, and can't be brought into this package without significant duplication.  This may be changed in the future, but for now, the test file is useless on its own.

  The minimum Python version is fixed at Python 3.7; to support Python 3.6, there would need to be a (3.6-specific) dependency on a backport of `dataclasses`, but also the `from __future__ import annotations` does not exist, so the entire `ast.py` file would need re-ordering into a bottom-up approach, which harms legibility for little gain.

* Run AST tests on full Python matrix

  This also changes the tests to ensure that the package builds and installs as expected.

* Add basic documentation for Python package

  This is a very basic first pass at the documentation for the reference Python package.  It sets up a very simple Sphinx configuration, and puts in references to all the constructed elements.

  The styling of the output is not especially nice yet (it all feels a bit squashed together), but graphic design is pretty far out of my skill set, so it's probably best for someone else to take a pass at that.

### DIFF
--- a/.github/workflows/tests-ast.yml
+++ b/.github/workflows/tests-ast.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Generate grammar
         working-directory: source/grammar
-        run: java -Xmx500M -jar antlr-${{ matrix.antlr-version }}-complete.jar -o ../openqasm/parser/antlr -Dlanguage=Python3 -visitor qasm3.g4
+        run: java -Xmx500M -jar antlr-${{ matrix.antlr-version }}-complete.jar -o ../openqasm/openqasm3/antlr -Dlanguage=Python3 -visitor qasm3.g4
 
       - name: Install Python packages for AST
         run: python -mpip install -r requirements.txt -r requirements-dev.txt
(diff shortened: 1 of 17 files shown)


aspcompiler:
I have pinned some collaborators to seek their thoughts on the directory structure. I noticed that you renamed qasm_parser to parser. We originally named it qasm_parser to leave room for pulse_parser. Like to see your thoughts on the name/path for the pulse parser.

jakelishman:
If the AST module is called `openqasm3.ast`, then it makes sense for the parser's module to be called `openqasm3.parser`.  In that situation, the name `qasm_parser` would be superfluous, and a bit confusing - we use a general name for the AST but a specific one for the parser.  The OpenPulse AST probably shouldn't be part of the OpenQASM 3 reference AST itself - with `defcalgrammar` taking any string, a fully compliant OpenQASM 3 parser doesn't _need_ to accept OpenPulse (though in practice there's currently no alternative), and this is meant to be a reference implementation (rather than the absolutely most performant).

Assuming that we _will_ also want to make a reference OpenPulse grammar, and distribute it as part of this package, I'd think it should be a full subpackage `openqasm3.openpulse` with its own `.ast` and `.parser`, which feels like a better namespacing split to me as well.

If we want to change the spec to say that "all valid implementations of OpenQASM 3 _must_ support OpenPulse", then I'm fine mixing them more, but at the moment that's not what we've got.  (Also in that situation, personally I'd still call the module `parser`, I'd just have it handle both.)

jakelishman:
I've just updated this to fix a couple of minor merge conflicts that had arisen, so from my side it should be good to merge.

There will be a little more merge conflict with the arrays PR #283. Since this is simpler and the array PR is on the verge of merging (fingers crossed), I think that we should merge this PR _after_ the array PR.  Please still review now, though.

jakelishman:
I've now rebased this over the arrays PR (#283) which merged earlier today.  @stevenheidel, @aspcompiler, @shiyunon: please could you make sure you're happy with the packaging from the AWS side now, because this is ready to merge, subject to review.

stevenheidel (approved):

stevenheidel:
@jakelishman Approved!

Apologies for the delay, I missed the notification in the flood of emails I get. In the future if you're ever blocked for more than a few days by my review then don't be afraid to ping me on Slack.

PR #289, opened 2021-10-14
Added pragma node to the reference AST

godott:
### Summary
- Added Pragma node in the AST
- Added test cases

### Details and comments
See above

### DIFF
--- a/source/openqasm/parser/antlr/qasm_parser.py
+++ b/source/openqasm/parser/antlr/qasm_parser.py
@@ -50,6 +50,7 @@ from openqasm.ast import (
     IntType,
     IODeclaration,
     IOKeyword,
+    Pragma,
     Program,
     QASMNode,
     QuantumArgument,
--- a/source/openqasm/tests/test_qasm_parser.py
+++ b/source/openqasm/tests/test_qasm_parser.py
@@ -36,6 +36,7 @@ from openqasm.ast import (
     IntType,
     IODeclaration,
     IOKeyword,
+    Pragma,
     Program,
     QASMNode,
     QuantumArgument,
@@ -1096,4 +1097,26 @@ def test_end_statement():
     """.strip()
     program = parse(p)
     assert program == Program(statements=[EndStatement()])
+    SpanGuard().visit(program)
+
+
+def test_pramga():
+    p = """
+    #pragma {verbatim;}
+    #pragma {my_statement1; my_statement2;}
+    end;
+    """.strip()
+    program = parse(p)
+    assert program == Program(
+        statements=[
+            Pragma(statements=[ExpressionStatement(Identifier("verbatim"))]),
+            Pragma(
+                statements=[
+                    ExpressionStatement(Identifier("my_statement1")),
+                    ExpressionStatement(Identifier("my_statement2")),
+                ]
+            ),
+            EndStatement(),
+        ]
+    )
     SpanGuard().visit(program)


shiyunon:
> With whitespace ignored, this change will just have ANTLR eat up every remaining token in the file the first time a pragma is seen - at the very least it needs to have a fixed, unalienable terminator. C/C++ has the line feed for that, as it allows any text, as far as I'm aware - it doesn't need to be tokenised into valid C tokens.
> 
> It's probably best to leave this for the time being - there's an agenda item for the TSC tomorrow to discuss forming a working group to standardise them a bit more, which may include fixing the minimal set of syntax rules needed to make this sort of change safe.

I added unit tests for the grammar and it seems to work (it did not eat rest of the program). But I agree we should wait for the TSC meeting to change the grammar. Thus, I kept the grammar and just added a Pragma node in the AST, that's a change we will need anyway no matter what.

jakelishman:
I was just typing a response, having thought about this more overnight.  Because it's a nongreedy match, it's ok in the situations in the unit test, but we're basically just encoding a huge conflict into the grammar here.  There are a few problems:

1. ANTLR will eat up the entire rest of the program into the pragma if any syntax later is invalid, because even non-greedy rules will never allow a program that _can_ be valid to fail parsing, and the `.*?` rule means that anything after a pragma that can be tokenised will be allowed:
    ```qasm
    #pragma __my_directive__
    const pi3 = pi / 3;
    x $0;
    == & || = +-;;;;;;;
    ```
    is a valid OpenQASM 3 program with this rule.

2. `.*?` in a parser rule means that it will consume all lexed tokens.  It can't consume something that won't lex, so `#pragma 3a;` will parse as `[Pragma("3"), ExpressionStatement(Identifier("a"))]` - the `3a` isn't consumed correctly.

3. This encodes a big conflict, which means that if a pragma content could be a valid QASM 3 statement, it will get parsed as a statement and lost from the pragma.  For example, `#pragma pi2 = pi/2;` will get parsed as an empty pragma followed by a classical assignment.  More insidiously, consider this test:
    ```python
    def test_pragma_with_trailers():
        p = """
        float[64] pi3;
        #pragma const
        pi3 = pi / 3;
        end;""".strip()
        program = parse(p)
        assert program == Program(
            statements=[
                ClassicalDeclaration(
                    type=FloatType(size=64),
                    identifier=Identifier("pi3"),
                    init_expression=None,
                ),
                Pragma(string_literals=["#pragma const"]),
                ClassicalAssignment(
                    lvalue=Identifier("pi3"),
                    op=AssignmentOperator["="],
                    rvalue=BinaryExpression(
                        lhs=Identifier("pi"),
                        op=BinaryOperator["/"],
                        rhs=IntegerLiteral(3),
                    ),
                ),
                EndStatement(),
            ]
        )
        SpanGuard().visit(program)
    ```
    
    This test will fail - it's parsed as:
    - declaration
    - empty pragma
    - const declaration
    - end statement
    
(As a side note, we need bigger integration tests of stuff like this as well - it's important that _following_ rules are parsed correctly.)

shiyunon:
> I was just typing a response, having thought about this more overnight - my initial comment was wrong because I wasn't thinking about the non-greediness. Because it's a non-greedy match, it's ok in the situations in the unit test, but we're basically just encoding a huge conflict into the grammar here. There are a few problems:
> 
> 1. ANTLR will eat up the entire rest of the program into the pragma if any syntax later is invalid, because even non-greedy rules will never allow a program that _can_ be valid to fail parsing, and the `.*?` rule means that anything after a pragma that can be tokenised will be allowed:
>    ```qasm
>    #pragma __my_directive__
>    const pi3 = pi / 3;
>    x $0;
>    == & || = +-;;;;;;;
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    is a valid OpenQASM 3 program with this rule.
> 2. `.*?` in a parser rule means that it will consume all lexed tokens.  It can't consume something that won't lex, so `#pragma 3a;` will parse as `[Pragma("3"), ExpressionStatement(Identifier("a"))]` - the `3a` isn't consumed correctly.
> 3. On the opposite side of the "syntax errors become allowed programs" issue, if a pragma content _could_ be a valid QASM 3 statement, it will get parsed as a statement and lost from the pragma.  For example, `#pragma pi2 = pi/2;` will get parsed as an empty pragma followed by a classical assignment.  More insidiously, consider this test:
>    ```qasm
>    float[64] pi3;
>    #pragma const
>    pi3 = pi / 3;
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    It looks like it should get parsed as a declaration, followed by assignment with a pragma directive saying it's actually a constant.  Instead, the test would fail - it's parsed as:
>    
>    * declaration (`float[64] pi3;`)
>    * empty pragma (`#pragma`)
>    * const declaration (`const pi3 = pi / 3;`)
>    
>    which is semantically invalid.
> 
> (As a side note, we need bigger integration tests of stuff like this as well - it's important that _following_ rules are parsed correctly.)

The last two examples are very interesting. I think it will be a good topic for the TSC meeting. Yes, I agree with your points and have already revert the grammar back after your initial comment. I just added a Pragma node in the AST.

PR #292, opened 2021-10-15
Make const a type modifier instead of a type

mbhealy:
### Summary
The `const` keyword in the current spec is defined as a type represented as a double precision floating point value that cannot be changed. This PR updates the `const` keyword to instead be a type modifier, more like its usage in C.

### Details and comments
In our discussion of the addition of arrays to the spec (#283), which are generally assumed to be of fixed length, the TAC WG has been assuming that it will be possible to define const integers to be used for defining multiple arrays with the same length. e.g.
```
const int[32] myArraySize = 10;
array[int[32], myArraySize] myIntArr;
array[float[32], myArraySize] myFloatArr;
array[bit, myArraySize] myBitArr;
array[qubit, myArraySize] myQubitArr;
```
The PR does not currently update anything but the textual spec. Grammar and test updates could be added before merge if desired.

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -215,11 +215,11 @@ be true and 0 will be false.
    // Assign a cast bit to a boolean
    my_bool = bool(my_bit);
 
-Real constants and const values
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Const values
+~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


blakejohnson:
💯

blakejohnson:
Please update the reference grammar as well.

PR #295, opened 2021-10-15
Added power operator for `complex` type

AbeerVaishnav13:
### Summary
Added scope for having power operator: `**`, for `complex` types in the language specification.

### Details and comments
Had a short discussion with @mbhealy and @jwoehr, and concluded that the cost of allowing the power operator is trivial, and added it for the sake of completeness of operations on the `complex` type.

### DIFF
--- a/source/grammar/tests/reference/declaration/complex.yaml
+++ b/source/grammar/tests/reference/declaration/complex.yaml
@@ -119,7 +119,7 @@ reference: |
                     32
                 ]
             ]
-            b
+            c
             equalsExpression
               =
               expression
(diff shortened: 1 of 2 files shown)


mbhealy:
In line with the recent TSC decision to keep the spec and grammar in sync as much as possible, would you also update the ANTLR grammar to support this change? Please also add a test that covers this as well.

AbeerVaishnav13:
Hi @jakelishman ,
> If everyone's happy, I'll merge this PR as it stands.

I'm fine with merging this PR as it is right now.
Thanks again for all your inputs and the detailed discussion on this subject!

blakejohnson (approved):

Issue #298, opened 2021-10-28
Specification needs update for built-in math functions


1ucian0:
The grammar was extended with some built-in math functions that were out of the specification. TSC decided that the functions included in https://github.com/Qiskit/openqasm/pull/275 should be part of the specification.

jwoehr:
@jakelishman any suggestion where such material should go in the document> `classic.rst`? `types.rst`?
Or is a whole new section on math expressions required?

jakelishman:
Shall we put the section inside the ["Const values"](https://qiskit.github.io/openqasm/language/types.html#const-values) part of `types.rst` for now?

At the moment, these builtins only apply to constant values.  I think all the types sections may expand into a few pages as the types group does its work, but I guess let's try not to predict things, and just stay in the same place for now.  I think if we put it anywhere else, we'll start to give the impression that they apply to variables at runtime, and the mechanics of how that might work haven't been discussed yet.  We don't even know if any of them will be available at runtime - perhaps that'll be a job for header files in the future?

athirab751:
Hi! I'm Athira Babu, Data Scientist at IBM Kochi with a PhD in Mathematics, participating in IBM Hacktoberfest 2026. I'd love to work on this issue — documenting mathematical functions is directly in my area of expertise.

Based on the TSC discussion, I understand the task is to add a list of compile-time built-in math functions to the "Const values" section of types.rst. I'll read Issue #275 for the full function list and submit a PR.
Could someone confirm this is still the correct approach, and please assign this issue to me? Thank you!

jwoehr:
@athirab751 ... possibly more useful these days is that we have a backlog of PRs that are very complex and which your mathematical skills might find application! May I suggest you look at PRs like #666 and comment them? It could be very helpful and would also orient you to the minutiae that the Technical Steering Committee needs to address in firming up the Specification.

athirab751:
Thank you so much @jwoehr! I just went through PR #666 - it's a fascinating discussion with some interesting mathematical questions around type casting semantics. I've left a comment there with my thoughts.

I'd be happy to look into other PRs in the backlog too wherever my mathematical background might be useful. Just point me in the right direction.

PR #300, opened 2021-11-02
Added built-in math functions used in constant initialization

jwoehr:
### Summary
Making spec conform to changes in #275 
Addresses #298 

### Details and comments
Table added.
Minor format corrections to `types.rst` in passing.


source/language/types.rst
```diff
@@ -267,6 +267,38 @@ are all of type ``float[64]``.
...
+      | exp    |
+      +--------+
+      | ln     |
+      +--------+
+      | sin    |
+      +--------+
+      | sqrt   |
+      +--------+
+      | tan    |
+      +--------+
```
jakelishman, line 300: We're missing `pow`, `floor`, `ceiling`, `div`, `mod` and `log` that are listed in the above paragraph at the moment.  Some things we'll need to mention:

- `pow`: for fractional powers, we need to say that it returns the "principal root", and specify how many arguments it takes (usually two, but sometimes a third, which is a modulus - I think we should only have the two-argument form).  It's questionable what input and output types this function should allow, given that it only has one name but `pow(-1.0, 0.5)` would be complex.  Perhaps for now it's best to say it's got overloads such that it allows
  - `pow(int, unsigned int) -> int`
  - `pow(float, int or float) -> float`
  - `pow(complex, int or float or complex) -> complex`
  
  and anything else is an error.

- `floor`, `ceiling`: these only make sense if they take a `float` argument.  Conceivably, they could return either a `float` or an `int`.  I think it's safest to return a `float`, and require casting to `int`; if you can represent the number `a.b` for digits `a` and `b` at a given floating-point precision, it's guaranteed that you can also exactly represent the integers `a` and `a + 1` in the same floating-point precision, so there's no risk there.

- `div`: not sure what was intended by this function when it was originally written into the spec.  In C it returns a struct of the quotient and remainder for efficiency reasons, but there's no need for that at compile time.  Possibly we should remove this from the list, but we can solicit suggestions.

- `mod`: assuming we mean modular division, we should have overloads for
  - `mod(int, int) -> int`
  - `mod(float, int or float) -> float`

  and we need to be sure to define what happens when any combination of the two arguments are negative.  Modular division doesn't seem (to me) to make sense on complex numbers (though we likely do want to include an `abs` function, which calculates the complex modulus).

source/language/types.rst
```diff
@@ -267,6 +267,38 @@ are all of type ``float[64]``.
...
+      | arcsin |
+      +--------+
+      | arctan |
+      +--------+
+      | cos    |
+      +--------+
+      | exp    |
+      +--------+
+      | ln     |
```
jakelishman, line 293: The above paragraph also mentions the function `log`.  We should define what the logarithm bases are.  I think in keeping with our verbose approach, perhaps we should have `log` as base-e log, not have `ln`, and if there's a very strong use-case for them, call base-10 or base-2 log `log10` or `log2` respectively.  They shouldn't generally be necessary, though, since `log(x) / log(y)` is the logarithm of `x` in base `y` in no matter what base `log` is.

The minimal expected overload would be `log(float) -> float`.  Conceivably, we could allow `log(complex) -> complex` as well, in which case we need to mathematically define how the branch cut is handled.

source/language/types.rst
```diff
@@ -267,6 +267,38 @@ are all of type ``float[64]``.
...
+   .. table:: Built-in mathematical functions in OpenQASM3.
+
+      +--------+
+      | arccos |
+      +--------+
+      | arcsin |
+      +--------+
+      | arctan |
+      +--------+
+      | cos    |
```
jakelishman, line 289: For the trig functions, I'd expect the allowed overloads to be

- `cos(float) -> float`
- `cos(angle) -> float` (danger)

and similar for other functions.  We should specify that `cos(float)` etc take their angle in radians.  _However_, it's not at all clear to me what the precision of the output `float` should be in the potential `cos(angle) -> float` overload; I'd expect the function to work, but I'm not sure our current type system can support it.  Perhaps best to omit it for now.

source/language/types.rst
```diff
@@ -267,6 +267,38 @@ are all of type ``float[64]``.
...
+      +--------+
+      | exp    |
```
jakelishman, line 291: `exp` has one argument, and likely should accept any numeric value, including a complex.  Its overloads are probably best as `exp(float) -> float` and `exp(complex) -> complex`.  There's no branch-cut shenanigans necessary with this one.

source/language/types.rst
```diff
@@ -267,6 +267,38 @@ are all of type ``float[64]``.
...
+      +--------+
+      | ln     |
+      +--------+
+      | sin    |
+      +--------+
+      | sqrt   |
```
jakelishman, line 297: `sqrt(float) -> float` (error on negative) and `sqrt(complex) -> complex`, and both should comment that they take the principal root, I would think.

jakelishman (commented):
I've left some comments inline about more details we can add.  Sorry there's a lot of text - there's a lot on my mind with this, and I might not have thought of everything yet.

I think for each function, we need to specify how many inputs it takes, what they mean, and some details about the output type.  I still have many questions about some details here (which is part of why I was keen to start trying to write them out!), so I'm very open to changing a lot of what I've said in my musing comments.

In general, these functions are going to need to support more than one set of input types.  This would be invalid in run-time OpenQASM 3 right now because of the "no implicit casting" rule, but since at the moment these are only valid at compile-time on constants, it might be clearest if we describe them as having a few allowed "overloads", and clearly define the types of each one.  That's not something OpenQASM 3 supports for `def` or `extern` statements, but we're only talking about convenience functions for defining compile-time constants here, so it might be something we want.  It would be good to hear other opinions on this - I will also bring it up in the types WG on Thursday.  In the comments, I'll refer to the overloads I'd expect to work, but just a note that this isn't final language, and depending on discussions, we may want to change that wording / further restrict the allowed types.

For now, I've generally kept all single-argument functions having the same input and output type.  That means, for example, that `exp(2)` is forbidden but `exp(2.0)` is allowed.  I did it this way because I don't want to get an inconsistency, where `sqrt(1.0)` is a `float`, but `sqrt(-1.0)` became `complex`.  My current suggestions make the syntax more restrictive than I'd like (I'd like `exp(2)` with a literal "2" to "just work" if it were possible), but my current idea for how to get to this situation isn't oven-ready yet, and I don't know if it will ever be.  So I've stayed conservative for now.

It seems logical that single-argument functions should return a type of the same precision.  I'm not sure how this is going to work at all with functions with multiple arguments, and I don't know right now what we should recommend when you're given a literal, since they don't have an assigned width.  This is also part of my not-yet-ready idea for how literals might work without having to attach a bunch of Hungarian-notation suffixes indicating their width (though I'm not entirely opposed to that).

We'll also want to specify what happens when you feed an invalid type to the function.  Some languages return NaN - if we're exactly following IEEE 754 semantics, even _runtime_ would need to support that as well - but at compile-time I think we should decree that producing a NaN is a compilation error.

jakelishman:
Also, there are some bitwise functions that are missing as well:
- `popcount` (Hamming weight) - this seems pretty intuitive to me, provided it only operates on bit-likes (and, _possibly_, unsigned integers)
- `rotl` and `rotr` (bitwise rotations) - these are something we'll need to be careful with in regards to literals.  If we allow unsigneds to be interpreted in bit functions elsewhere, then we're going to have consistency problems because integer literals may not have a bit width attached to them.

jwoehr:
@jakelishman all good points and like having an avalanche fall on my head :)

Will have to look at this at length.

jwoehr:
Working on this ...

jwoehr:
@jakelishman w/r/t `mod` there is always the issue of which way the modulus goes when one argument is negative.

jwoehr:
Some of it still needs work, awaiting more clarifying comments from @jakelishman

source/language/types.rst
```diff
@@ -279,25 +279,53 @@ their argument expression in parentheses:
...
+      +==========+================================+=====================+
+      | arccos   | [-1, 1]                        | [0, pi rad]         |
+      +----------+--------------------------------+---------------------+
+      | arcsin   | [-1, 1]                        | [-pi/2, pi/2 rad]   |
+      +----------+--------------------------------+---------------------+
+      | arctan   | [-∞, Real, ∞]                  | [-pi/2, pi/2 rad]   |
+      +----------+--------------------------------+---------------------+
+      | ceiling  | [float], [float]               | [float]             |
+      +----------+--------------------------------+---------------------+
+      | cos      | [float]                        | [float]             |
```
jakelishman, line 293: Some of this is slightly confusing to me - sometimes we use ranges, and sometimes we use types.  Would it be appropriate to give the types for everything, and then give further limitations on ranges where applicable?  What do we think a sensible response would be when the function is called with something out-of-range?  Should we leave it implementation-defined (hardware interrupt?) or we could specify that (for `float` at least) the functions return `NaN`?  Being low-level, we don't have an exception mechanism, nor any real error handling.

jakelishman, line 293: Is there a reason for the type of `cos` being `float`, but of `tan` being `angle`?  I'd generally expect them both to accept both and return `float`.

source/language/types.rst
```diff
@@ -279,25 +279,53 @@ their argument expression in parentheses:
...
+      +----------+--------------------------------+---------------------+
+      | floor    | [float], [float]               | [float]             |
+      +----------+--------------------------------+---------------------+
+      | log *(e)*| [float]                        | [float]             |
+      +----------+--------------------------------+---------------------+
+      |  mod     | [int], [int]                   | [int]               |
+      |          |                                |                     |
+      |          | [float], [int,float]           | [float]             |
+      +----------+--------------------------------+---------------------+
+      | popcount | [bits, unsigned]               | [positive int]      |
```
jakelishman, line 309: if it outputs a positive int, perhaps it just ought to output a `uint`?

source/language/types.rst
```diff
@@ -279,25 +279,53 @@ their argument expression in parentheses:
...
+      +----------+--------------------------------+---------------------+
+      | pow      | [int], [unsigned int]          | [int]               |
+      |          |                                |                     |
+      |          | [float], [int,float]           | [float]             |
+      |          |                                |                     |
+      |          | [complex], [int,float,complex] | [complex]           |
+      +----------+--------------------------------+---------------------+
+      | rotl     | ??                             | ??                  |
+      +----------+--------------------------------+---------------------+
+      | rotr     | ??                             | ??                  |
```
jakelishman, line 319: These presumably take in `bit`-likes (`bit` and potentially `uint` depending on how the type system shakes out) and theoretically they could take an `int` or `uint`.  I'd probably expect as a user that `rotl(a, b) == rotr(a, -b)`, and I can't immediately think of any problems with that.

source/language/types.rst
```diff
@@ -279,25 +279,53 @@ their argument expression in parentheses:
...
+      |          |                                |                     |
+      |          | [float], [int,float]           | [float]             |
+      +----------+--------------------------------+---------------------+
+      | popcount | [bits, unsigned]               | [positive int]      |
+      +----------+--------------------------------+---------------------+
+      | pow      | [int], [unsigned int]          | [int]               |
+      |          |                                |                     |
+      |          | [float], [int,float]           | [float]             |
+      |          |                                |                     |
+      |          | [complex], [int,float,complex] | [complex]           |
```
jakelishman, line 315: This is another one we'll need to specify the principal root for, but this table probably isn't the right place to do it.

jakelishman (commented):
Sorry I was absent for quite so long - I let myself build up such a backlog from issues on Terra and elsewhere, I'd barely looked at QASM for a month.  I've left some more comments, but I've slightly forgotten if there were any more questions you wanted answers from me about.

At any rate, I'll be gone til the new year now, so happy Christmas/new year!

source/language/types.rst
jwoehr: Sounds reasonable. I'm moving my household and data setup between the holidays so I am counting on being back in action in the New Year in time for the Pragma WG meeting but not sure before.  I'll look at all these and do what I can when I can. Thanks for your attention to this PR, @jakelishman and happy holidays!

jwoehr: Makes sense to me, which isn't of itself definitive, but helps me edit :)

source/language/types.rst
jwoehr: Probably true

source/language/types.rst
jwoehr: Thanks

source/language/types.rst
jwoehr: Not sure I understand your comment.

jakelishman: `pow` can take a fractional power when the right-hand value is a float, so it's subject to the same concerns as `sqrt` about defining which value is returned.  For example, `pow(x, 0.5)` is (mathematically) equivalent to `sqrt(x)`, so we should specify the value returned.  Roughly, in the situation that there's several possible solutions, we return the one which has the largest positive real part, and break ties by picking the one with the positive imaginary part.  I need to think a little more about how that works with regards to complex powers, but I think the definition extends fairly easily - it's the same that I imagine all programming languages use.

source/language/types.rst
jwoehr: Done.

source/language/types.rst
jwoehr: Done

source/language/types.rst
jwoehr: Done.

source/language/types.rst
jwoehr: The missing ranges and missing types will have to be specified. Maybe we can edit this together online sometime.

source/language/types.rst
jwoehr: We also need to spec `div`. I left that an open question.

source/language/types.rst
```diff
@@ -268,6 +268,72 @@ are all of type ``float[64]``.
...
+   .. table:: Built-in mathematical functions in OpenQASM3.
+
+      +----------+--------------------------------+---------------------+
+      | Function | [Input Range/Type] , [...]     | [Output Range/Type] |
+      +==========+================================+=====================+
+      | arccos   | [-1, 1]                        | [0, pi rad]         |
+      +----------+--------------------------------+---------------------+
+      | arcsin   | [-1, 1]                        | [-pi/2, pi/2 rad]   |
+      +----------+--------------------------------+---------------------+
+      | arctan   | [-∞, Real, ∞]                  | [-pi/2, pi/2 rad]   |
```
jakelishman, line 290: We can just put the type of `arctan` as `float` for consistency with the others, and it doesn't need a range because it accepts everything.

source/language/types.rst
```diff
@@ -268,6 +268,72 @@ are all of type ``float[64]``.
...
+   :name: tab:built-in-math
+
+   .. table:: Built-in mathematical functions in OpenQASM3.
+
+      +----------+--------------------------------+---------------------+
+      | Function | [Input Range/Type] , [...]     | [Output Range/Type] |
+      +==========+================================+=====================+
+      | arccos   | [-1, 1]                        | [0, pi rad]         |
+      +----------+--------------------------------+---------------------+
+      | arcsin   | [-1, 1]                        | [-pi/2, pi/2 rad]   |
```
jakelishman, line 288: We resolved in a different comment to put the type information in for every function, and specify the ranges where appropriate.  So perhaps these two should be `[float: [-1, 1]]`, or put the ranges in a separate "notes" column?

source/language/types.rst
```diff
@@ -268,6 +268,72 @@ are all of type ``float[64]``.
...
+      +----------+--------------------------------+---------------------+
+      | arctan   | [-∞, Real, ∞]                  | [-pi/2, pi/2 rad]   |
+      +----------+--------------------------------+---------------------+
+      | ceiling  | [float], [float]               | [float]             |
+      +----------+--------------------------------+---------------------+
+      | cos      | [float]                        | [float]             |
+      |          |                                |                     |
+      |          | [angle]                        | [float]             |
+      +----------+--------------------------------+---------------------+
+      | div      | ??                             | ??                  |
```
jakelishman, line 298: Let's just remove this from the list until there's a compelling reason to add it - if we can't even define it (e.g. how does it differ from the `/` operator?), it's probably an indication we don't want it.

source/language/types.rst
```diff
@@ -268,6 +268,72 @@ are all of type ``float[64]``.
...
+      +----------+--------------------------------+---------------------+
+      | exp      | [float]                        | [float]             |
+      |          |                                |                     |
+      |          | [complex]                      | [complex]           |
+      +----------+--------------------------------+---------------------+
+      | floor    | [float], [float]               | [float]             |
+      +----------+--------------------------------+---------------------+
+      | log *(e)*| [float]                        | [float]             |
+      +----------+--------------------------------+---------------------+
+      |  mod     | [int], [int]                   | [int]               |
```
jakelishman, line 308:
```suggestion
      | mod      | [int], [int]                   | [int]               |
```

source/language/types.rst
```diff
@@ -268,6 +268,72 @@ are all of type ``float[64]``.
...
+      +----------+--------------------------------+---------------------+
+      | div      | ??                             | ??                  |
+      +----------+--------------------------------+---------------------+
+      | exp      | [float]                        | [float]             |
+      |          |                                |                     |
+      |          | [complex]                      | [complex]           |
+      +----------+--------------------------------+---------------------+
+      | floor    | [float], [float]               | [float]             |
+      +----------+--------------------------------+---------------------+
+      | log *(e)*| [float]                        | [float]             |
```
jakelishman, line 306: We might want a "notes" field for the "base-e" comment so it doesn't look like part of the function.  Come to think of it, we could put all the domain/range constraints in a notes field as well.

source/language/types.rst
```diff
@@ -268,6 +268,72 @@ are all of type ``float[64]``.
...
+      +----------+--------------------------------+---------------------+
+      |  mod     | [int], [int]                   | [int]               |
+      |          |                                |                     |
+      |          | [float], [int,float]           | [float]             |
+      +----------+--------------------------------+---------------------+
+      | popcount | [bits, unsigned]               | [unsigned int]      |
```
jakelishman, line 312: I think we ought to use `uint` instead of `unsigned` or `unsigned int` to match OQ3 keywords.

source/language/types.rst
```diff
@@ -268,6 +268,72 @@ are all of type ``float[64]``.
...
+      +----------+--------------------------------+---------------------+
+      | rotr     | [bit], [int,uint]              | [int,uint]          |
+      +----------+--------------------------------+---------------------+
+      | sin      | [float]                        | [float]             |
+      |          |                                |                     |
+      |          | [angle]                        | [float]             |
+      +----------+--------------------------------+---------------------+
+      | sqrt     | [float] *principal root*       | [float]             |
+      |          |                                |                     |
+      |          | [complex] *principal root*     | [complex]           |
```
jakelishman, line 330: A "notes" column would be a good place to put the "principal root" comments in as well.

jakelishman (commented):
Sorry this really took so long.  I'm happy with everything here, bar the minor typographical stuff I mentioned.  The major change I had was to suggest that we just remove `div` entirely - we don't have a meaning for it that differs from the `/` operator, so it's probably best just to get rid of it.  We can always add it back later if it becomes useful.

source/language/gates.rst
```diff
@@ -341,7 +341,7 @@ respectively.
    qubit[5] a;
    qubit f;
    reset f;
-   negctrl(2) @ ctrl(3) @ x a, f;
+   negctrl(2) @ ctrl(3) @ x a[0], a[1], a[2], a[3], a[4], f;
```
jakelishman, line 344: A minor point, but #325 will fix this, so best to keep it out of this PR to avoid conflicts.

jakelishman:
@jwoehr: I made jwoehr/openqasm#1 which should implement everything I commented on (and fixes a problem with `ceiling` and `floor` that I missed).  If you're happy with that, we can merge that and merge this PR.

jakelishman (approved):

PR #301, opened 2021-11-03
Add ImagLiteral node in the reference AST

shiyunon:
### Summary
1. Added ImagLiteral node in the reference AST.
2. Change the typing annotation of `QuantumMeasurementAssignment` target to be Optional due to the fact that the current grammar parses measurements with no targets to `QuantumMeasurementAssignment`.
3. Reformatted source/openqasm/ast.py source/openqasm/tests/test_qasm_parser.py using black.

### Tests done
Tested the new changes in source/openqasm/tests/test_qasm_parser.py.

### DIFF
--- a/source/openqasm/ast.py
+++ b/source/openqasm/ast.py
@@ -194,9 +194,7 @@ class UnaryExpression(Expression):
     expression: Expression
 
 
-BinaryOperator = Enum(
-    "BinaryOperator", "> < >= <= == != && || | ^ & << >> + - * / % **"
-)
+BinaryOperator = Enum("BinaryOperator", "> < >= <= == != && || | ^ & << >> + - * / % **")
 
 
 @dataclass
@@ -1012,9 +1010,7 @@ class DurationOf(QASMNode):
     target: Union[Identifier, List[QuantumStatement]]
 
 
-AssignmentOperator = Enum(
-    "AssignmentOperator", "= += -= *= /= &= |= ~= ^= <<= >>= %= **="
-)
+AssignmentOperator = Enum("AssignmentOperator", "= += -= *= /= &= |= ~= ^= <<= >>= %= **=")
 
 
 @dataclass
--- a/source/openqasm/ast.py
+++ b/source/openqasm/ast.py
@@ -194,7 +194,9 @@ class UnaryExpression(Expression):
     expression: Expression
 
 
-BinaryOperator = Enum("BinaryOperator", "> < >= <= == != && || | ^ & << >> + - * / % **")
+BinaryOperator = Enum(
+    "BinaryOperator", "> < >= <= == != && || | ^ & << >> + - * / % **"
+)
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


jakelishman:
I think my problem here is that the spec doesn't have "imaginary numbers" in the type system.  There are real numbers and complex numbers, but no imaginary-only numbers.  With the current text, I don't think we should be exposing a new type in this manner in the AST.

To be fair, the current form of the spec with regard to complex numbers is not great.  The exact text is
> `complex[type[size]]` types are initalized as `a + b im`, where `a` and `b` must be of the same type as `type[size]`. `b` must occur to the left of `im` and the two can only be seperated by spaces/tabs (or nothing at all).

Taken literally this implies that `z = 3 im;` is invalid OpenQASM 3 because it doesn't have the `a + ` bit in it.  Even setting that aside (a later example has `z = -2 - 7im;`, which also would be invalid), the AST representation of `3 im` should likely be a complex number, not an imaginary one, if we were supporting complex-number literals.

The spec currently only really allows the `im` keyword in initialisers - the form of the initialiser doesn't really make sense to be used in any other situation because it can take expressions as `a` and `b`: the spec lists the following example:
```
complex[float[64]] d = 2.0+sin(π) + 3.1*5.5 im;
```
which it presumably intends to be equivalent to `2.0 + 17.05im`. It doesn't define how tightly the `+` symbol necessary for the initialiser should bind, so the `sin(pi)` could be attached to either the real or imaginary part right now. Having this adds an extra rule into the expression tree: as well as something like `AddExpression: Expression "+" Expression` you also need `ComplexNumber: Expression "+" Expression "im"` or something to that effect, and then you start getting super weird stuff like `2.0 + 3.0im * 2` could be parsed as `MultiplyExpression(ComplexNumber(2.0, 3.0), 2)`, or `2 * (2.0 + 3.0im)`, and the order of operations (`+` before `*`??) has gone to hell.  

I'm also marginally concerned that there may be a valid OpenQASM 3 statement that requires an arbitrary number of tokens of lookahead to parse here, but I'm not certain at all.  ANTLR can handle that (slowly), but others who have more performant parsers (e.g. a Bison-based LALR(1)) might not be able to support it, for no real benefit.

If anything, with the current spec, I think the AST node for this type of expression should be along the lines of
```
class ComplexInitialiserValue(Expression):
    real: Expression
    imag: Expression
```
but honestly I think we'd be best completely revisiting complex numbers in the types group before.  I'm actually in favour of having an imaginary literal lexeme similar to that implied by the AST node in this PR, if we can decide on a rock solid set of allowed implicit casting rules - I'm not certain if we will manage that, though.

shiyunon:
Thanks for the comment @jakelishman . I agree with you that "we'd be best completely revisiting complex numbers in the types group before". What I did in this PR is just to map what's in the grammar to what's in the AST. If the grammar is problematic, it's understandable to not have that mapping for the problematic part. I will make this PR a draft and revisit it after we have a decision in the type group.

1ucian0:
This PR is `on hold` as it is blocked by the discussion in https://github.com/Qiskit/openqasm/issues/302

jakelishman:
Since #337 will weaken the type system by adding implicit conversions, a lot of the original issues with the complex type should disappear.  We (hopefully) won't need a "complex literal" token any more, only the imaginary literal added by this PR.  We've just started discussing this in the types group, but it looks like this will be the way forwards.

jwoehr:
Too far behind main, no longer relevant. Please re-open if needed.

PR #315, opened 2021-12-10
Fix #313. Solves some symbol clashes in the spec.

### LINKED ISSUE #313: Type and scope declaration conflicts in stdgates.inc
In the current stdgates.inc OQ3 header file we find the following two gate declarations:

```
// sqrt(S) or T gate
gate t a { pow(1/2) @ s a; }

// more gate declarations here

// controlled-NOT
gate cx c, t { ctrl @ x c, t; }
```

I believe that the declaration for the ```cx``` gate contains a conflict with the declaration of the ```T``` gate: the second qubit parameter to the ```cx``` gate declaration has the same name [```t```] as the ```T``` gate.

Since gates have global scope, this is an implicit re-declaration conflict (qubit parameters are not explicitly declared, they are implicitly declared and have gate scope).

There are several other similar conflicts currently present in stdgates.inc.

I think these conflicts should be cured.

boschmitt:
### Summary
Many places in the specification use variables/parameters/arguments
names that would inevitably clash with names used in gates declarations
in the standard gate library (`stdgates.inc`). This PR tries to solves most 
(but unlikely all) of them.

### Details and comments
Someone more familiar with the `openpulse` part of the specification
needs to go through it to fix potential problems. (I not that familiar with it.)

The discussion in #313 raises points that are likely to span other issues
and deserve more discussion.

### DIFF
--- a/examples/dd.qasm
+++ b/examples/dd.qasm
@@ -5,10 +5,10 @@
 OPENQASM 3.0;
 include "stdgates.inc";
 
-stretch s;
-duration start_stretch = -0.5 * durationof({x $0;}) + s;
-duration middle_stretch = -0.5 * durationof({x $0;}) - 5 * durationof({y $0;}) + s;
-duration end_stretch = -0.5 * durationof({y $0;}) + s;
+stretch a;
+duration start_stretch = -0.5 * durationof({x $0;}) + a;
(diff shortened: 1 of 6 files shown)


blakejohnson (approved):

jakelishman (changes requested):
One minor typo, otherwise looks good.

blakejohnson (approved):

PR #325, opened 2022-02-17
Clarify description of the `ctrl` and `negctrl` modifiers

### LINKED ISSUE #320: Question about n of control  modiﬁer
Hi. I have a small question regarding understanding the control modifier
```
"qubit[5] a;\n"
"qubit[5] b;\n"
"qubit f;\n"
"ctrl(3) @ x a, b[1], b[2] f;"
```
What control bits should be defined for the case above?
`a[0], a[1], a[2]` or `a[0], a[1], a[2], a[3], a[4], b[1], b[2]`

Thanks.

vasilenkoalexey:
### Summary

Fix for #320. Examples refactor and extend description of the control modifier.

### Details and comments

1. From my point of view `control qubit` should be replaced by `control argument` that could be single bit\qubit or bit\qubit register.
2. Added reference to examples of broadcast rule if control argument is a register
3. `control argument n` replaced by `control parameter n`. Here we already describe the control arguments (which prepended before gate arguments) and call `n` as the `argument` again. This may cause some confusion. Also, I think `n` as `parameter` is more appropriate to model `gate name(params) qargs`. Then gate with `ctrl` modifier might look like this `ctrl(cparam) @ gate name(params) cargs qargs`, which seems logical enough.
4. Added requirement that the count of control arguments be equal to the sum of `n` values.
5. The current examples basically repeats each other. I have replaced them with more general examples that cover as many different cases as possible.

All these changes explain some misunderstandings that have caused questions in #320

### DIFF
--- a/source/language/gates.rst
+++ b/source/language/gates.rst
@@ -289,9 +289,11 @@ The modifier ``ctrl @`` replaces its gate argument :math:`U` by a
 controlled-:math:`U` gate. If the control bit is 0, nothing happens to the target bit.
 If the control bit is 1, :math:`U` acts on the target bit. Mathematically, the controlled-:math:`U`
 gate is defined as :math:`C_U = I \otimes U^c`, where :math:`c` is the integer value of the control
-bit and :math:`C_U` is the controlled-:math:`U` gate. The new control qubit is prepended to the
-argument list for the controlled-:math:`U` gate. The modified gate does not use any additional
-scratch space and may require compilation to be executed.
+bit and :math:`C_U` is the controlled-:math:`U` gate. The control modifier argument can be a
+register, and in this case controlled gate broadcast over it (as it shown in examples above for
+CNOT gate). The new control argument is prepended to the argument list for the
+controlled-:math:`U` gate. The modified gate does not use any additional scratch space and may
+require compilation to be executed.
 
 We define a special case, the controlled *global* phase gate, as
 :math:`ctrl @ gphase(a) = U(0, 0, a)`. This is a single qubit gate.
@@ -319,9 +321,11 @@ and :math:`N_U` is the negative controlled-:math:`U` gate.
        negctrl @ x q1, q2;
    }
 
-``ctrl`` and ``negctrl`` both accept an optional positive integer argument ``n``, specifying the
-number of control bits (omission means ``n=1``). ``n`` must be a compile-time constant. For an ``N``
-qubit operation, these operations are mathematically defined as
+``ctrl`` and ``negctrl`` both accept an optional positive integer parameter ``n``, specifying the
+number of control arguments (omission means ``n=1``). The sum of all control modifier parameters
+must be equal to the count of all control arguments, otherwise the statement is an invalid.
+``n`` must be a compile-time constant. For an ``N`` qubit operation,these operations are
+mathematically defined as
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


source/language/gates.rst
```diff
@@ -319,9 +321,11 @@ and :math:`N_U` is the negative controlled-:math:`U` gate.
        negctrl @ x q1, q2;
    }
 
-``ctrl`` and ``negctrl`` both accept an optional positive integer argument ``n``, specifying the
-number of control bits (omission means ``n=1``). ``n`` must be a compile-time constant. For an ``N``
-qubit operation, these operations are mathematically defined as
+``ctrl`` and ``negctrl`` both accept an optional positive integer parameter ``n``, specifying the
+number of control arguments (omission means ``n=1``). The sum of all control modifier parameters
+must be equal to the count of all control arguments, otherwise the statement is an invalid.
```
jakelishman, line 326: I think the inserted sentence here might be a bit more complicated than just not having it - with the bad examples corrected, I think the discussion above of "`ctrl(n)` adds `n` arguments to the list" is sufficient to explain this.  Introducing new terminology like "sum of control modifier parameters" etc makes it seem to me like it's more complicated than it really is (like it makes me double-take over whether `ctrl` behaves differently to `ctrl(1)` because the former doesn't have a "modifier parameter").

source/language/gates.rst
```diff
@@ -338,15 +342,12 @@ respectively.
...
-   negctrl @ ctrl(3) @ x a[0], a[1], a[3], a[4], f;
-   negctrl @ ctrl(3) @ x a[1], a[0], a[3], a[4], f;
-   ctrl(3) @ x a[0], a[1], a[2], f;
-   negctrl(3) @ ctrl @ x a[0], a[1], a[2], a[3], f;
+  qubit[3] a;
+  qubit[3] b;
+  ctrl(3) @ x a[1], a[0], a[2], b[2];
+  negctrl (3) @ ctrl @ x a[0], b[1], a[2], b[0], b[2];
+  negctrl @ ctrl(2) @ negctrl @ x a[0], b[0], a[2], a[1], b[2];
+  negctrl (2) @ ctrl @ x b[1], a, b[0], b[2];
```
jakelishman, line 350:
```suggestion
  negctrl(3) @ ctrl @ x a[0], b[1], a[2], b[0], b[2];
  negctrl @ ctrl(2) @ negctrl @ x a[0], b[0], a[2], a[1], b[2];
  negctrl(2) @ ctrl @ x b[1], a, b[0], b[2];
```

(just for style consistency)

source/language/gates.rst
```diff
@@ -289,9 +289,11 @@ The modifier ``ctrl @`` replaces its gate argument :math:`U` by a
 controlled-:math:`U` gate. If the control bit is 0, nothing happens to the target bit.
 If the control bit is 1, :math:`U` acts on the target bit. Mathematically, the controlled-:math:`U`
 gate is defined as :math:`C_U = I \otimes U^c`, where :math:`c` is the integer value of the control
-bit and :math:`C_U` is the controlled-:math:`U` gate. The new control qubit is prepended to the
-argument list for the controlled-:math:`U` gate. The modified gate does not use any additional
-scratch space and may require compilation to be executed.
+bit and :math:`C_U` is the controlled-:math:`U` gate. The control modifier argument can be a
+register, and in this case controlled gate broadcast over it (as it shown in examples above for
+CNOT gate). The new control argument is prepended to the argument list for the
```
jakelishman, line 294: I think the previous terminology of "new control qubit" was slightly clearer, because it doesn't clash with the value of `n` - you mentioned it in your main post, but it's quite a subtle distinction that I think will just be a bit harder to understand for the reader.  We could say "quantum argument" if you prefer not to say "qubit" (since it can also be a collection of qubits).

I'd put the bit about "the new control quantum argument is prepended to the argument list" back in its original place, and put your new sentence about "The new arguments may also be quantum registers, in which case the standard broadcasting rules apply" after - we want to put the most important, common circumstances first, and handle the special cases later.

source/language/gates.rst
jakelishman, line 350: The indentation here has got broken - I think a space got removed.  The new set of examples looks fine, though I'd keep the non-register `qubit f` in the mix just for completeness.

jakelishman (commented):
Thanks for looking at this - it's good to get the examples fixed.  I had a couple of places where I think the new terminology might be making things a little more confusing, but in principle I think we'll be happy to accept the new sentence about broadcasting.

source/language/gates.rst
vasilenkoalexey: Ok, I deleted this phrase

source/language/gates.rst
vasilenkoalexey: Done

source/language/gates.rst
vasilenkoalexey: Fixed

PR #326, opened 2022-03-09
Details for implicit casting and literals

mbhealy:
### Summary

Add details for implicit casting and literals.

### Details and comments

Generally, implicit casting is not allowed, specify the few situations in which implicit casts are allowed (const float expressions to angle on assignment and as arguments to calls).
Also:
- Added pi_a as a reserved keyword for an angle type version of the constant pi
- Specified that mixed-type expressions are disallowed except when a mixed-type operator is specified.
- Described array reshaping (similar to C++ reinterpret cast)

This change would break backwards compatibility with OQ2 because the expression `3*pi/4` interprets the integer literals 3 and 4 as floats. After this PR expressions of that type would have to be of the form `3.0*pi/4.0`. We did not find a good solution to allowing that expression to be implicitly promoted to float without making the rules very complex for when promotion does and does not occur, and we generally dislike the idea of allowing integer to float promotion in general given the bit specificity of the OpenQASM 3 language.

### DIFF
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -103,12 +103,12 @@ multiplication, division, and power and the corresponding assignment operators.
 
 .. code-block:: c
 
-   angle[20] a = pi / 2;
+   angle[20] a = pi / 2.0;
    angle[20] b = pi;
    a + b; // 3/2 * pi
    a ** b; // 4.1316...
    angle[10] c;
-   c = angle(a + b); // cast to angle[10]
+   c = angle[10](a + b); // cast to angle[10]
 
 Complex numbers
 ~~~~~~~~~~~~~~~
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -126,6 +126,15 @@ assignment operators.
    complex[float[64]] f = a / b; // f = (-55.0+60.0im)/53.0
    complex[float[64]] g = a ** b; // g = (0.10694695640729072+0.17536481119721312im)
 
+Mixed-type expressions
+~~~~~~~~~~~~~~~~~~~~~~
+
+In general, mixed-type expressions are not allowed. All expression terms must
+have the same type or be explicitly cast to have the same type. Literal types
+must also match. For example, the expression ``3 * pi / 4`` is not legal,
+because it mixes a const float (``pi``) with integer literals, the correct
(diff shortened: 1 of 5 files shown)


mbhealy:
Still need to add grammar updates and operators for dividing and multiplying angles by floats.

steleman:
> Added pi_a as a reserved keyword for an angle type version of the constant pi

This would imply that there will also be a `tau_a` and `euler_a` to represent the corresponding angles?

Wouldn't it be more intuitive to imitate the naming convention of the C Standard Math Library?

The C constant naming convention starts with `M_`.

- `m_pi`, `m_tau`, `m_euler` for the numeric constants?
- `a_pi`, `a_tau`, and `a_euler` for the corresponding angles?

steleman:
1. What happens when the programmer wants to use Unicode to identify these reserved keywords?

As in: `π`, `τ`, `ε`.

Do these now become `α_π`, `α_τ`, `α_ε` and/or `μ_π`, `μ_τ`, `μ_ε`?

2. What happens when the programmer wants to write a gate definition that uses implicit angle identifiers? As in:

```
gate example(λ, θ, φ) q, r {
  // ...
}
```

mbhealy:
> 1. What happens when the programmer wants to use Unicode to identify these reserved keywords?
> 
> As in: `π`, `τ`, `ε`.
> 
> Do these now become `α_π`, `α_τ`, `α_ε` and/or `μ_π`, `μ_τ`, `μ_ε`?
> 
> 2. What happens when the programmer wants to write a gate definition that uses implicit angle identifiers? As in:
> 
> ```
> gate example(λ, θ, φ) q, r {
>   // ...
> }
> ```

The behavior of normal identifiers (including implicitly typed ones) is unchanged. The `_a` is only special for literals and the built-in angle constant pi_a.

steleman:
My theory is that we don't need any of this to accomplish the desired result, which I would attempt to formally describe as: _type-discriminated expression evaluation overloading_.

That is a complicated way of saying: casting an arithmetic expression to an Angle Type changes the semantics of the expression operand and any sub-expressions thereof.

mbhealy:
We have changed direction based on feedback here and in the TSC meeting. Closing this in favor of #337

Issue #294, opened 2021-10-15
What are the use cases for pragmas


1ucian0:
During the TSC meeting on Oct 15 it was decided that we are going to collect some draft use cases for `#pragmas` as a first step to work on syntax and functionality. Please, add you use case as a comment (do not worry too much about the syntax).

jheckey:
@steleman I readily concede the point. Most of these suggestions I'm agnostic to, but I think they solve various problems (some of which may not exist) that customers may be interested in. I just tend to be loud and opinionated, so I want see what the smarter people here think.

What about the simulations? Does it make sense to keep those as pragmas that can be passed to simulators and ignored when the circuit is executed? The examples could be stronger, and arguably the most interesting simulations would be system level where modelling subsystems' interactions to evaluate the execution of the circuit are more interesting than the circuit itself could be better modeled in other languages like SystemC or PySim.

@jwoehr I like the suggestion of dropping the `#`.

steleman:
> What about the simulations?

I thought these were actually the only ones that had meaningful pragma semantics.

But without the scope resolution operator `::` because the scope resolution operator really has no semantics here.  It just looks like cosmetics.

```
pragma sim advance-time(1us)
```

```
pragma sim measurement(expectation)
```

This tracks pretty closely with the pragma syntax as it is implemented in other language, where the general syntax form is:

```
TOK_PRAGMA <Pragma-Command> [<Pragma-Subcommand>] ['(' <Pragma-Operand> [, <Pragma-Operand> ...]')'] 
```

Obviously, not all pragmas have a sub-command, and not all pragmas have operands. The square brackets above mean "optionally" in this context, they aren't meant to be parsing lexemes.

NoahGWood:
Well a big use case is related to issue #230 which could be resolved by simply declaring the proper gates for each machine ala:
```
#if SomeArchitecture
gate h a { U(π/2, 0, π) a; }
#elif OtherArchitecture
gate h a { U(π/2, π, 0) a; }
...
#endif
```

This would enable you to create cross-platform code and libraries making it much easier to port a given algorithm across different architectures and paradigms while also allowing you finer grained control on any given architecture.

Another example would be code optimization, to generate a random number on IBM's QExperience you might use:
```
include stdgates.inc;
qubit a;
reset a;
h a;
x a;
measure a;
```
[...]

PR #333, opened 2022-03-22
Clarify and provide examples for Pragmas and Annotations

jheckey:
### Summary

Following up with textual changes for [#294](https://github.com/Qiskit/openqasm/issues/294). The initial PR is for discussion with the TSC; required grammar updates will be included once the textual changes are approved.

### Details and comments

This PR updates the [Directive](https://qiskit.github.io/openqasm/language/directives.html#directives) section to be more explicit about what pragmas are and how they should be used, including examples. A new directive type (Annotations) is also proposed to imply binding pragma-like behavior to a scoped area of code in a concise way.

#### Topics discussed and alternatives

Pre-processor macros (e.g., `#if`, `#endif`, `#define`) were discussed and rejected based on a desire to not add a pre-processor into OpenQASM implementions. This is largely to avoid issues with adding and specifying a pre-processor in the OpenQASM specification. We justified this by thinking of OpenQASM as an IR, not a human-written language; the pre-processing should be done in the native language that the human is writing and the OpenQASM should be
generated to match.

Scoping and precedence for pragmas was discussed and ultimately rejected as overly prescriptive. We offered examples and recommendations about how to use pragmas safely as global settings, and to avoid implementing stateful behavior,
but ultimately left the decision to the implementation. Annotations were added as a simplified mechanism for associating pragma-like behavior that was local to a piece of code.

Removing the leading hash symbol on `#pragma` was chosen for consistency with `include` not having it, and focusing on pragmas being native keywords in OpenQASM. It is a vestigial symbol from C/C++ and OpenQASM 2.0, but it is
redundant with a keyword (`#pragma ifdef foo` is awkward syntax). Since we are focused on making `pragma` a fully parsed member of OpenQASM, and no other keywords are prefixed with a symbol, it didn't make sense to keep it. Another
acceptable alternative would be to allow just `#` followed by text.

The decision to use `#pragma` in favor of `pragma` is undercut by the use of `@` as the marker for annotations (borrowed from Python and Java). The keyword `tag` is a reasonable alternative that is more inline with the previous decision.

Multi-line pragmas were discussed (see example below) and rejected as complicated to handle for naive OpenQASM scripts that could simply omit a single line. This does come at the cost of readability, but fits with the principle of
OpenQASM being generated instead of human written. 

    #pragma
        noise matrix [
            [1], [0],
            [0], [1im]
        ]
    #endpragma

### DIFF
--- a/source/language/directives.rst
+++ b/source/language/directives.rst
@@ -1,13 +1,85 @@
 Directives
 ==========
 
-OpenQASM supports a directive mechanism that allows other information to
-be included in the program. A directive begins with ``#pragma`` and
-terminates at the end of the line. Directives can provide annotations
-that give additional information to compiler passes and the target
-system or simulator. Ideally the meaning of the program does not change
+OpenQASM supports directive mechanisms that allows other information to
+be included in the program. Directives are either pragmas or annotations.
+Both are used to supply additional information to the compiler passes and the
+target system or simulator. Ideally the meaning of the program does not change
 if some or all of the directives are ignored, so they can be interpreted
 at the discretion of the consuming process.
+
+Pragmas
+-------
+
+Pragma directives start with ``pragma`` and continue to the end of line. The
+``#pragma`` syntax is also accepted for reverse compatibility. Pragmas should be
+processed as soon as they are encountered; if a pragma is not supported by a
+compiler pass it should be ignored and preserved intact for future passes.
+Ideally, pragmas should be considered global and immutable; avoiding stateful
+interactions is recommended to prevent unexpected behaviors between source
+files.
+
+Pragmas are useful for extending OpenQASM functionality that is not described in
+this specification, such as adding noise operations.
+
+\*Note: The following examples are simply possible implementations, this
+specification does not define any pragmas. Please consult your tool's
+documentation for supported pragmas.
+
+.. code-block:: c
+   :force:
+
+   rx(0.15) q;
+   pragma noise bit_flip(0.1)
+
+Pragmas can also be used to specify system-level information or assertions for
+the entire circuit.
+
+.. code-block:: c
+   :force:
+
+   OPENQASM 3.0;
+
+   // Attach billing information
+   pragma user alice account 12345678
+
+   // Assert that the QPU is healthy to run this circuit
+   pragma max_temp qpu 0.4
+
+   qubit[2] q;
+
+
+Annotations
+-----------
+
+Annotations can be added to supply additional information to the following
+OpenQASM statement. These annotations will apply to the subsequent declaration,
+box, or single line of code. Annotations will start with a ``@`` symbol, have an
+identifying keyword and continue to the end of the line.
+
+\*Note: The following examples are simply possible implementations, this
+specification does not define any annotations. Please consult your tool's
+documentation for supported annotations.
+
+.. code-block:: c
+   :force:
+
+   // Manage port binding on a physical device
+   @bind IOPORT[3:2]
+   input bit[2] control_flags;
+
+   // Instruct compiler to create a reversible version of the function
+   @reversible
+   gate multiply a, b, x {
+      x = a * b;
+   }
+
+   // Prevent swap insertion
+   @noswap
+   box {
+      rx(pi) q[0];
+      cnot q[0], q[1];
+   }
 
 
 Input/output
--- a/source/language/directives.rst
+++ b/source/language/directives.rst
@@ -13,15 +13,16 @@ Pragmas
 
 Pragma directives start with ``pragma`` and continue to the end of line. The
 text after ``pragma`` is a single string, and parsing is left to the specific
-implementation. The ``#pragma`` syntax is deprecated, but is accepted for
-compatibility. Pragmas should be processed as soon as they are encountered; if a
+implementation. Implementations may optionally choose to support the older ``#pragma``
+keyword as a custom extension.
+Pragmas should be processed as soon as they are encountered; if a
 pragma is not supported by a compiler pass it should be ignored and preserved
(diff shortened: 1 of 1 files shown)


source/language/directives.rst
```diff
@@ -1,14 +1,95 @@
...
 if some or all of the directives are ignored, so they can be interpreted
 at the discretion of the consuming process.
 
+Pragmas
+-------
+
+Pragma directives start with ``pragma`` and continue to the end of line. The
+text after ``pragma`` is a single string, and parsing is left to the specific
+implementation. The ``#pragma`` syntax is deprecated, but is accepted for
+compatibility. Pragmas should be processed as soon as they are encountered; if a
```
taalexander, line 17: Should we maintain backward compatibility or remove this and treat it as a custom vendor extension if a specific implementation cannot pivot to the updated syntax immediately?

source/language/directives.rst
```diff
@@ -1,14 +1,95 @@
...
+pragma is not supported by a compiler pass it should be ignored and preserved
+intact for future passes.  Ideally, pragmas should be considered global and
+immutable; avoiding stateful interactions is recommended to prevent unexpected
+behaviors between source files.
+
+Pragmas are useful for extending OpenQASM functionality that is not described in
+this specification, such as adding noise operations.
```
taalexander, line 24:
```suggestion
this specification, such as adding noise operations for a program to be executed on a simulator.
```

source/language/directives.rst
```diff
@@ -1,14 +1,95 @@
...
+   // Assert that the QPU is healthy to run this circuit
+   pragma max_temp qpu 0.4
+
+   qubit[2] q;
+
+
+Annotations
+-----------
+
+Annotations can be added to supply additional information to the following
```
taalexander, line 56: I don't think the distinction between pragmas and annotations in clear in this document.

taalexander (changes requested):

source/language/directives.rst
levbishop: I think the latter is simpler. Vendors can always accept an extended language.

source/language/directives.rst
jheckey: The primary difference is scoping - a pragma is global, while (an) annotation(s) (is/are) scoped to the next OpenQASM non-annotation statement as defined by the grammar. There is nothing preventing an implementer from using pragmas instead of annotations, but by specifically binding annotations to code they can at least be manipulated together in a compiler.

source/language/directives.rst
```diff
@@ -1,14 +1,95 @@
...
+
+\*Note: The following examples are simply possible implementations, this
+specification does not define any pragmas. Please consult your tool's
+documentation for supported pragmas.
+
+.. code-block:: c
+   :force:
+
+   rx(0.15) q;
+   pragma noise bit_flip(0.1)
```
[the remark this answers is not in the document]
jheckey: Maybe a more accurate statement should be "stateless"; they may have positional relevance in an implementation, but should not (not "must not") implement behavior that needs to be enabled and disabled. Annotations essentially provide enclosures around any scoped code. The goal is to avoid issues such as included files in large codebases changing previously defined settings unexpectedly (accidental `translate_off` that isn't closed, changing `simulation_timescale` between files, etc.).

source/language/directives.rst
```diff
@@ -1,14 +1,95 @@
...
+
+   qubit[2] q;
+
+
+Annotations
+-----------
+
+Annotations can be added to supply additional information to the following
+OpenQASM statement. These annotations will apply to the subsequent declaration,
+box, or single line of code. Annotations will start with a ``@`` symbol, have an
```
jakelishman, line 58: By "single line of code", do we mean that I can do:
```qasm
@storage fast
int[8] a1 =
@overflow check
100 +
@overflow nocheck
100;
```
(i.e. deliberately splitting the line to apply different annotations to different subexpressions)?

I'd be in favour of annotations being applied to the following "statement" rather than line (maybe this was what was intended already). I can only think of grossly contrived examples for splitting within statements, and implementation could be very tricky if annotations can be attached to arbitrary groups of tokens rather than fixed parser productions.

Applying to the following "statement" is potentially awkward if somebody decides to follow an annotation with several statements on the same line:
```qasm
@annotation
int[8] a = 1; uint[8] b = 2;
```
so it's not necessarily clear cut.  I'd prefer to say that this example only applied the annotation to `int[8] a = 1;` and push the burden of complaining about the misleading formatting onto a linter.

jakelishman (commented):
I can help with writing lexer/parser rules for the reference grammar/AST+parser once this is finalised.  For these ANTLR-based things I'll need #278 to merge first.  I can't think of a neat user-facing AST if annotations can be attached to arbitrary groupings of tokens, rather than statements, but I might just be misinterpreting what's meant by the PR (more detail in the comment).

source/language/directives.rst
jheckey: The intention was "statement" in terms of the grammatical definition. I over-elaborated because I wasn't necessarily certain that a casual reader would understand the word "statement" in context, but it may be clearer to not over explain (and maybe provide a link to a definition for "statement").

jakelishman: Perfect, thanks - that makes more sense to me.

source/language/directives.rst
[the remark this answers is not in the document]
jheckey: This example should be removed - it is better as an annotation.

source/language/directives.rst
jakelishman: This comment chain is completely addressed by da1799d.

taalexander (approved):

PR #339, opened 2022-04-11
Add release notes directives to build

jwoehr:
### Summary

* Added `release_notes.rst` to `source` dir
* Added `'reno.sphinxext'` to `source/conf.py`

### DIFF
--- a/.github/workflows/build-live-version.yml
+++ b/.github/workflows/build-live-version.yml
@@ -12,6 +12,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           persist-credentials: false
+          fetch-depth: 0
 
       - uses: actions/setup-python@v2
 
--- a/.github/workflows/publish-estable-version.yml
+++ b/.github/workflows/publish-estable-version.yml
@@ -21,6 +21,7 @@ jobs:
         with:
           ref: ${{ github.event.inputs.version-tag }}
           path: ${{ github.event.inputs.version-tag }}
+          fetch-depth: 0
 
       - uses: actions/setup-python@v2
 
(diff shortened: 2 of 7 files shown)


jwoehr:
> I think you'll need to make the same fetch depth change in the two publish jobs: https://github.com/Qiskit/openqasm/blob/main/.github/workflows/publish-estable-version.yml and https://github.com/Qiskit/openqasm/blob/main/.github/workflows/publish-live-version.yml they'll only start failing post-merge (or on a release).

Done and tests passed. Thank you.

jakelishman (commented):
We'll need to include the release notes in one of the toctrees somewhere as well, so it gets linked into the documentation.  I think just putting it in `source/index.rst` at the bottom of the main tree should be fine.

jwoehr:
> We'll need to include the release notes in one of the toctrees somewhere as well, so it gets linked into the documentation. I think just putting it in `source/index.rst` at the bottom of the main tree should be fine.

One sphinx question, @jakelishman ... if the release notes directory is `releasenotes` then does the `source/release_notes.rst` file read:
```
.. release-notes:: ../releasenotes
```

I'm not very clued on `sphinx`

jakelishman:
Sphinx is temperamental, but at any rate, the current path appears to be configured correctly - I tried putting in a temporary release note, and it all built correctly.

I've pushed a commit that adds the release notes RST file to the main index, so it'll now show up in the documentation, although for now it's still empty because there aren't any notes.

jwoehr:
Please do a "review" and I'll merge.

jakelishman (approved):
Looks correct to me.  After this merges, we need every reviewer on any PR to this repo to ensure that the correct release notes have been generated in the correct place, but that's not so onerous.

PR #340, opened 2022-04-11
Clarify rules for scoping

mbhealy:
### Summary
Clarify the rules surrounding scoping of variables. 

### Details and comments
This PR is the result of a discussion among the Types and Casting WG. It defines global scope and block scope. Variable scope begins at the point of declaration and ends at the end of the enclosing block. Shadowing of variables is allowed. Global variables are `const` within the scope of subroutines, gates, and defcals.


source/language/scope.rst
```diff
@@ -0,0 +1,61 @@
...
+.. code-block:: c
+
+  int ii = 100; // global scope
+  def myMul() -> int {
+    return ii*2; // ii is in scope, but 'const'
+  }
+  ii = 200;
+  int jj = myMul(); // jj gets the value 400
+
+Block scope
```
psivaraj, line 32: **question**: How do qubits behave in the context of block scoping? Can we shadow, etc...? In any case, I think we should 
(1) amend the statement on lines 18-21 to reflect that fact 
(2) add an example below.

source/language/scope.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Identifiers declared as subroutines, gates, and defcals have global scope, 
+*i.e.* they are visible at every scope of the program, and may not be shadowed.
+Subroutines, gates, and defcals may only be declared at the top-level of the
+program. Redeclaration or overloading of subroutines and gates is not allowed.
+Defcals may be overloaded as described in :any:`pulse-gates`.
+
+Non-quantum variables defined at the top level of the program have global scope,
+and are visible at all lower scopes. Globally-scoped variables are
+implicitly ``const`` inside the body of subroutine and gate definitions, though
```
stevenheidel, line 17: Other than discouraging a bad practice, what other motivations led to the implicit const inside the body of subroutine and gate definitions?

source/language/scope.rst
```diff
@@ -0,0 +1,61 @@
...
+    ii *= 2;    // global ii takes the value 200
+    int ii = 1; // inner ii takes the value 1
+    ii *= 2;    // inner ii takes the value 2
+  } // inner ii scope ends
+  // global ii is 200
+  ii *= 2; // global ii takes the value 400
+  int sum = 0;
+  for ii in [1:10] {
+    // global ii is shadowed by the iteration variable ii
+    ii *= 2;   // error: iteration variable may not be modified by the loop
```
stevenheidel, line 53: Iteration variables are effectively const?

source/language/scope.rst
jakelishman: Mostly the motivation was to limit the ability for subroutines as far as possible, to make that only possible through parameters that are passed as mutable.  Bad practice is a lot of it, with the idea was also to make it clearer to both the compiler and the user about what variables a subroutine may mutate.

With the ability to have "output" variables, it's still possible to mutate global-scope variables, there was just a general consensus within the group that it was better to have them explicitly passed as mutable.

jakelishman: I'll also reword the text about "implicitly `const`", since that's actually quite misleading - `const` in OQ3 is a _compile-time_ constant, whereas in this case we mean the variable may have been assigned at run-time, but it's read-only within the subroutine.

jakelishman: This behaviour was changed in 07a76c8 (subject to review from the working group), since I believe it's more in-line with another PR we have in development about the meaning of the `const` modifier.

jakelishman:
I've expanded on all this text to address the review comments above, and to reflect further discussions about `const` and the for-loop iteration variable in the types-and-casting working group.  The commit message has some more detail:

This adds more text and formalism on what variables are in scope, and
expands on the types of scopes.  In particular, we identify and talk
about the differences between the scoping of subroutine/gate bodies and
the scoping of local blocks with regard to symbol visibility.

Most of this commit is just additional details and examples on top of
the previous commit in the PR.  There are some changes to pre-existing
text:

- Removed mention of the for-loop iteration variable being immutable -
  this behaviour was removed from the PR about for-loop variables.

- Changed the visibility of symbols within `def` and `gate` statements
  such that *only* `const` variables, subroutines/gates and built-ins
  are visible.  The reason to remove making mutable run-time values
  visible is to forbid closing over these variables, which may be more
  difficult for immediate implementations with no current motivating
  use-case.

source/language/scope.rst
```diff
@@ -0,0 +1,318 @@
...
+      // `U` is the single-qubit gate that is implicitly defined in the global
+      // scope of all OpenQASM 3.0 programs, and so is available here.
+      // Similarly `pi` is one of the implicitly defined constants.
+   }
+
+   int i = 100;
+
+   include "my_definitions.qasm";
+
+   // Identifiers 'h', 'my_gate', 'i' and 'SIZE' are defined and in scope.
```
blakejohnson, line 83: Do you mean `j` is in scope after the `include`? What is `SIZE`?

source/language/scope.rst
```diff
@@ -0,0 +1,318 @@
...
+symbols can differ slightly between these.  The types of scope are:
+
+* :ref:`Global scope <scope-global>`, which is the current scope only when no
+  other scopes are active.
+
+* :ref:`Gate and function scope <scope-subroutine>`, which is entered in the
+  body of ``gate`` and ``def`` definitions.
+
+* :ref:`Local block scope <scope-block>`, which becomes active on entry to a
+  non-gate and non-subroutine control-flow block, such as ``if`` or ``for``.
```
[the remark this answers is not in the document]
blakejohnson: I see that we include `box` in the list below.

source/language/scope.rst
```diff
@@ -0,0 +1,318 @@
...
+accessible after the body of the loop.
+
+The blocks associated with ``if`` and its corresponding ``else`` define two
+different scopes; the variables and definitions are not shared between them.
+
+As with subroutine scopes, variables defined locally in these scopes (including
+the for-loop iteration variable) may shadow variables with the same name in
+outer scopes.  When the defining scope of a shadowing variable ends, the
+previous variable (which was shadowed) becomes accessible again.  Qubits and
+arrays cannot be declared within local block scopes, but aliases can.
```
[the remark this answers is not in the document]
jakelishman: This is fair enough from a language-design perspective.  I think it might be good to talk to some implementors about making global variables available in function scopes before we move to do it - the way I've written it, we're a bit more conservative in the language, because we never need to worry about global-state mutation.  This was a feature of Michael's original draft, I just made it stricter, and perhaps a bit more consistent in the type system here by removing their visibility.

I think if global variables are visible within functions, there's probably not as strong an implementation reason to forbid mutating them.  We can discuss it more in the meeting on Thursday?

source/language/scope.rst
jakelishman: Whoops, I changed the example about five times while writing it, and must have got them out of sync.  Thanks for catching it.  Fixed in cf6469f.

source/language/scope.rst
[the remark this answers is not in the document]
blakejohnson: I wasn't actually advocating for being able to mutate globals within subroutines. What makes me uneasy is the ability to shadow variables in non-subroutine scopes.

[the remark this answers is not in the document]
jakelishman: Oh right, I understand.  This I think is the least surprising way of doing things.  It's necessary, for example, to allow the for-loop variable to shadow a previously-defined variable.  I can't think of any language off the top of my head with explicit variable declarations (e.g. C but not Python) that allow declarations in control-flow scopes, but don't permit shadowing.

source/language/scope.rst
```diff
@@ -0,0 +1,318 @@
...
+   int a = 1;
+   int b = 2;
+   const int c = 3;
+   const int d = 4;
+
+   def my_routine(uint a, uint c) {
+      // In this body, 'a' refers to the subroutine parameter, not the external
+      // variable, which wouldn't be visible even without the shadowing.
+
+      int local = 5;
```
psivaraj, line 194: **nitpick**: Use a different name than `local` since the word already carries a meaning in the discussion. Same deal on line 223.

source/language/scope.rst
```diff
@@ -0,0 +1,318 @@
...
+Some further examples:
+
+.. code-block:: c
+   :linenos:
+
+   OPENQASM 3.0
+
+   int ii = 100;        // 'ii' is declared in the global scope.
+   qubit[5] q;          // 'q' is declared in the global scope.
+   let alias = q[0:2];  // alias 'alias' is declared in the global scope.
```
psivaraj, line 281: **nitpick**: Use a different name than `alias` to disambiguate the variable and the meaning.

source/language/scope.rst
```diff
@@ -0,0 +1,318 @@
...
+     // has a different type.  The outer 'sum' is still accessible.
+
+     sum += ii;  // Iteration variable 'ii' is added to global 'sum'
+     if (sum > 10) {
+       float ii = 10.0; // For-loop iteration variable shadowed.
+       sum += uint(ii*2);
+     } else {
+       sum += ii;      // 'ii' is the for-loop iteration variable.
+     }
+
```
psivaraj, line 309: **thought**: Perhaps we should add an example like this 
```c
    int ii = 5;
    for uint i in [1:10] {
        ii += 2;
        int ii = 2;
        ...
    }
```

Here is equivalent C output 
```
iteration i = 0
ii before declaration = 7
ii after declaration = 2

iteration i = 1
ii before declaration = 9
ii after declaration = 2
```

source/language/scope.rst
```diff
@@ -0,0 +1,318 @@
...
+   // of type 'int' and 'const int' defined on lines 3 and 5 respectively.
+   // 'local' (from the subroutine body) is not in scope, while 'my_routine'
+   // (the subroutine) is.
+
+   const float[64] new_variable = 1.5;
+
+   def second_subroutine(qubit[4] q) {
+      int local = 8;
+
+      let alias = q[0:2];
```
psivaraj, line 225: nitpick: Use a different name than alias to disambiguate the variable and the meaning.

jakelishman: Done in dbc451c.

source/language/scope.rst
jakelishman: Done in dbc451c.

source/language/scope.rst
jakelishman: Good idea - done in dbc451c.

source/language/scope.rst
```diff
@@ -0,0 +1,338 @@
...
+   gate h q {
+      U(pi/2, 0, pi) q;
+      // `U` is the single-qubit gate that is implicitly defined in the global
+      // scope of all OpenQASM 3.0 programs, and so is available here.
+      // Similarly `pi` is one of the implicitly defined constants.
+   }
+
+   int i = 100;
+
+   include "my_definitions.qasm";
```
psivaraj, line 81: **thought**: The idea that "my_definitions.qasm" is not a valid QASM was a bit jarring. Is this unavoidable (or already present) in the language as it's now written?

jakelishman: [Right now, the spec says](https://openqasm.com/language/comments.html#included-files):

> The statement `include "filename";` continues parsing `filename` as if the contents of the file were inserted at the location of the `include` statement.

That implies that `filename` could be an incomplete OQ 3 program (like the example here).  I think it's worth making a positive decision whether that's exactly what we want, or if we want something different.

I think the two ends of the spectrum that we can define `include` on are:
- textual "copy-paste the contents of this file here" includes, like C's `#include`
- the file must be a completely semantically valid OQ 3 program, starting off with nothing in scope, and anything defined at the end becomes in scope in the containing context.

Since we don't have any visibility control mechanisms (namespacing / some sort of `public`/`private` keywords), any `include` system has issues bringing in the symbols it needs without causing clashes with other libraries.  You can't have `library_a.qasm` and `library_b.qasm` co-exist in the same file if they both `include "stdgates.inc";`, because they'll redefine all the symbols, and that's forbidden in OQ 3.

A work-around would be to say that the _user_ is responsible for `include`ing all the dependencies themselves, but that's not good extensible design for a language.  Some languages solve this with namespacing, some with explicitly marking the symbols made public from a file, and some with preprocessor conditional `include` guards.  Right now, we don't have the mechanisms for any of those solutions, so we're kind of requiring the "user supplies the dependencies" method.

levbishop: Is there any use case for having `include` anywhere other than global scope? For the common case of including some gate/extern defs it is only at global scope. I think 99% of use cases could be captured by saying you can only have an `include` at global scope and the included file must end at global scope (ie it shouldn't end half-way through a block). This wouldn't preclude nested includes (`main` includes `lib` that in turn includes `module`).

Namespacing is probably a good idea for a future enhancement something like `include "special_gates" prefix abc;` making all the global symbols introduced within the included file available with a prefix `abc::special_cnot` etc but i think that can be added backward compatible with the above

jakelishman: I agree it seems pretty useless anywhere but global scope (and the reference parser already rejects `include` in non-global scopes).  The problems of leaking gate definitions is independent of that, though.  As a more complete example, consider two files `library_a.qasm`:
```qasm
include "stdgates.inc";

gate gate_a q {
    x q;
}

// at this point, `gate_a` is in scope and will be exported,
// but so is everything defined in `stdgates.inc`.
```
and `library_b.qasm`:
```qasm
include "stdgates.inc";

gate gate_b q {
    z q;
}
```

So far this is ok, but now if I have my program:
```qasm
OPENQASM 3.0;

include "library_a.qasm";
// ok.
include "library_b.qasm";
// Error: cannot redefine {x, y, z, h, cx, ...} (everything in stdgates.inc).
```

Currently, the way around this is for neither library to actually `include` its dependencies, and to rely on the user to do that exactly once.  That requires that `include` statements start with all the outer identifiers in scope, which is how the current text is written.

There's a few other potential ways around that, including:
- some form of header-guard syntax (`#if !defined(_STDGATES_INC)`)
- some form of namespacing
- some form of marking what symbols should be exported from a file/scope (such as a `public` keyword)

source/language/scope.rst
```diff
@@ -0,0 +1,338 @@
...
+does not need to be the same as the type of the variable it is shadowing.
+
+Not all identifiers declared in outer scopes are visible within inner scopes.
+The visibility depends on the type of the variable, and the type of the inner
+scope.  Approximately, ``const`` variables, gates and subroutines are visible
+within *all* inner scopes, while other variables in outer scopes are visible
+within inner control-flow scopes but not gate and function scopes.
+
+The ``include`` statement should be seen as extending the scope it is contained
+within; all variables that are in scope at the time of the ``include`` statement
```
jakelishman, line 47: Make this clearer that the `include` is only global scope (and update the `include` section too).

source/language/scope.rst
jakelishman: From TSC 2022-06-10: we acknowledge that these effects aren't ideal, but there _is_ a workaround, and fixing it is out of scope for this PR.  We don't immediately agree on the details of an alternative, so we don't necessarily need to fix it for 3.0.

source/language/scope.rst
jakelishman: Done in 01ed7cc.

source/language/scope.rst
jakelishman: `include` is explicitly stated as global in 01ed7cc per the TSC decision.  I'll leave the rest of this comment chain for context in the future.

jakelishman: I've moved the context of this comment chain into its own issue in #365, so I'll mark this as resolved for this PR, since we decided not to make any changes here.

jakelishman:
Just for clarity: this is ready for TSC review, and has the changes discussed in the TSC meeting on Friday the 10th of June.  There are two comment chains marked unresolved pending responses, but I don't think any major changes are required from them.

source/language/scope.rst
psivaraj: Looks good.

source/language/scope.rst
```diff
@@ -0,0 +1,339 @@
...
+      // `U` is the single-qubit gate that is implicitly defined in the global
+      // scope of all OpenQASM 3.0 programs, and so is available here.
+      // Similarly `pi` is one of the implicitly defined constants.
+   }
+
+   int i = 100;
+
+   include "my_definitions.qasm";
+
+   // Identifiers 'h', 'my_gate', 'i' and 'j' are defined and in scope.
```
mbhealy, line 84:
```suggestion
   // Identifiers 'h', 'my_gate' and 'i' are defined and in scope.
```

I don't believe that `j` has been defined yet at this point

blakejohnson:
@stevenheidel I think you had asked to take one last look at this one?

stevenheidel (approved):

PR #337, opened 2022-04-06
Allow implicit casting and details for literals.

mbhealy:
### Summary
This PR is the result of discussions among the members of the Types and Casting Work Group, and introduces more detail in the spec regarding the rules of implicit casting. In general, classical standard types mimic the C99 spec. Angles and cbits do not implicitly cast to the standard types and vice versa, however, angles will promote/degrade to target angles with differing precision.

I've also included more detail on the format for literals in the language.

### Details and comments
This PR supercedes #326

PR #341, opened 2022-04-11
Optional type def for loop induction variable

mbhealy:
### Summary
This PR updates the spec and grammar to allow an optional type specifier for the `for` loop induction variable.

### Details and comments
This enables iteration over values larger than can fit in whatever default type might be assumed for the loop induction variable, and makes the casting of the iteration variable more explicit. This was discussed in the Types and Casting WG.


psivaraj (approved):

jakelishman (dismissed):

blakejohnson:
I definitely agree that we need to be able to provide a type for loop induction variable. The open question is whether to make the type specifier optional or required.

bettinaheim:
Just quickly capturing my comment from the meeting here: From an implementation perspective, I would encourage to treat the loop variable the same way as other assignments; i.e. if we require a type I'd require a type here, too. I think in principle, the type can be optional for both (now or in the future).

jakelishman:
To go with Bettina's comment, currently the following things require an explicit type:
- declaration of variables in global or local scopes, including those with the `const`, `input` or `output` modifiers, e.g.
  ```qasm
  { int[8] x; }
  ```
- subroutine parameters, e.g.
  ```qasm
  def my_routine(int[8] x) {}
  ```
- gate parameters when under `defcal` statements (I think - the spec doesn't say it outright), e.g.
  ```qasm
  defcal my_gate(angle[20] theta) $0 {}
  ```

The parameters to `gate` currently _do not_ require an explicit type, and (not trying to pre-empt the on-going discussion) it seems likely that we will allow but not require explicit typing there.

From the types group, the arguments I remember in favour of allowing it to be implicit were:
- it's convenient for the programmer
- if we wanted there to be some small amount of type inference on the variable (e.g. `int`/`uint`), it would be completely localised to a single statement (i.e. just read off the value from the join type of the index set using the implicit casting promotion rules)
- we allow implicit typing in the parameters to gate, so this isn't a unique feature
- it's what's in the paper

The arguments in favour of _requiring_ it to be explicit were:
- it means the type is always absolutely clear to the programmer
- it's more consistent with most other local declarations

(Please add/mention any I forgot - I think I was on holiday for at least one meeting where it was discussed.)

I personally prefer requiring an explicit type here (my argument is the same as Bettina's, but from a user-consistency perspective rather than implementation), but I don't want to override any group decision - the types group consensus was to write "a missing value implies `int`", as this PR currently says.

stevenheidel:
> we allow implicit typing in the parameters to gate, so this isn't a unique feature

The only reason we do this is to support OpenQASM 2. I don't think we would do the same thing if we were writing gates again from scratch (relates to discussion on #346)

> I personally prefer requiring an explicit type here (my argument is the same as Bettina's, but from a user-consistency perspective rather than implementation)

Adding a third voice to this point. 

Curious about https://github.com/Qiskit/openqasm/pull/337 though since that seems to make loop variables effectively `const`, which means the inferred type really `const int`.

blakejohnson:
I think you all have convinced me that we should make the type explicit.

jakelishman:
From the types meeting today:

My comment above is incorrect about the meaning of `const` - I was thinking about a C-style modifier, but that's not actually the defined semantics in OpenQASM 3.  Right now we don't have too much by way of `const` semantics for anything other than global-scope definitions (which become compile-time constants).  We agreed to discuss this in more detail with two initial topics:
- should `const` be a part of the type system, or should it be a keyword that specifically makes a variable compile-time constant (and consequently forbid any runtime expressions from ever being assigned to one).
- should we add constructs to the language to distinguish between C-style `const` (value decided at runtime, but after that it's read-only) and compile-time constants (e.g. something more akin to C's `#define`)

We also resolved to allow modifications to the variable within the `for` loop.  There's no motivating case for us to forbid this, and then it removes all the concerns about it having different (C-style `const`-like) semantics.  I will update this PR accordingly and fix the merge conflicts.

steleman:
> should const be a part of the type system, or should it be a keyword that specifically makes a variable compile-time constant (and consequently forbid any runtime expressions from ever being assigned to one).

I'm a bit confused about what this means.

In C && friends, `const` is a _type qualifier_. Compiler people sometimes refer to it as being a _CVR-qualifier_: `CVR` as in _const_, _volatile_, _restrict_.

In C && friends, and for scalars[0], a `const` qualified type variable is always initialized at compile-time. It's never run-time. A variable declared `const` cannot be modified at run-time after its declaration:

```
/* const.c */
#include <stdio.h>

int main()
{
  const int i = 3;
  i = 5;

  return 0;
}
```
```
%> clang -g -O2 -std=c99 -pedantic -Wall -Wextra -c const.c
const.c:6:5: error: cannot assign to variable 'i' with const-qualified type 'const int'
   i = 5;
~ ^
const.c:5:13: note: variable 'i' declared const here
   const int i = 3;
~~~~~~~~~~^~~~~
1 error generated.
```
In fact, the compiler will constant-fold the variable `i` to the numeric value `3`, and the symbol `i` will disappear (unless you explicitly ask the compiler to save the symbols for constant-folded variables).

> should we add constructs to the language to distinguish between C-style const (value decided at runtime, but after that it's read-only) and compile-time constants (e.g. something more akin to C's #define)

In C && friends, here is no functional difference between saying:

```
const int i = 5;
```
and saying:

```
#define i 3
```
```
/* constarray.c */
#include <stdio.h>

#define k 32

int main()
{
  const int i = 16;
  char a[i];
  char b[k];

  a[0] = 'a';
  b[0] = 'b';

  return 0;
}
```
```
%> clang -g -O2 -std=c99 -pedantic -Wall -Wextra -c constarray.c
%> echo $status
0
```

In C, the size of an array must be expressed as a compile-time constant. If `i` and `k` weren't functionally identical for the purpose of the Language spec, as compile-time constants, we would get an error for the declaration `char a[i];`.

-----
[0]. For pointers it's slightly more complicated. A `const` pointer can be reassigned or modified. The memory it _points_ to cannot be. But we don't have pointers in OQ3, so we don't need to worry about them.

jakelishman:
Apologies if my terminology is mistaken there.  By "runtime assignment" I meant that in C you can do things like
```c
int subroutine(int input)
{
    const int temp = input;
    printf("%d\n", temp * temp);
    return temp;
}
```
without any errors or warnings, even if you call that function with something that's only calculated at runtime.  So an object whose type is qualified `const` is immutable through that reference after initialisation, but that initialisation doesn't need to be at what I'd call compile-time (though it's entirely possible my terminology is wrong there).*

When I wrote the text 3 comments prior, I had this C-like model of the `const` keyword in mind for OpenQASM 3 as well, but that's incorrect as it's not properly supported by the specification.  OpenQASM 3's `const` is like a scoped version of C's `#define`.  The spec starts by saying that

> On declaration, they take their assigned value and cannot be redefined within the same scope

and then a couple of sentences later it says

> `const` values are compile-time constants, allowing the compiler to do constant folding and other such optimizations.

Taken together, that allows for a `const` value to be declared inside a nested scope, but it can't be given a value determined at runtime - it's just a compile-time value with a more restricted scope.  If all `const` values in OpenQASM 3 are required to be compile-time constants, it somewhat limits the usefulness of `const` other than in the global scope. Certainly the current text means that my original statement that the for-loop variable became "effectively `const`" in the body was incorrect.

The extra discussion is about whether we see a need for some sort of "read-only" modifier (like C's `const`) in addition to the "compile-time constant" keyword we already have (like C's `#define`).  As you say, we certainly need the compile-time constant keyword to simplify declaring arrays / integer bit-widths / register sizes.

---

\* I always found it really interesting that `const` in C doesn't even mean constant: `const volatile int` is a valid type, and the compiler isn't even allowed to assume that it stays the same within a scope.

steleman:
> it's just a compile-time value with a more restricted scope.

I still don't get it. `const` has nothing to do with scope. In C, `const` simply means "you only get one shot at assigning to it, and you can't change it after declaration", and that's it. It has nothing to do with scope. `const` behaves in the exact same way at global scope as it does at local scope.

> On declaration, they take their assigned value and cannot be redefined within the same scope.

I really don't understand what this sentence was meant to express. Does it mean that a `const` declaration must have an initializer? In C, it's legal to declare a `const` variable without initializer:

```
/* badconst.c */
#include <stdio.h>

int main()
{
  const int i;
  i = 3;
  return 0;
}
```

with predictable results:
```
clang -g -O2 -std=c99 -pedantic -Wall -Wextra -Wpedantic -c badconst.c
badconst.c:6:5: error: cannot assign to variable 'i' with const-qualified type 'const int'
   i = 3;
   ~ ^
badconst.c:5:13: note: variable 'i' declared const here
   const int i;
   ~~~~~~~~~~^
1 error generated.
```

Since OQ3 has no concept of stack, maybe it makes sense to require direct initialization on `const` declarations in OQ3.

blakejohnson:
The connection with scope is relevant to OQ3 because identifiers can be shadowed within `def`s. So, a `const` identifier can only be assigned to once within a given scope.

steleman:
> So, a `const` identifier can only be assigned to once within a given scope.

Like this:

```
/* localconst.c */
#include <stdio.h>

const int i = 5;  /* global scope */

int function(int x, const char* s) {
  const int i = x * 4;  /* local scope */

#if defined(BOMB_LOCAL)
  i += 17;
  (void) fprintf(stdout, "i=%i\n", i);
#endif

  if (i == 8 && s)
    return i + 3;

  return i + 100;
}

int main(int argc, char* const argv[])
{
  const int i = function(argc, argv[1]);  /* local scope */
  (void) fprintf(stdout, "i=%i\n", i);

#if defined(BOMB_GLOBAL)
  i += 31;
  (void) fprintf(stdout, "i=%i\n", i);
#endif

  return i;
}
```
```
%> clang -g -O2 -std=c99 -pedantic -Wall -Wextra -Wpedantic localconst.c -o localconst
%> echo $status
0
```
```
%> ./localconst
i=104
%>
(branch not taken)
```
```
%> ./localconst 10
i=11
%>
(branch taken)
```

```
%> clang -g -O2 -std=c99 -pedantic-Wall -Wextra -Wpedantic -DBOMB_LOCAL localconst.c -o localconst
localconst.c:9:5: error: cannot assign to variable 'i' with const-qualified type 'const int'
  i += 17;
  ~ ^
localconst.c:6:13: note: variable 'i' declared const here
  const int i = x * 4;
  ~~~~~~~~~~^~~~~~~~~
1 error generated.
```
```
%> clang -g -O2 -std=c99 -pedantic-Wall -Wextra -Wpedantic -DBOMB_GLOBAL localconst.c -o localconst
localconst.c:24:5: error: cannot assign to variable 'i' with const-qualified type 'const int'
  i += 31;
  ~ ^
localconst.c:20:13: note: variable 'i' declared const here
  const int i = function(argc, argv[1]);
  ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

jakelishman:
I think we're in agreement there.  OpenQASM's `const` has the same scoping rules as C's `const` (the ones you just showed), but unlike in C, any valued assigned to an OpenQASM `const`-modified type must be a compile-time constant.  OpenQASM currently doesn't have any type modifier/keyword that exactly matches the semantics of C's `const`.  The further discussion is about whether it would be useful to add that.

If we want to continue this thread right now, perhaps we can do it in a new issue?  I don't want to get too distracted from for loops here. We resolved in the types group to allow assignment to the for-loop variable within the loop, so this PR isn't going to involve any text about constants or immutability.

jakelishman:
I've just updated this PR to _require_ the type of the loop variable to be explicit, as discussed above, and with sign-off from the types-and-casting working group.

I've also included more text that allows the loop variable to be any scalar type, not limited to the `int` and `uint` (with optional widths) that it was before.  I also added more detail to the section on allowed targets for iteration, and newly allowed iteration through one-dimensional arrays and classical registers.  Please comment on those decisions - these particular ones are preliminary, and still under discussion in the types group, I just wanted to put up some sample text in order to discuss them.

The reason for removing the restriction on values being integers is that with an explicit type and the new implicit-promotion rules, we no longer need to do any form of type inference on the type of the loop variable, and so I don't think there is any technical need to restrict the values any more.

The argument in favour of allowing iterating through one-dimensional arrays and registers is that it is a simple syntactic sugar around iteration over integers and indexing, i.e. the following two loops are equivalent:

```qasm
bit[5] results;
let register = results[1:2];

// Direct iteration over the register.
for b in register {
   // use b as a value
}

// Iteration over indices of the register
for _tmp_ in [1:2] {
  bit b = _tmp_;
  // use b as a value
}
```

Note I haven't permitted iteration over _qubit_ registers.  I think this could actually be a very neat convenience that doesn't have problems of dynamic dereferencing of qubits, but it should be part of a larger discussion.

jakelishman:
Here's a link to review [just the changes to the example `.qasm` files and the spec text](https://github.com/Qiskit/openqasm/pull/341/files?file-filters%5B%5D=.qasm&file-filters%5B%5D=.rst&show-viewed-files=true).

levbishop:
This LGTM with required type any scalar type. I don't have a strong opinion either way on iterating over registers.

source/language/classical.rst
```diff
@@ -184,26 +189,85 @@ may optionally be followed by ``else { program }``.
...
    }
 
-The statement ``while ( bool ) { program }`` executes program until the Boolean evaluates to
+   // Loop over an array of floats.
+   array[float[64], 4] my_floats = {1.2, -3.4, 0.5, 9.8};
+   for float[64] f in my_floats {
+      // do something with 'f'
+   }
+
+   // Loop over a register of bits.
```
levbishop, line 256: Is there a guaranteed order in which the iteration hits the bits? Relevant in the situation that the body has side effects

jakelishman: I had assumed that registers have an order since they can be indexed with successive integers, and it'd be in that order.  I can make that explicit, assuming general agreement - it was an oversight that I didn't mention it.

jakelishman: That example was meant to assert that names bound in `let` statements are valid targets for iteration, rather than forbidding slicing.  I think that the question of slicing is more just one of the type system in general - if the type of `a[b:c]` is the same as something that can be iterated over, then it should be.  Right now that gets into funny questions about what the type of that expression is for registers - the existence of `let` makes things pretty weird.

jakelishman:
This is ready for TSC review.

Of note:
- this proposal does not currently allow using a `qubit[n]` type or the quantum aliases as the iteration variables
- this proposal does not answer the question of whether
  ```qasm
  qubit[5] q;
  for uint i in [0:2] {
      h q[i];
  }
  ```
  would result in an executable circuit, because the `q[i]` would be some form of dynamic dereference here.
  
The types group have a further proposal related to the `const` discussion above that, as a side-effect, could be used to define these "qubit loop" semantics.  It's a larger topic, though, so for another PR.

blakejohnson (approved):

levbishop:
@ajavadia i think you wanted a chance to comment here

jakelishman:
The last commit doesn't change the spec, it's just fixing merge conflicts.

ajavadia (approved):

blakejohnson (approved):

PR #342, opened 2022-04-13
clarify subroutine parameters documentation

ajberdy:
### Summary

Clarifying some confusing/inaccurate wording on the interface for parameters for subroutines.

### Details and comments

I believe the existing reference to `qargs` is outdated. I don't believe classical values are passed in by reference to  qargs, especially since the next line talk about classical types being passed by value to parameters. Aggregated the parameter-related sentences for better readability.

### DIFF
--- a/source/language/subroutines.rst
+++ b/source/language/subroutines.rst
@@ -2,13 +2,13 @@ Subroutines
 ===========
 
 Subroutines are declared using the statement ``def name(parameters) -> output_type { body }``.
-Zero or more quantum bits
-and classical values are passed to the subroutine by reference or name in ``qargs``.
-Classical types are passed by value in ``parameters``.
+The subroutine will define zero or more parameters as input, consisting of both quantum and classical arguments.
+Quantum bits and registers are passed in by reference or name, while classical types are passed in by value.
+All arguments are declared together with their type,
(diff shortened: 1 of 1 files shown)


source/language/subroutines.rst
```diff
@@ -2,13 +2,13 @@ Subroutines
 ===========
 
 Subroutines are declared using the statement ``def name(parameters) -> output_type { body }``.
-Zero or more quantum bits
-and classical values are passed to the subroutine by reference or name in ``qargs``.
-Classical types are passed by value in ``parameters``.
+The subroutine will define zero or more parameters as input, consisting of both quantum and classical arguments.
+Quantum bits and registers are passed in by reference or name, while classical types are passed in by value.
```
jakelishman, line 6: These are pre-existing problems, but since we're touching it right now:

- I'm not sure what "passed in by ... name" means entirely in this context.
- The passing type of classical values is a bit more complex than this - arrays are passed by reference.  We might want to just leave it as-is for the time being, though, since the types and casting WG will propose a PR to add a "by reference" modifier for subroutine arguments.

jakelishman (approved):
The actual changes look good to me, thanks, especially getting rid of the undefined term `qargs`.  I left a comment about some pre-existing issues, but we don't need to deal with them right now.

blakejohnson (approved):

Issue #323, opened 2022-02-04
RFC: `unitary` subroutines, as a gate-like subroutine acting on quantum registers and with more versatile control flow


ndebeaudrap:
### Summary

In the "Circuit Families & Generics" workgroup, we have drafted a specification for a variety of subroutine, which we propose to call `unitary` subroutines . These are in a sense intermediate between a `gate` subroutine and a `def` subroutine. The following is a high-level description of such subroutines, and the syntax with which they would be defined and invoked. We propose this as a feature to be added to OpenQASM3.

### Motivation

It is useful to be able to define subroutines which act on aggregate quantum data, grouped into quantum registers. This may be done in principle in OpenQASM3 using def subroutines, but doing so does not allow possibility of inverting these subroutines or controlling them coherently, even when the subroutine happens to consist entirely of unitary operations. Such 'circuit'-level unitary subroutines are commonplace in the quantum literature. Enabling such subroutines to be easily expressed will serve to increase the versatility of OpenQASM3 for programmers, and also for educators. This is also a preliminary step towards making it possible to define portable families of such operations (such as the quantum Fourier transform on varying register sizes) through subroutine declarations.

The distinct features which we aim to capture with `unitary` subroutines, are:
- in contrast to `gate` subroutines, they may take registers as quantum operands;
- also in contrast to `gate` subroutines, the subroutine definition may involve control flow using run-time constants;
- in contrast to `def` subroutines, these subroutines may be inverted and coherently controlled using gate modifiers `inv @` and `ctrl @` respectively.

The intent is to provide a means of defining unitary subroutines of greater versatility than is possible with `gate` subroutines, while retaining some of the advantages of a subroutine in which there is a clear separation between classical program and quantum data (e.g., no measurement or feedback loops between the two kinds of data).

A prototypical typical example would be a quantum Fourier transform, or similarly involved subroutine which one might like to use while also taking advantage of the fact that it is a unitary operation.

These subroutines would in principle be easily compiled to sequences of basic blocks similar to a top-level quantum program, even when invoked with modifiers such as `ctrl @` or `inv @`.

### Proposed syntax, with examples

We present the proposed syntax below, with simple 'toy' examples which may be more typical of an educational setting rather than application.

#### A simple unitary subroutine declaration

[...]

PR #346, opened 2022-04-21
Describe extension to ``gate`` functionality

jrcdebeaudrap:
### Summary

This PR changes the open specification, consistent with the proposals in (the latter half of) [Issue #323](https://github.com/Qiskit/openqasm/issues/323). This is submitted following discussions in the Circuit Families and Generics WG and in the TSC meetings, and collaboration with the Types and Casting WG.

### Details and comments

We introduce a supplemental, more versatile syntax for defining ``gate`` subroutines, and a more versatile specification of what a ``gate`` body may consist of.

 * This allows the classical parameters of a ``gate`` to have different types (the 'default' type of a ``gate`` is considered retrospectively to correspond to an ``angle``).*
  * It also allows the quantum operands of a ``gate`` to consist of qubit registers of a specific size (which are not considered to be subject to broadcasting).

To do this, the additional declaration syntax is expanded to permit the programmer to explicitly specify the types of all of the parameters and operands. Example gate invocations are provided to demonstrate how such gates may be invoked, and under what conditions the invocations are considered to be well-typed.

(* That the parameters of ``gate`` declarations are of type ``angle``, is a tentative position reached by the Types and Casting WG. Discussion of this point should involve other members of that WG. Note that the more versatile declaration syntax described in this PR, allows direct support of ``gate`` subroutines which accept ``float`` parameters instead, for those cases where this would be desirable.)

The ``gate`` body is allowed to contain program logic in the form of ``for`` loops and ``if`` statements, which may involve the classical ``gate`` parameters (or expressions which use those parameters), so long as these can be determined at compile-time. It is also allowed to involve declaration of classical storage types, subject to the same constrants as ``def`` subroutines, albeit with the limitation that locally defined identifiers (and classical parameters) may not be re-assigned values after they are initialised.


jwoehr:
Niel, there seems to be some automation fu blocking you signing the CLA ... can you please click the " [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user) " link and straighten out the angry machine?!

ndebeaudrap:
It appears to me, novice Git user that I am, that I did not configure my email properly on my local system. I do not believe that the email message in the commit is one that I would be able to 'verify', and that this is the root of the problem.

I am attempting a fix, having configured my local system better. If this does not work, I would appreciate advice on the best way forward in producing a better formed PR along these lines, given the current state of it.

ndebeaudrap:
(If — as I suspect — it would be expedient to simply remove this branch, please do so; I will try again once that has been done.)

jwoehr:
> (If — as I suspect — it would be expedient to simply remove this branch, please do so; I will try again once that has been done.)

Looks like it's fixed.

source/language/gates.rst
```diff
@@ -167,13 +167,13 @@ and applies the controlled gate
...
-Hierarchically defined unitary gates
-------------------------------------
+User-defined unitary gates
+--------------------------
 
-For new gates, we associate them with a corresponding unitary
-transformation by a sequence of built-in gates. For example, a CPHASE
-operation is shown schematically in :numref:`fig_gate`
-corresponding OpenQASM code is
+Programmers may define new gates, which may be resolved to a sequence of built-in
```
stevenheidel, line 173: Nit: `which can be resolved` - it's required not optional that resolution exists to a sequence of built-in gates.

source/language/gates.rst
```diff
@@ -233,24 +236,83 @@ of the gate definition.
...
+   }
+
+   // this is also ok:
+   gate qutritX : qubit[2] a {
+     x a[1];
+     cx a[1], a[0];
+     cx a[0], a[1];
+   }
+
+   // this is invalid:
```
blakejohnson, line 277: would be useful to call out why it is invalid

source/language/gates.rst
```diff
@@ -233,24 +236,83 @@ of the gate definition.
...
 
-The quantum circuit given by
+For either kind of gate declaration, the ``body`` may consist of the following:
+
+ * Declaration and initialisation of classical variables (but not re-assignment to them);
+
+ * Program logic (``if`` statements and ``for`` loops) with conditions/bounds involving constants,
+   and simple expressions depending on the gate arguments and local identifiers / loop iterators;
+
+ * Timing directives; and
```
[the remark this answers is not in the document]
blakejohnson: I guess the alternatives would connect `delay` to some kind of drift operator (e.g. a static coupling term) or a Kraus operator representing amplitude dampling and/or dephasing. But, all of these break out of the model of `gate`s having a 1-1 correspondence with unitaries.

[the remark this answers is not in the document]
stevenheidel: The point is that they look the same in one representation but they look different in another representation. Meaning a compiler storing gates as unitaries would think the gates interchangeable even though they aren't.

On the other hand that could already be accomplished today by adding an identity gate to a gate definition.

When a programmer puts a delay inside a gate definition, I'd guess that they really want a delay and would be unhappy if the compiler were to come up with a different implementation of the gate.

source/language/gates.rst
ndebeaudrap: Noted and amended

source/language/gates.rst
```diff
@@ -233,24 +236,83 @@ of the gate definition.
...
+If there are no variable parameters, the parentheses are optional. At least
+one quantum operand is required, and the quantum operands
+must be immediately preceded by a ``:`` delimiter.
+The arguments in ``typedQargs`` can be indexed within the body
+of the gate definition if, and only if, it is a register or array of qubits.
+
+.. code-block:: c
+
+   // this is ok:
+   gate g (angle alpha, int k): qubit[3] a
```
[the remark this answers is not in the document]
ndebeaudrap: Noted and amended

source/language/gates.rst
ndebeaudrap: Noted, I've amended this example and the previous one for the ``gate`` with the OpenQASM2 declaration syntax.

source/language/gates.rst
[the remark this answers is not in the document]
ndebeaudrap: Fair point; amended.

blakejohnson (dismissed):

source/language/gates.rst
```diff
@@ -230,27 +233,84 @@ of the gate definition.
...
+     U(0, 0, alpha/k) a[1];
+     U(0, 0, alpha/(k**2)) a[2];
+   }
+
+
+For either kind of gate declaration, the ``body`` may consist of the following:
+
+ * declaration and initialisation of classical variables (but not re-assignment to them);
+
+ * program logic (``if`` statements and ``for`` loops) with conditions/bounds involving constants,
```
stevenheidel, line 290: Ideally this would include an example of a gate defined with conditionals and loops. (I recall seeing examples in earlier versions)

taalexander (commented):
I have no major objections to this, other than raising the question of if this should be required for version 1.0?

levbishop:
My concern with this is that by changing from treating the parameters as exact angles to considering them as inexact types, then we can lo longer represent the mathematical equivalence between parameterized gates, which is the intention of the `gate` keyword. Example:
```openqasm
# from some common gate library:
gate g1(a, b, c) q {
    U(a*5, b/5, 0) q;
}
gate g2(a, b) q {
    U(a*3, b/3, 0) q;
}

# from hardware definition:
defcal g2(angle[13] a, angle[13] b) $0 {...}

# from user code:
angle[12] phi = pi/7;
angle[12] lam = pi/7;
g1(phi, lam) $0;
```
We'd like a compiler to be able to see this and infer the equivalence between `g1` and `g2` and rewrite `g1(phi, lam)` as `g2(3*phi/5, 5*lam/3)` which can then match with the `defcal` and the program can execute. The calculation of `3*phi/5` will be done at `angle[12]` precision which is what the author of the user code asked for and presumably they have determined this is sufficient precision for their algorithm, so that seems fine (maybe[^1]). When it goes into the `defcal` that result is cast to `angle[13]` because that is the best the hardware can support, which also seems fine and unavoidable.

If instead the include files have:
```openqasm
gate g1(angle[8] a, angle[8] b) q {
    U(a*5, b/5, 0) q;
}
gate g2(angle[8] a, angle[8] b) q {
    U(a*3, b/3, 0) q;
}
```
then the equivalence is unclear:

- Do we search for an `angle[8] x` such that the rounded `x*5` is equal to `angle[8] y = angle[8](angle[12](pi/7))*3`, and then cast back to angle[12]? What if there is no such `angle[8] x`? 
- Do we search for an `angle[8] x` such that the rounded `x/5` is equal to `angle[8] y = angle[8](angle[12](pi/7))/3`, and then cast back to angle[12]? What if there is more than one such `angle[8] x`?
- Do we just rewrite `3*phi/5` like in the arbitrary-precision case? If so, then what is the purpose of the type annotation in the `gate` definition?
- On what basis is the precision `angle[8]` (or other precision in the `gate` definition) to be chosen? It would be nice to write the `gate` in a generic way that's independent of both the user algorithm and the hardware implementation. For example, I think we'd like to give a `gate crz(theta) c, t` in hardware/algortihm independent way as in `stdgates.inc`.

[^1]: hmm. maybe assuming the precision of the user variable is appropriate for gate-rewriting is not ideal. What if the user algorithm is doing 4 rounds of iterative phase estimation and thus defines the phase as `angle [4] phase_est` and does the bit twiddling to make that work. If the gates involving `phase_est` need to be rewritten to the hardware, the user probably still wants that done at higher accuracy than `angle[4]`

jwoehr:
From a process point of view, it might be a good to aim for explicit syntax that won't be invalidated by later simplifying assumptions. Can it be the case for 3.0 that angle precision must, for now, either match perfectly or go unstated so as to let the compiler decide on the basis of the applicable `defcal`? 6 months of writing library code constrained thusly should reveal the correct path for the future.

jakelishman:
Just for reference for everyone: I asked Lev to write out some examples of his concerns ahead of the types-and-casting group meeting.  I'm planning to make his comment here the agenda for this week's meeting.

levbishop:
Pulled the parameter typing question out into a separate issue #375.
General consensus from discussion at the last TSC meeting was that the remainder of this PR is in good shape and can. be ready to merge after we have some decision on how to proceed on #375.

ndebeaudrap:
I believe, subject to the resolution of [#375](https://github.com/openqasm/openqasm/issues/375) (and in particular the presumption that OpenQASM2-style gate declarations take angles as parameters), this PR is ready to be merged. Do we concur?

bettinaheim:
Looks good to me. I think this is just waiting for the updates from Lev, and then this should tie in nicely.

blakejohnson:
I assume this will need to be rebased after #393 is merged, but I agree that this seems fine to proceed following that change.

blakejohnson:
#393 is finally merged. Can we refactor this change set accordingly?

hodgestar:
I've merged main into this branch and resolved conflicts. The text likely needs a final round of clean ups and reviews. I'm happy to do the initial round of clean ups but reviews and comments welcomed.

blakejohnson:
Thank you @hodgestar for doing that work. I will add re-reading this to my TODO list.

source/language/gates.rst
```diff
@@ -103,9 +103,126 @@ definitions for which :ref:`defcal blocks <pulse-gates>` are known. A more sophi
 might use the `gate` definitions of the gates with associated ``defcal`` blocks to
 build a gate library, and use methods based on KAK decompositions to rewrite into this hardware library.
 
-The ``gate`` statement also allows defining parameterized families of unitaries. For example, a CPHASE
-operation is shown schematically in :numref:`fig_gate`
-and the corresponding OpenQASM code is
+Controlled gates can be constructed by adding a control modifier to an existing gate. For example,
```
blakejohnson, line 106: Prior to this change set, the discussion of controlled gates sat under a "quantum gate modifiers" section. I'd suggest preserving that organizational structure.

blakejohnson: Actually, this entire section of looks like it refers to a stale version of the "quantum gate modifiers" section.

source/language/gates.rst
```diff
@@ -122,18 +239,20 @@ and the corresponding OpenQASM code is
...
 .. figure:: ../qpics/gate.svg
 
-   New gates are defined from previously defined gates. The gates are applied using the statement
-   ``name(params) qargs;`` The parentheses are optional if there
-   are no parameters. The gate :math:`{cphase}(\theta)` corresponds to the unitary matrix
-   :math:`{diag}(1,1,1,e^{i\theta})` up to a global phase.
+New gates are defined from previously defined gates.
+The gates are applied using the statement ``name(params) qargs;`` just like the built-in gates.
+The parentheses are optional if there are no parameters. The gate :math:`{cphase}(\theta)`
+corresponds to the unitary matrix :math:`{diag}(1,1,1,e^{i\theta})` up to a global phase.
```
blakejohnson, line 245: This change looks fine.

source/language/gates.rst
```diff
@@ -122,18 +239,20 @@ and the corresponding OpenQASM code is
...
 Again, this definition does not imply that ``cphase`` must be implemented with
 this particular series of gates. Rather, we have specified the unitary
 transformation that corresponds to the symbol ``cphase``. The particular
 implementation is up to the compiler, given information about the basis
 gate set supported by a particular target.
 
-In general, new gates are defined by statements of the form
+In general, new gates may be declared in two different ways: a 'short'
+declaration syntax, and a more versatile 'general' declaration syntax.
+'Short' gate declarations are statements of the form
```
blakejohnson, line 255: This looks like the start of the changes we want to keep.

source/language/gates.rst
```diff
@@ -164,23 +283,140 @@ of the gate definition.
...
-statements are valid.
+'General' gate declarations have a similar structure, with minor differences:
 
-The statements in the body
-can only refer to the symbols given in the parameter or argument list,
-and these symbols are scoped only to the subroutine body.
+.. code-block:: c
 
-An empty body corresponds to the identity gate.
+   // comment
```
blakejohnson, line 293: The ``// comment` isn't part of the gate declaration, no?

source/language/gates.rst
```diff
@@ -164,23 +283,140 @@ of the gate definition.
...
+          for uint k in [j+1 : n-1]
+           ctrl @ Rz(pi / 2**(k-j+1)) q[j], q[k];
+          }
+        for uint j in [0:n-1]
+          if (j != n-1-j)
+            swap q[j], q[n-1-j]
+  }
+
+Classical storage types and parameters in a ``gate`` body are treated as being immutable,
+and cannot be assigned to more than once. (For loop induction variables are treated as being constant within
```
blakejohnson, line 361: I struggled to scan "for loop" as a compound noun, and so was confused by the parenthetical statement. Would it be clearer to just say "(Loop induction variables ...)"?

source/language/gates.rst
```diff
@@ -164,23 +283,140 @@ of the gate definition.
...
+the scope of any single iteration of the ``for`` loop, and can only be modified by the logic of the loop
+itself.) External functions, ``reset`` operations, ``measure`` operations (or other operations with potentially
```
blakejohnson, line 363:
```suggestion
itself.) ``extern`` functions, ``reset`` operations, ``measure`` operations (or other operations with potentially
```
Probably better to say `extern` rather than "External".

source/language/gates.rst
```diff
@@ -164,23 +283,140 @@ of the gate definition.
...
+random outcomes), cannot be involved in the program logic of a ``gate`` body. This ensures that an invocation of
+a user-defined ``gate``, corresponds to a definite finite sequence of built-in unitary gates.
```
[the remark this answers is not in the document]
blakejohnson: I think the intent here was still to be able to associate generalized `gate`s with unitary matrices, but allowing convenient short hands. I suppose we have failed to do that, though, because it is trivial to construct examples where this is no longer true, e.g.:
```
gate classically_controlled_not(bit b): qubit q {
    if (b) { x q; }
}
```
I'm not sure if there are additional constraints we could introduce to restore this property? Or do we just give up on trying to keep it and allow `gate`s to not always be associated with unitaries?

source/language/gates.rst
```diff
@@ -164,23 +283,140 @@ of the gate definition.
...
+
+An empty ``body`` corresponds to the identity gate.
+
+Gates must be declared before use, and 
 cannot call themselves. The statement ``name(params) qargs;`` applies the gate,
-and the variable parameters ``params`` can have any type that can promote to ``angle`` type.
+and the variable parameters ``params`` must have the appropriate type (or be expressions which can be implicitly 
+cast to the appropriate type).
+
+The quantum operands of a ``gate`` invocation must have the appropriate types, to the declaration of the ``gate``.
```
blakejohnson, line 374:
```suggestion
The quantum operands of a ``gate`` invocation must have the appropriate types to the declaration of the ``gate``.
```

source/language/gates.rst
```diff
@@ -164,23 +283,140 @@ of the gate definition.
...
+   gate g : qubit qb0, qubit qb1, qubit[5] qreg
+   {
+     // body
+   }
+   qubit a[2];
+   qubit b[2];
+   qubit c[5];
+   qubit d[10];
+   g a, b, c; // ok: performs "g a[0], b[0], c; g a[1], b[1], c;"
+   g a, b, d; // error! third operand expects a register of size 5, not 10
```
blakejohnson, line 418: This entire section can be replaced with a reference to the broadcasting section above.

blakejohnson:
Having refreshed my memory on this topic by re-reading the PR, I am still broadly in favor of this change. However, I think @braised-babbage put his finger on the critical issue of whether this change breaks the correspondence between `gate`s and unitary matrices. If so, we either need to accept that, or figure out other changes to the proposal.

source/language/gates.rst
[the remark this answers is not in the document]
meamy: Just weighing in here since this popped up in my inbox --- I don't see the example above as disassociating gates from unitary matrices. Rather, a gate is a family of unitary matrices defined by the values of all classical parameters. In this sense it's morally no different (I believe) from the interpretation of openQASM 2.0 style gates, it's just allowing other types of parameters and a different type of computation if the compiler actually intended on generating the unitary given a set of parameters. This proposal was a long time ago however so my memory of what was proposed and the issues may be a bit fuzzy.

source/language/gates.rst
braised-babbage: Agreed. For example, the broadcasting discussion below is already discussed adequately elsewhere https://openqasm.com/language/gates.html#broadcasting.

braised-babbage:
Following up on our discussion of this at the last TSC, there are a few conceptual pieces here.

1. The introduction of a typed parameter list for parametric gates (allowing classical parameters beyond angles). 
2. The introduction of a typed quantum argument list for gates (allowing more compact and sophisticated description of gate applications).
3. An enrichment of what is allowed in the gate body (namely, `if` and `for` statements which involve only only `const` or gate parameter values in a way that can be resolved at compile time).

To a degree, these items can be considered separately.

For example, item (1) above can be useful when considering the interaction between OpenQASM's `gate` and `defcal` constructs. Whereas a gate formally represents a unitary operation, a `defcal` provides an operational description of how to perform such an operation. At present, a `defcal` cannot be parameterized by meaningful quantities (e.g. DAC scale factors, IQ classification thresholds, etc) that may otherwise be manipulated by a low-level OpenPulse program because these quantities are not of `angle` type. Without such parametrization, calibration definitions necessarily rely on a certain amount of global state. For this reason I find (1) to be of some interest, even if we do not expect that the corresponding `gate` definitions to have any mathematical dependence on these classical quantities. 

While there is an interaction between (2) and (3), (e.g. if we have a qubit register we can loop over, this motivates supporting `for` statements in a gate body), we can still consider them separately. Fundamentally (3) is of interest to people defining gates, whereas (2) is also of interest to people invoking gates. The current [broadcasting](https://openqasm.com/language/gates.html#broadcasting) mechanism can allow for compact expression e.g. of transversal operations, but is still relatively transparent: when you read a gate application, you can tell from the arguments what is going on. With the extension proposed here, we lose this property -- a reader of the program is ultimately going to have to dig in to the gate definition to make sense of it.

source/language/gates.rst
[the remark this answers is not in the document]
blakejohnson: I see your point, @meamy. The thing which is changed is that I no longer have a mathematical structure for the associated unitary which is fixed other than for substitution of a variable into some algebraic expressions. But, my example is definitely still unitary for any value of `b`.

source/language/gates.rst
```diff
@@ -164,23 +283,140 @@ of the gate definition.
...
+     U(0, 0, alpha/(k**2)) a[2];
+   }
+
+
+For either kind of gate declaration, the ``body`` may consist of the following:
+
+ * declaration and initialisation of classical variables (but not re-assignment to them);
+
+ * program logic (``if`` statements and ``for`` loops) with conditions/bounds involving constants,
+   and simple expressions depending on the gate arguments and local identifiers / loop iterators;
```
blakejohnson, line 341: I think we need to be more precise about what a "simple expression" is.

source/language/gates.rst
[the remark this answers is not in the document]
meamy: Indeed, you could say it takes it from a `matrix with symbolic expressions inside` to a `symbolic expression of a matrix`. The question comes down to which is more useful in practice, and how that balances against the perspective of the programmer (if openQASM is intended to be a programming language rather than an IR) and other considerations like broadcasting.

braised-babbage:
Following up on today's TSC meeting (8/7/24), I created an issue https://github.com/openqasm/openqasm/issues/547 representing a kind of "item 0" for https://github.com/openqasm/openqasm/pull/346#issuecomment-2176983470.

blakejohnson:
Superseded by #346 which implements a subset of this changeset.

PR #347, opened 2022-04-22
Correct type of arguments in QuantumGateDefinition

### LINKED ISSUE #345: Inconsistent definition of AST Node QuantumGateDefinition
### OpenQASM version:

3.0

### Python version:

3.8.13

### Operating system:

macOS Monterey 12.3.1

The current [definition](https://github.com/Qiskit/openqasm/blob/c39eac9f5c87b80df9e8eaeced96cbf4b477e5d2/source/openqasm/openqasm3/ast.py#L172-L188) of `QuantumGateDefinition` in `openqasm3.ast` is inconsistent with the definition in the [grammar](https://github.com/Qiskit/openqasm/blob/c39eac9f5c87b80df9e8eaeced96cbf4b477e5d2/source/grammar/qasm3.g4#L213-L219): its `arguments` field is of type `List[ClassicalArgument]`, whereas according to the grammar it should be `List[Identifier]`.

Note that the [parser](https://github.com/Qiskit/openqasm/blob/c39eac9f5c87b80df9e8eaeced96cbf4b477e5d2/source/openqasm/openqasm3/parser.py#L231-L248) actually does create the `QuantumGateDefinition` node according to the grammar.

N/A

The `QuantumGateDefinition` AST node should be consistent with the grammar.

Change the definition of `QuantumGateDefinition.arguments` field to `List[Identifier]`.

jakelishman:
### Summary

The ANTLR to AST conversion was taking the type of the arguments to be
`Identifier`, but the AST hierarchy itself wanted them to be
`ClassicalArgument`.  The spec currently says that they are
"arbitrary-precision numerical types", which does not line up with
`ClassicalArgument` (which would include, e.g. `IntType`), and
`Identifier` is more correct for the current spec, though this may
change in the future.

### Details and comments

Fix #345 (the immediately actionable parts - the discussion touches on other on-going discussions, but those are orthogonal to #345 itself).

### DIFF
--- a/source/openqasm/openqasm3/ast.py
+++ b/source/openqasm/openqasm3/ast.py
@@ -183,7 +183,7 @@ class QuantumGateDefinition(Statement):
     """
 
     name: Identifier
-    arguments: List[ClassicalArgument]
+    arguments: List[Identifier]
     qubits: List[Identifier]
     body: List[QuantumStatement]
 


stevenheidel:
> though this may change in the future

This would need to change after #346 I imagine?

stevenheidel (approved):

jakelishman:
Yeah, exactly - I don't know how long #346 will take to merge, so I'm just trying to get on top of things as they stand.

PR #331, opened 2022-03-14
change references to master for references to main

1ucian0:

PR #349, opened 2022-05-04
Update target branch of deployment action

jakelishman:
### Summary

The current CD action is still targeting the `master` branch, which isn't getting pushed to any more.  This updates it to the right `git` ref.

### Details and comments

A minor oversight in #331.


jwoehr (approved):
Looks good, thanks for catching that, @jakelishman

PR #352, opened 2022-05-10
Openpulse reference implementation

aspcompiler:
### Summary

This is a reference implementation of [openpulse](https://qiskit.github.io/openqasm/language/openpulse.html), based on the experimental compiler developed at AWS Center for Quantum Computing. It is implemented as an extension to OpenQASM, i.e., it imports and extends openqasm3 in both grammar and Python parser/AST. This import and extend approach allows us to incorporate the latest changes in openqasm3 with minimum changes in openpulse.

### Details and comments

The openpulse grammar extends the openqasm3 grammar in following ways:
1. It add three new pulse types, `port`, `frame` and `waveform`.
2. It extends openqasm3 `classicalType` with `calType` to use the pulse types.
3. It overrides `calibrationDefinition` and adds `calBlock` where statements involving pulse types can be used.

In the Python code, we added the `openpulse` package. The `openpulse.ast` module reexports `openqasm3.ast` classes and adds/overrides a few. Likewise, the `openpulse.parser` module reuses most `visit*` methods from the `openqasm3.parser` module. We did not add another visitor as `openqasm3.visitor` works fine with `openpulse.ast`.

This PR incorporated (i.e., implemented on top of) several recent changes:
1. [Frame manipulation](https://github.com/openqasm/openqasm/pull/354).
2. [Antlr grammar and Python parser refactoring](https://github.com/openqasm/openqasm/pull/351)
3. [Formatter/printer](https://github.com/openqasm/openqasm/pull/362)


source/openpulse/openpulse/parser.py
```diff
@@ -0,0 +1,264 @@
...
+        if ctx.WAVEFORM():
+            return ast.WaveformType()
+        if ctx.PORT():
+            return ast.PortType()
+        if ctx.FRAME():
+            return ast.FrameType()
+
+    @span
+    def visitArrayLiteral(self, ctx: openpulseParser.ArrayLiteralContext):
+        array_literal_element = (
```
aspcompiler, line 133: Unfortunately, we had to duplicate a few methods from openqasm3.parser.QASMNodeVisitor because of the instance check. The instances, such as `openpulseParser.ExpressionContext` below and the counterparts in `qasm3Parser` are independently generated from Antlr.

source/openpulse/openpulse/ast.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import List, Optional
+
+# Re-export the existing AST classes from openqasm3
+# pylint: disable=unused-import
+from openqasm3.ast import *
```
aspcompiler, line 19: We import everything from openqasm3.ast to reuse, but override `CalibrationDefinition`.

source/grammar/openpulseParser.g4
```diff
@@ -0,0 +1,70 @@
+/***** ANTLRv4  grammar for OpenPulse. *****/
+// This grammar is OpenPulse extension for QOpenQASM3.0
+// We introduce several new types for OpenPulse.
+// We override several nodes in qasm3 to consume the new types.
+// The OpenPulse grammar can be used inside cal or defcal blocks.
+
+parser grammar openpulseParser;
+import qasm3Parser;
```
aspcompiler, line 8: We import the qasm3Parser, override `defcalStatement` with concrete implementation of OpenPulse. We also override `scalarType` to incorporate `OpenPulse` types and override `statement` to add `calStatement`.

source/openpulse/openpulse/parser.py
```diff
@@ -0,0 +1,264 @@
...
+    "visitTerminal",
+    "visitArrayLiteral",
+    "visitCalibrationGrammarStatement",
+    "visitDefcalStatement",
+    "visitIndexOperator",
+    "visitRangeExpression",
+    "visitReturnStatement",
+    "visitScalarType",
+]
+for m in QASMNodeVisitor.__dict__:
```
aspcompiler, line 262: We import most methods from `openqasm3.parser.QASMNodeVisitor`. We exclude the one we override. We also had to reimplement a few extra due to `ininstance` check like the one mentioned above.

source/openpulse/openpulse/printer.py
```diff
@@ -0,0 +1,76 @@
...
+    does not need to be the case.
+
+    For more details on the available keyword arguments, see :ref:`printer-kwargs`.
+    """
+    out = io.StringIO()
+    dump(node, out, **kwargs)
+    return out.getvalue()
+
+
+class Printer(QASMPrinter):
```
aspcompiler, line 33: OpenPulse `Printer` is a straightforward subclass of `openqasm3.printer.Printer`.

examples/t1.qasm
```diff
@@ -23,8 +23,8 @@ defcal x $q {
    play drive($q), gaussian(100, 30, 5);
 }
 
-for p in [0 : points-1] {
-    for i in [1 : shots] {
+for uint p in [0 : points-1] {
```
aspcompiler, line 26: Wonder why this wasn't caught earlier by https://github.com/openqasm/openqasm/pull/341? Maybe the `defcal` statement earlier matched to the end of the file in the OpenQASM test?

source/openpulse/tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+
+import pytest
+
+import openpulse
+
+TEST_DIR = pathlib.Path(__file__).parent
+ROOT_DIR = TEST_DIR.parents[2]
+EXAMPLES_DIR = ROOT_DIR / "examples"
+# The following 2 examples need to fixed to match the latest openpulse specification
+EXCLUDES = ["defcal.qasm", "pong.qasm"]
```
aspcompiler, line 13: Skipping these 2 examples because the defcal body cannot be passed by openpulse, e.g, `gaussian(...)`.

jakelishman: If those examples are valid OpenPulse, we probably need to update those examples.

examples/t1.qasm
jakelishman: Yeah, that'll be exactly the issue.  Now the ANTLR grammar refactor is in place, I'll make a PR to fix those issues.

source/grammar/openpulseParser.g4
jakelishman: My interpretation of the spec is that the calibration grammar components don't take effect until the _block_ of the `defcal` is entered, so parsing the identifiers should be the responsibility of the OpenQASM 3 component of the parser.  This particular case might be a bug in the OQ 3 ANTLR grammar - I have a small comment on the bottom of #367 about it, but assuming we want to keep that, I'll update the OQ 3 grammar.

source/grammar/openpulseLexer.g4
```diff
@@ -0,0 +1,7 @@
+lexer grammar openpulseLexer;
+import qasm3Lexer;
+
+CAL: 'cal';
```
jakelishman, line 4: This should be parsed by the OQ 3 parser, not the OpenPulse one - it's a bug that the `cal` statement is missing from there.  I've opened #368 to track that.

source/grammar/openpulseParser.g4
```diff
@@ -0,0 +1,70 @@
...
+
+options {
+    tokenVocab = openpulseLexer;
+}
+
+/**** OpenQASM3.0 overrides ****/
+defcalStatement: DEFCAL Identifier (LPAREN argumentDefinitionList? RPAREN)? hardwareQubitList returnSignature? LBRACE statement* RBRACE;
```
jakelishman, line 15: Per my interpretation of the OQ 3 spec, the `defcal` statement should be valid OQ 3, only the _block_ should be in the calibration grammar, so we shouldn't need to do this bit, I think.  `defcal` blocks presumably shouldn't be able to take or return `waveform` (because it's not an OQ 3 type), and the qubit lists parts should be handled by the OQ 3 parser.

There's definitely bugs in the reference ANTLR one surrounding this, but I also think there's a little bit of mixing in the spec between where exactly calibration-grammar-specific constructs are allowed.  I opened #367 about that.

jakelishman (commented):
It's definitely good to get an OpenPulse reference implementation out into the wild as well, thanks!

That said, this isn't quite the packaging direction I expected we'd take, and I'm not sure defining the OpenPulse implementation as an extension of the OpenQASM 3 parser is the right path forwards.

I expected the OQ 3 parser to the be "controller" parser, and for it to "delegate" certain parts of the character stream to subparsers, dependent on the `defcalgrammar "...";` statements the controller parser has encountered.  I expected to do that by having a registration / entry-point method in the `openqasm3.parser` package, and for us to write something along the lines of:
```python
class QASM3NodeVisitor:
    def get_calibration_parser(self):
        try:
            return PARSERS[self._calibration_grammar]
        except KeyError:
            raise QASM3Error(f"No parser is registered for calibration grammar '{self._calibration_grammar}'.")

    def visit_DefcalgrammarStatement(self, ctx):
        self._calibration_grammar = ctx.StringLiteral().getText()[1:-1]
        # ... handle the node ...

    def visit_DefcalStatement(self, ctx):
        identifier = self._visit_identifier(ctx.Identifier())
        signature = self.visit(ctx.Signature())
        return_type = self.visit(ctx.ReturnType())
        calibration_parser = self.get_calibration_parse()
        body = calibration_parser(ctx.body.getText())
        # ... build the AST node, where `body` is of type `Any` ...
```
The idea with this is that it clearly defines the separation between the calibration grammar and OpenQASM 3, which are (to my understanding at least), potentially _entirely_ separate languages.  This lets us built up the reference parser by composition - anybody can write their own new calibration grammar and parser, and plug it into the existing reference implementation.  As well as being (to me) more faithful to the spirit of the spec and easier for users, I think this will be easier to maintain, too - calibration-grammar maintainers will only need to be concerned with things that are strictly part of the calibration grammar.

The form in this PR, where the OpenPulse parser is a strict superset of the OpenQASM reference parser, doesn't seem quite right to me.  The most general parser of OpenQASM 3 (which is ideally what a reference implementation is - again, to me, at least), should be able to dispatch to the correct calibration-grammar parser based on the active `defcalgrammar` statement.  This form requires the user to call the correct parser ahead of time, so the `defcalgrammar` statement isn't really parsed/interpreted.

source/openpulse/README.md
```diff
@@ -0,0 +1,32 @@
+# OpenPulse Python Reference
+
+[![License](https://img.shields.io/github/license/Qiskit/openqasm.svg)](https://opensource.org/licenses/Apache-2.0)<!-- long-description-skip-begin -->[![Release](https://img.shields.io/pypi/v/openqasm3)](https://pypi.org/project/openqasm3)<!-- long-description-skip-end -->
```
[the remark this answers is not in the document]
aspcompiler: Updated the link. I also claimed `openpulse` on pypi.

source/grammar/openpulseParser.g4
aspcompiler: So do you envisioned that we construct OpenPulse parser/ast, for the body of `cal` and `defcal`, completely independent of OpenQASM?

jakelishman:
Update: Li, Prasahnt and I met to discuss this PR, and resolved that:
- it would be nice to have a reference implementation that delegates the parsing of the pulse blocks as opposed to the inheriting superset parser that this PR is.  The current form is better for a more performant implementation, but is not the most general OpenQASM 3 parser.
- the design of OpenPulse inherits a lot of the types and expression rules from OpenQASM 3, so there will always be a fair amount of duplication in the grammar and parser, and it's not clear what's clearest for users here
- pragmatically, we need a reference implementation of OpenPulse for the first release, and I (being one of the active code maintainers on the repository) will be on paternity leave for the next several weeks, so it's best to merge this parser with this type of design (subject to other review).
- if we want to change the reference parser, we can do it in the future.

So I will try to provide a review of the parser in this form, and look to get it merged before I go on paternity leave.

source/openpulse/README.md
```diff
@@ -0,0 +1,32 @@
+# OpenPulse Python Reference
+
+[![License](https://img.shields.io/github/license/openqasm/openqasm.svg)](https://opensource.org/licenses/Apache-2.0)<!-- long-description-skip-begin -->[![Release](https://img.shields.io/pypi/v/openpulse)](https://pypi.org/project/openpulse)<!-- long-description-skip-end -->
+
+OpenPulse is a minor extension to OpenQASM 3 with a concrete implementation of pulse grammar.
```
taalexander, line 5: Can we add a TODO note here about refactoring the grammar as @jakelishman has suggested?

source/openpulse/openpulse/parser.py
```diff
@@ -0,0 +1,264 @@
...
+        " such as by 'pip install openqasm3[parser]'."
+    ) from exc
+
+from openqasm3.parser import (
+    get_span,
+    add_span,
+    span,
+    QASMNodeVisitor,
+    _visit_identifier,
+    _raise_from_context,
```
taalexander, line 44: Should these methods be renamed to be public, ie., remove the snakecase?

source/openpulse/openpulse/parser.py
taalexander, line 262: IMO, this is quite sketchy. It may be better to just copy and paste?

source/openpulse/setup.cfg
```diff
@@ -0,0 +1,39 @@
+[metadata]
+name = openpulse
+url = https://github.com/Qiskit/openqasm
```
taalexander, line 3:
```suggestion
url = https://github.com/Qiskit/openqasm/source/openpulse
```
?

source/openpulse/tests/conftest.py
taalexander: Have these been updated?

source/openpulse/tests/test_openqasm_tests.py
```diff
@@ -0,0 +1,13 @@
+# Copied from openqasm3. OpenPulse parser should be able to parse openqasm3 examples
+import openpulse
+
+
+def test_examples(example_file):
+    """Loop through all example files, verify that the ast_parser can parse the file.
+
+    The `example_file` fixture is generated in `conftest.py`.  These tests are automatically skipped
```
taalexander, line 8:
```suggestion
    The `example_file` fixture is generated in `conftest.py`. These tests are automatically skipped
```

source/openpulse/tests/test_openqasm_tests.py
```diff
@@ -0,0 +1,13 @@
+# Copied from openqasm3. OpenPulse parser should be able to parse openqasm3 examples
+import openpulse
+
+
+def test_examples(example_file):
```
taalexander, line 5: Is this parameterized as a separate pytest test for each test file?

source/grammar/openpulseParser.g4
taalexander: For tracking of discussion for my own review see - https://github.com/openqasm/openqasm/pull/352#issuecomment-1179209502

taalexander (changes requested):
Some thoughts:
1. I agree with @jakelishman about the ideal implementation structure. However, in the name of making progress this PR is a good first step.
2. Are we missing license/copyright headers in the source files? Or have these not yet been added for the project?
3. What other changes do we need to get into this PR (if any) to make sure we can merge this?

source/grammar/openpulseParser.g4
aspcompiler: > My interpretation of the spec is that the calibration grammar components don't take effect until the block of the defcal is entered, so parsing the identifiers should be the responsibility of the OpenQASM 3 component of the parser

That is true. However, the `defcal` body can contain classical statements. We took the liberty of reusing the OpenQASM classical statements in OpenPulse (so we don't have to introduce a different grammar). Therefore we try to reuse qasm3Parser to pase the classical statements in OpenPulse.

source/openpulse/tests/conftest.py
aspcompiler: @jakelishman removed `pong.qasm` in #376 . However, `defcal.qasm` still contains some abbreviations such as:

```
defcal x $0 {
   play drive($0), gaussian(...);
}
```

This example was from a [couple years ago](https://github.com/openqasm/openqasm/commit/8c5d723aee624a8f044dbceb469d3ae483d36029#diff-e54db80c3f513f2802d83657bc1316420cc14a08bd3c943f5aace58d201e5e50) and the syntax has changed significantly so we should consider to create a separate issue to fix this example.

source/openpulse/setup.cfg
aspcompiler: I updated the url to `https://github.com/openqasm/openqasm/` for now. To further link into the project, github added `tree/main` in the url, such as `https://github.com/openqasm/openqasm/tree/main/source/openqasm`. We don't have a valid url for `openpulse` yet until it is merged into main.

source/openpulse/openpulse/parser.py
aspcompiler: Just want to make sure I understand. These methods are imported from `openqasm3.parser`. Do you suggest that I modify `openqasm3.parser` to make them public?

source/openpulse/openpulse/parser.py
aspcompiler: I changed these imports to explicit imports.

source/openpulse/README.md
aspcompiler: Added TODO to README.

aspcompiler:
@taalexander and @jakelishman, I have worked through your comments and try to respond. Please take another look to see if there is anything not sufficiently addressed. One thing that I was not sure is about your license header comment. Since none of the files in [openqasm](https://github.com/openqasm/openqasm/tree/main/source/openqasm) directory has license header, I wonder what convention I should follow.

During the cause of this work, I have grow increasing suspicious whether we can merge the PR in this form. The main concern is that OpenPulse imports classical types/statements from OpenQASM to use in `cal`/`dercal` body. Any movement in OpenQASM causes the same movement in OpenPulse. Each time when I merged latest OpenQASM into branch, a seemingly small change in OpenQASM can cause multiple unit tests to fail in OpenPulse. This could become an experience issue for many OpenQASM contributors as these fail unit tests will force them to dive deep into OpenPulse in order to pass the tests. We will discuss this in the next TSC meeting for possibly ways around this, including the possibility of maintaining OpenPulse in a separate repo until OpenQASM becomes more stable.

psivaraj:
Following up on @aspcompiler comments, I want to provide some context to the TSC on our current roadblock and 3 options for how to proceed. 

# Context

In the [spec](https://openqasm.com/language/openpulse.html) and in the [paper](https://arxiv.org/pdf/2104.14722.pdf), we allow calling *OpenQASM* constructs from within the *OpenPulse* `defcal`s. In other words, *OpenPulse* is in practice treated as superset of *OpenQASM* even though it's not explicitly stated as such. 

For example, pg. 37 of the paper has a snippet that uses *OpenQASM* syntax within a `defcal` to make a nested gate call. 
  ```c
  defcal ecr $0 $1 {
   // -- snip -- 
   // call another defcal
   x $0;
   // -- snip -- 
  }
  ```
Likewise, there a few examples in the spec that use *OpenQASM* control flow syntax and types within a `defcal`, e.g. 

  ```c
  defcal reset $0 {
     bit res = // measure qubit $0
     if (res == 1) {
        // flip the qubit
     } else {
        // delay for an equivalent amount of time
     }
  }
  ```

There are still more examples of this "superset" behavior in *OpenPulse*:
  - use of `angle`,  `float`, and `duration` types 
  - use of `delay` and `barrier` syntax to align frames 
  - use of `extern` variable syntax to for opaque declarations

By no means should a calibration grammar be **required** to inherit and extend *OpenQASM*, but the examples above show that *OpenPulse* uses it in practice. 

This PR aligns closely with this practicality by importing the *OpenQASM* ANTLR grammar/AST and extending it with additional *OpenPulse* specific constructs. 

# Roadblock 

During code-review and ensuing rebases, we've encountered a few scenarios where a change to *OpenQASM* grammar break the *OpenPulse* implementation such that tests fail. Most of this was brought on during @jakelishman's major refactoring PR (https://github.com/openqasm/openqasm/pull/351) and perhaps is unlikely to re-occur as the language stabilizes. Nonetheless, we wanted to notify the TSC that merging this PR would take on this dependency. 

# Options

@aspcompiler and I see 3 options on how to proceed: 

## 1. Accept the dependency

Accept this dependency and merge this PR once the reviewers comments are resolved. The advantagers and disadantages layed out in the **Context** section.

## 2. Use a delegating parser

@jakelishman has in mind a more general implementation of calibration grammars using a delegated parsing as outlined in this [comment](https://github.com/openqasm/openqasm/pull/352#pullrequestreview-1010546674). 

<ins>Advantages</ins>

This is closer to the ideal way that all defcal grammars are supposed to be implemented and serve as a more useful *reference*. 

<ins>Disadvantages</ins>

- This implementation does not yet exist and it's unclear if it'll make a 3.0 cut. 
- It's not what's done in practice. From offline conversations and community demos, both IBM and AWS seem to pursued a "superset" style approach to compliation of the calibration grammar. This is due to some combination of simplicity and performance. It would be valuable for the community to have a reference that reflects this practice. 
- *OpenPulse* would need to copy over all the *OpenQASM* grammar, parser, and AST files. This creates the same problem as **Option 1** by a different name. Contributor vigilance would now be the way for *OpenPulse* to maintain alignment with the *OpenQASM*. 

## 3. Create a new OpenPulse repository

Invert the dependency by creating a new *OpenPulse* repo under the *OpenQASM* organization i.e. `http://github.com/openqasm/openpulse`. Since *OpenPulse* depends on *OpenQASM*, we could rely on the git submodule to consume a tagged version of *OpenQASM*. If we later choose to use a delegating parser style as presented in *Option 2* it could be done without impacting *OpenQASM* 

<ins>Advantages</ins>

- *OpenPulse* implementations could be pinned to specific OpenQASM revisions, avoiding drift.  
- *OpenQASM* contributors could work on it without the burden of worying about *OpenPulse*.
- *OpenPulse* and *OpenQASM* could progress more asynchronously. This is particularly advantageous given that *OpenPulse* is still experimental and prone to change. 

<ins>Disadvantages</ins>
Creating a new repository under the *OpenQAMS* umbrella is heavy-handed and elevates *OpenPulse* to a level it should not occupy.

source/grammar/openpulseParser.g4
```diff
@@ -0,0 +1,69 @@
...
+    | DURATION
+    | STRETCH
+    | COMPLEX LBRACKET scalarType RBRACKET
+    | WAVEFORM
+    | PORT
+    | FRAME
+    ;
+
+// we extend the statement with the calStatement
+statement:
```
jakelishman, line 36: Is this override still necessary now that `calStatement` is correctly in the OQ 3 ANTLR grammar, or would `statement` pull incorrect definitions of `calStatement` and `defcalStatement` if it isn't overridden?

source/grammar/openpulseParser.g4
```diff
@@ -0,0 +1,69 @@
...
+        | ifStatement
+        | includeStatement
+        | ioDeclarationStatement
+        | measureArrowAssignmentStatement
+        | oldStyleDeclarationStatement
+        | quantumDeclarationStatement
+        | resetStatement
+        | returnStatement
+        | whileStatement
+        | calStatement
```
jakelishman, line 67: Incredibly minor nitpicking, but the rest of the list is alphabetical except for `calStatement`.

source/openpulse/openpulse/__init__.py
```diff
@@ -0,0 +1,26 @@
+"""
+===================================
+OpenPulse Python reference package
+===================================
+
+This package contains the reference abstract syntax tree (AST) for representing
+OpenPulse 3 programs, tools to parse text into this AST, and tools to manipulate
```
jakelishman, line 7: Do we refer to it as "OpenPulse 3" or just "OpenPulse"?

source/openpulse/openpulse/parser.py
jakelishman: It's horribly hacky, but we _can_ get around this without duplicating the code here.  This parser is so coupled to the internal state of the OQ 3 one that if an update would break the hack, it'd break the rest of the parser anyway.

The principle is to monkey-patch the internals of the `openqasm3` ANTLR-generated Python parser to modify the instance checks:

```python
class OpenPulseVersion:
    pass

class FakeMeta(type):
    @classmethod
    def __instancecheck__(cls, instance):
        return isinstance(instance, OpenPulseVersion) or super().__instancecheck__(instance)
    @classmethod
    def __subclasscheck__(cls, other):
        return issubclass(other, OpenPulseVersion) or super().__subclasscheck__(other)

class FakeOpenQASMVersion(openqasm3.antlr.qasmParser.OpenQASMVersion, metaclass=FakeMeta):
    pass

openqasm3.antlr.qasmParser.OpenQASMVersion = FakeOpenQASMVersion
```

source/openpulse/openpulse/printer.py
```diff
@@ -0,0 +1,92 @@
...
+    @_maybe_annotated
+    def visit_DelayInstruction(self, node: ast.DelayInstruction, context: PrinterState) -> None:
+        self._start_line(context)
+        self.stream.write("delay[")
+        self.visit(node.duration, context)
+        self.stream.write("]")
+        if node.qubits:
+            self.stream.write(" ")
+            self._visit_sequence(node.qubits, context, separator=", ")
+        self._end_statement(context)
```
jakelishman, line 92: Is this still broken?

source/openpulse/pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+
+[tool.black]
+line-length = 100
+target-version = ["py36"]
```
jakelishman, line 6: Since we're using `dataclasses`, we must be using Python 3.7 at least.

source/openpulse/tests/test_openqasm_tests.py
jakelishman: Yes - `pytest` has a weird (to me, at least) style where tests that take an argument implicitly look for things in pytest's scope that are decorated as `@pytest.fixture`.  The `example_file` fixture is in `conftest.py`, and it's parametrised to list each example file (except for excludes).

jakelishman (commented):
This is purely (rather minor) comments on technical implementation, since there's more ongoing discussion in the TSC about the parts in Prasahnt's comment above.

source/openpulse/openpulse/parser.py
jakelishman: Just to finish off my thought, because I apparently I pressed submit too soon:

That mutates the actual ANTLR-generated parser (module) to override a context object with a subclass of it, whose entire purpose is to override the behaviour of `isinstance` and `issubclass` so that the new OpenPulse forms type-check as if they were really the OpenQASM types.  Since OpenPulse isn't modifying them from ANTLR and they only appear differently because ANTLR has generated two distinct Python modules, all the functionality is actually identical.

A second alternative is for us to patch the Python output of ANTLR so that the relevant OpenPulse context objects are stubs that directly inherit from the OpenQASM 3 version.  That's perhaps going to be a bit trickier to do, but it's plausible, since we fix to pretty tight versions of the ANTLR runtime.

source/openpulse/openpulse/printer.py
aspcompiler: I just merged latest main into my branch and this is still broken. It will print something like `delay[1000.0ns] ;` with an extra space before semicolon.

jakelishman: Is it just the extraneous space causing you problems? I can squash that pretty immediately.

.github/workflows/tests-ast.yml
```diff
@@ -57,6 +57,23 @@ jobs:
...
+        run: black --check --diff openpulse tests
+
+      - name: Check openpulse style
+        working-directory: source/openpulse
+        run: pylint openpulse tests
+
+      - name: Run openpulse tests
+        working-directory: source/openpulse
+        run: pytest -vv --color=yes tests
+
```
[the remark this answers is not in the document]
aspcompiler: Since we have decided to put OpenPulse reference implementation into a different repo, I will make a new action in the new repo.

source/openpulse/openpulse/printer.py
jakelishman: Fix PR'd in #387 - if you approve, I can merge it and unblock this.

source/grammar/openpulseParser.g4
aspcompiler: Good catch. This is no longer necessary and the entire section can be removed!

source/grammar/openpulseParser.g4
aspcompiler: OK. No need to re-order any more since the entire block is removed.

source/openpulse/openpulse/__init__.py
aspcompiler: Good catch! Fixed.

source/openpulse/pyproject.toml
aspcompiler: I updated the one under openpulse but left the one under oepnqasm intact. That one is also referencing py36.

aspcompiler:
@jakelishman , I implemented most of your small feedbacks. I saw you have already created the [new openpulse repo](https://github.com/openqasm/openpulse-python). I am going to open a new PR to the new repo and worked on the remaining items there:

1. Better way of monkey patching.
2. New github action for openpulse.
3. Parse only the `cal`/`defcal` bodies in OpenPulse parser.

I will reference OpenQASM repo as submodule in the new repo.

PR #354, opened 2022-05-12
Change frame manipulation to use `set`/`shift`/`get` instructions.

psivaraj:
### Summary

This is a minor revision to the [OpenPulse specification](https://qiskit.github.io/openqasm/language/openpulse.html) that changes  the current `phase` and `frequency` attributes access style to an extern/subroutine style for more consistency with the other call patterns in _openpulse_ and _openqasm_ . It also the style used in the reference implementation in #352.

### Details and comments

Succinctly, updates to `phase` and `frequency` are changed as follows

```c
drive_frame.frequency = 5.2e9 -> set_frequency(5.2e9, drive_frame);
drive_frame.frequency -= 0.1e9 -> shift_frequency(-0.1e9, drive_frame);
drive_frame.phase = π/2 -> set_phase(π/2, drive_frame);
drive_frame.phase -= π/4 -> shift_phase(-π/2, drive_frame);
```

Reading the `phase` and  `frequency` is changed as follows
```
float my_freq = drive_frame.frequency; 
↓
float my_freq = get_frequency(drive_frame);

angle my_phase = drive_frame.phase;
↓
angle my_phase = get_phase(drive_frame);
```

This also allows for a more consistent way for vendors to extend frame manipulation via `extern`s e.g. 
```c
extern chirp_frequency(float freq, frame fr);
extern swap_phases(frame fr1, frame fr2);
```

cc: @aspcompiler

### DIFF
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -189,7 +189,7 @@ of type ``float``.
   get_frequency(frame fr) -> float;
 
 Changing the frequency or phase behaves as an instantaneous operation (ie., its
-duration is zero device ticks) at the current time point of the frame. If a vendor If a vendor
+duration is zero device ticks) at the current time point of the frame. If a vendor
 is unable to support such instantaneous operations, it is expected that the
 compiler shall raise a compile-time error when encountering such frame manipulations.
 
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -156,12 +156,45 @@ NCO in analogy to virtual to physical register allocation.
 Frame Manipulation
 ~~~~~~~~~~~~~~~~~~
 
-The (frequency, phase) tuple of a frame can be manipulated throughout program
-by referencing ``.frequency``, and ``.phase``, with updates applied immediately
-at the current time of the frame. Operations must be appropriate for the respective type,
-``float`` for frequency and ``angle`` for phase. Again, the exact precision of these calculations
-is hardware specific. If either the frequency or phase are set to values that are invalid for
-the hardware, the compiler shall raise a compile-time error.
+The ``phase`` and ``frequency`` states of a frame can be manipulated throughout the program
+by using ``set`` and ``shift`` instructions and read using a ``get`` instruction. In particular,
+the `set_phase` and `shift_phase` instructions allow one to supply the frame and a value of type
+``angle`` representing the amount by which to set/shift the phase.
+
+.. code-block:: javascript
+
+  set_phase(angle phase, frame fr);
+  shift_frequency(angle phase, frame fr);
+
+The `get_phase` instruction allows one to supply the frame from which to retrieve the phase of
+type ``angle``.
+
+.. code-block:: javascript
+
+  get_phase(frame fr) -> angle;
+
+Analogously, the `set_frequency` and `shift_frequency` instructions allow one to supply the frame
+and a value of type ``float`` representing the amount by which to set/shift the frequency.
+
+.. code-block:: javascript
+
+  set_frequency(float freq, frame fr);
+  shift_frequency(float freq, frame fr);
+
+The `get_frequency` instruction allows one to supply the frame from which to retrieve the frequency
+of type ``float``.
+
+.. code-block:: javascript
+
+  get_frequency(frame fr) -> float;
+
+Changing the frequency or phase is an instantaneous operation. If a vendor
+is unable to support such instantaneous operations, it is expected that the
+compiler shall raise a compile-time error when encountering such frame manipulations.
+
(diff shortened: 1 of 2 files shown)


source/language/openpulse.rst
```diff
@@ -156,12 +156,45 @@ NCO in analogy to virtual to physical register allocation.
...
+  shift_frequency(float freq, frame fr);
+
+The `get_frequency` instruction allows one to supply the frame from which to retrieve the frequency
+of type ``float``.
+
+.. code-block:: javascript
+
+  get_frequency(frame fr) -> float;
+
+Changing the frequency or phase is an instantaneous operation. If a vendor
```
taalexander, line 191:
```suggestion
Changing the frequency or phase behaves as an instantaneous operation (ie., its duration is zero device ticks) at the current time point of the frame. If a vendor
```

taalexander (changes requested):

source/language/openpulse.rst
```diff
@@ -156,12 +156,45 @@ NCO in analogy to virtual to physical register allocation.
...
+is unable to support such instantaneous operations, it is expected that the
+compiler shall raise a compile-time error when encountering such frame manipulations.
+
+Moreover, the exact precision of these calculations is hardware specific. If either the frequency
+or phase are set to values that are invalid for the hardware, the compiler shall raise a
```
blakejohnson, line 196: I think I am interpreting this statement about frequencies to refer to a valid frequency range (as opposed to a frequency precision). What does this mean for phases?

blakejohnson (changes requested):
Looks good. I just have the one question about the new statement about setting phase values that are invalid for the hardware. I assume that some implementors will make a float -> fixed conversion when setting frequency or phase, in which case I think we need to call out what we mean by an invalid float value.

source/language/openpulse.rst
```diff
@@ -156,12 +156,45 @@ NCO in analogy to virtual to physical register allocation.
...
-the hardware, the compiler shall raise a compile-time error.
+The ``phase`` and ``frequency`` states of a frame can be manipulated throughout the program
+by using ``set`` and ``shift`` instructions and read using a ``get`` instruction. In particular,
+the `set_phase` and `shift_phase` instructions allow one to supply the frame and a value of type
+``angle`` representing the amount by which to set/shift the phase.
+
+.. code-block:: javascript
+
+  set_phase(angle phase, frame fr);
+  shift_frequency(angle phase, frame fr);
```
stevenheidel, line 167:
```suggestion
  shift_phase(angle phase, frame fr);
```

stevenheidel:
Decision was made at the TSC meeting today to put the frame first in the set/shift calls. So the updated syntax should be:

```
drive_frame.frequency = 5.2e9 -> set_frequency(drive_frame, 5.2e9);
drive_frame.frequency -= 0.1e9 -> shift_frequency(drive_frame, -0.1e9);
drive_frame.phase = π/2 -> set_phase(drive_frame, π/2);
drive_frame.phase -= π/4 -> shift_phase(drive_frame, -π/2);
```

source/language/openpulse.rst
psivaraj: Good point. Upon revision, I don't think it's meaningful to discuss phase because in regards to precision we already say in the spec

> `angle name;` declares an angle with machine-architecture-specified width

Meanwhile, there is no meaningful "out of range" for phase. I've updated the text to only discuss frequency. Let me know what you think.

psivaraj:
Changed order of operations to put the frame first in the set/shift calls.

stevenheidel (approved):

source/language/openpulse.rst
```diff
@@ -156,12 +156,46 @@ NCO in analogy to virtual to physical register allocation.
...
+
+The `get_frequency` instruction allows one to supply the frame from which to retrieve the frequency
+of type ``float``.
+
+.. code-block:: javascript
+
+  get_frequency(frame fr) -> float;
+
+Changing the frequency or phase behaves as an instantaneous operation (ie., its
+duration is zero device ticks) at the current time point of the frame. If a vendor If a vendor
```
blakejohnson, line 192: Accidental repetition of "If a vendor".
```suggestion
duration is zero device ticks) at the current time point of the frame. If a vendor
```

blakejohnson (approved):
One minor typo, otherwise LGTM

source/language/openpulse.rst
blakejohnson: The new text looks good.

taalexander (approved):
LGTM

blakejohnson (approved):

PR #358, opened 2022-05-24
Restrict complex types to floating-point

jakelishman:
### Summary

This more completely defines the semantics for working with complex
types in OpenQASM 3.0, including restricting them to only complex
values.  We would have needed special-case rules to define things like
division between `complex[int]` types, and the hardware support and
targetted use cases did not seem to be strong enough to spend time doing
so for the first version of the language.  The promotion rules may also
have been unusual.

### Details and comments

This is the recommendation of the types-and-casting working group (unless any of them comment saying I've made a mistake).  We discussed use-cases for `complex[int]`, mostly deciding that the primary use would be some potential interoperability with OpenPulse. This was seen as coming with a high implementation cost, and requiring a large amount of time and space to fully define all the semantics and promotion rules, and the potential use case not sufficient to merit inclusion in the OQ 3.0 spec.

We leave the `complex[float[size]]` syntax in place to make it more convenient if individual vendors want to implement their own non-floating-point semantics, but we make it clear in the spec that these are not part of the core language for 3.0, and we would reserve the right to define our own semantics in a later language version.


source/language/types.rst
```diff
@@ -210,24 +210,51 @@ compatible with run-time values on some platforms.
...
-``type`` (``int``, ``float``, ``angle`` plus their optional size). The real
-and imaginary parts of the complex number are ``type[size]`` types. For instance, ``complex[float[32]] c``
-would declare a complex number with real and imaginary parts that are 32-bit floating point numbers. The
-``im`` keyword defines the imaginary number :math:`sqrt(-1)`. ``complex[type[size]]`` types are initalized as
-``a + b im``, where ``a`` and ``b`` must be of the same type as ``type[size]``. ``b`` must occur to the
-left of ``im`` and the two can only be seperated by spaces/tabs (or nothing at all).
+Complex numbers may be declared as ``complex[float[size]] name``, where ``size``
+is the size of the IEEE-754 floating-point number used to store the real and
+imaginary components.  Each component behaves as a ``float[size]`` type.  The
+designator ``[size]`` can be omitted to use the default hardware ``float``.
```
stevenheidel, line 220: Should we go further and say that if `float` is omitted then it is assumed? ie. `complex` == `complex[float]`

blakejohnson: I am okay with that.

blakejohnson (dismissed):
I'm okay with this as is. @stevenheidel's suggestion seems like a good additional refinement.

source/language/types.rst
jakelishman: We'd originally left this in deliberately as a nod towards Blake and Prasahnt's potential extension in a later language spec.  I'm happy to change it if Blake's happy.

Changed in b3caee1.

blakejohnson (dismissed):

source/language/types.rst
```diff
@@ -210,24 +210,52 @@ compatible with run-time values on some platforms.
...
-and imaginary parts of the complex number are ``type[size]`` types. For instance, ``complex[float[32]] c``
-would declare a complex number with real and imaginary parts that are 32-bit floating point numbers. The
-``im`` keyword defines the imaginary number :math:`sqrt(-1)`. ``complex[type[size]]`` types are initalized as
-``a + b im``, where ``a`` and ``b`` must be of the same type as ``type[size]``. ``b`` must occur to the
-left of ``im`` and the two can only be seperated by spaces/tabs (or nothing at all).
+Complex numbers may be declared as ``complex[float[size]] name``, where ``size``
+is the size of the IEEE-754 floating-point number used to store the real and
+imaginary components.  Each component behaves as a ``float[size]`` type.  The
+designator ``[size]`` can be omitted to use the default hardware ``float``, and
+``complex`` with no arguments is a synonym for ``complex[float]``.
```
psivaraj, line 221: > ``complex`` with no arguments is a synonym for ``complex[float]``.

Will allowing this now make it  awkward if we later decide to support non-``float`` component types?

psivaraj: Oh, I see the discussion above that the consensus is to add it.

psivaraj (approved):
One question but otherwise 👍.

blakejohnson (approved):

PR #360, opened 2022-05-26
Remove vim plugin from mono repository

jakelishman:
### Summary

Now that we have the openqasm GitHub repository, we can move the vim
plugin (and any potential future plugins) into its own repository.  This
now exists as openqasm/openqasm.vim, and the complete git history of the
`plugins/vim` subdirectory is extracted into that repository (including
re-linking the issues numbers to the correct location).

### Details and comments

I've put the plugin at https://github.com/openqasm/openqasm.vim, and done a little bit of git trickery to maintain only the relevant parts of the commit history for that project.  This PR removes the (now) duplicate files from this repository.


jwoehr (approved):
This seems purely procedural, @jakelishman ... shall we just merge?

PR #362, opened 2022-05-30
Add basic formatter to reference AST

jakelishman:
### Summary

This adds a basic print visitor to the reference AST, which can be used
as a (very simple) code formatter.  It outputs valid OpenQASM 3.0, as
opposed to the AST tree format that is automatically provided by the
`dataclasses` structure (e.g. call `repr` on a node).

The structure of the printer is very similar to the print visitor in
Qiskit Terra (which I also wrote), which is a simple single-pass printer
into the stream.  This is perhaps the simplest way of writing the
printer, but it doesn't leave much room for extensibility on look-ahead
(e.g. breaking lines in good places or dynamically choosing the spacing
around binary operators based on the precedence of pairs of operands).
It could be extended to handle these, but a general-purpose,
configurable formatter might be a better fit in a stand-alone project.

### Details and comments

I think this is a very useful utility to package along with the reference AST, but would also be happy to package it up as a separate Python package.  We'd be interested in using this in Qiskit assuming we use the reference AST as our Python-bound AST in the longer term (right now we have a quick and dirty one that dates from before the reference implementation here).


jakelishman:
Note I'll need to update this a little assuming #351 merges before it (which it should), but the changes are minor.

blakejohnson:
This does not need TSC review. So, as soon as @aspcompiler is good, go ahead and merge.

source/openqasm/openqasm3/printer.py
```diff
@@ -0,0 +1,700 @@
...
+
+from typing import Sequence, Optional
+
+from . import ast, properties
+from .visitor import QASMVisitor
+
+__all__ = ("dump", "dumps")
+
+
+def dump(node: ast.QASMNode, file: io.TextIOBase, **kwargs):
```
aspcompiler, line 13: A Docstring would be nice since this is an exported function.

source/openqasm/openqasm3/printer.py
```diff
@@ -0,0 +1,700 @@
...
+    _Printer(file, **kwargs).visit(node)
+
+
+def dumps(node: ast.QASMNode, **kwargs):
```
aspcompiler, line 17: Docstring.

source/openqasm/openqasm3/printer.py
```diff
@@ -0,0 +1,700 @@
...
+    out = io.StringIO()
+    _Printer(out, **kwargs).visit(node)
```
aspcompiler, line 19: Optional suggestion:
```suggestion
    dump(node, out, **kwargs)
```
This will allow the tests to cover `dump()`. Also, if we ever need to change the interface of `_Printer`, we don't have to change it multiple places.

aspcompiler (approved):
Just a few minor suggestions. Otherwise LGTM.

source/openqasm/openqasm3/printer.py
jakelishman: Whoops, I forgot to write this.  Thanks for pointing it out - I've done the documentation completely in ca19088.

source/openqasm/openqasm3/printer.py
jakelishman: Done in ca19088.

source/openqasm/openqasm3/printer.py
jakelishman: Done in 09bbb86.

jakelishman:
@aspcompiler: I've updated this with the necessary changes on top of #351, and with your comments (thanks!).

aspcompiler (approved):

PR #363, opened 2022-06-01
Clarify bit representation and motivation for `angle` type

jakelishman:
### Summary

This expands the spec with more detailed descriptions of the effective
bit representations of the `angle` type, and expands on the reasoning
for the introduction of the type, and how various mathematical
operations apply on it.  This is meant as a clarification of the
specification; it is attempting to describe the same behaviour as
before, but in more precise terms.

It is worth mentioning that hardware is not _compelled_ to implement the
angle type in a hardware register that stores these exact bit patterns
in memory.  It is only required that the logical interactions with the
``bit`` and other types, and bit-indexing operations return the same
values as described by the specification.

### Details and comments

This PR has existed and staled for a little while on my own fork as jakelishman/openqasm#2.  There is some internal types-and-casting group discussion still extant on that PR, though I've attempted to copy across the comments that I believe are still not entirely addressed here for visibility.


jakelishman (commented):
I've copied across the comments from jakelishman/openqasm#2 that I think might not be fully resolved / are important for visibility.  If anybody with a comment on the other PR wants to bring it here as well because they don't think I'd fully resolved them, please feel free.

source/language/classical.rst
```diff
@@ -39,12 +41,13 @@ assigned variable on the LHS.
...
-integer, and the corresponding assignment operators. The shift operators
-shift bits off the end. They also support not ``~``, ``popcount`` [1]_, and left and
-right circular shift, ``rotl`` and ``rotr``, respectively.
+Classical registers, ``bit``, ``uint`` and ``angle`` support bitwise operators
+and the corresponding assignment operators with registers of the same size:
+and ``&``, or ``|``, xor ``^``. They support left shift ``<<`` and right shift
+``>>`` by an unsigned integer, and the corresponding assignment operators. The
+shift operators shift bits off the end. They also support bitwise negation ``~``,
+``popcount`` [1]_, and left and right circular shift, ``rotl`` and ``rotr``,
+respectively.
```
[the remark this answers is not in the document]
blakejohnson: It seems to me that these operations are still well-defined even if the width of these operands are not specified in the source file. The compiler will interpret them as having specific widths in the context of a particular target.

blakejohnson (approved):
LGTM

source/language/classical.rst
```diff
@@ -56,6 +59,17 @@ right circular shift, ``rotl`` and ``rotr``, respectively.
    a | b; // Produces "11111111"
    a & b; // Produces "00000000"
 
+For ``uint`` and ``angle``, the results of these operations are defined as if
+the operations were applied to their defined bit representations::
+
+   angle[4] a = 9 * (pi / 8);  // "1001"
+   a << 2;  // Produces pi/2, which is "0100"
+   a >> 2;  // Produces pi/4, which is "0010"
```
psivaraj, line 67: **suggestion**: Might be clearer to include an assignment e.g.
```c
angle[4] b = a << 2;  // Produces pi/2, which is "0100"
angle[4] c = a >> 2;  // Produces pi/4, which is "0010"
```

source/language/classical.rst
```diff
@@ -79,6 +93,13 @@ an index set, for example ``i in {0,3}`` returns ``true`` if i equals 0 or 3 and
    // Susceptible to floating point casting errors
    e == float(d);
 
+.. note::
+
+   Angles are naturally defined on a ring, and so comparisons may not work
+   exactly how you might expect.  For example, ``2 * ang >= ang`` is not
+   necessarily true; if ``ang`` represents the angle :math:`3\pi/2`, then
+   ``2*ang == pi`` and ``pi < ang``.
```
psivaraj, line 101: **question**: It might not be intuitive, but is it always well-defined? If so, it might be worth illustrating with concrete examples. If not, I think we should explicitly state it is not well-defined.

source/language/classical.rst
```diff
@@ -110,6 +179,12 @@ multiplication, division, and power and the corresponding assignment operators.
    angle[10] c;
    c = angle(a + b); // cast to angle[10]
 
+.. note::
+
+   Real hardware may well not have access to floating-point operations at
+   runtime.  OpenQASM 3 compilers may reject programs that require runtime
+   operations on these values if the target backend does not support them.
```
psivaraj, line 186: :100:

source/language/classical.rst
jakelishman: Would me putting the exact type (`angle[4]`) be enough for you?  I'd wanted to avoid explicitly writing an assignment, because that introduces the possibility of an explicit cast, whereas here I wanted to be super clear we're talking about the type and value of the expression.

source/language/classical.rst
jakelishman: I am open to forbidding the ordering comparisons of `angle` (i.e. disallowing `<`, `<=`, `>` and `>=`, but permitting `==` and `!=`).  I think we need more input from more people on that, if so.

If that's not what we decide, do you have a suggestion to make it clearer here (I'm totally open to that)?  It is always well defined (the semantics are very nearly the same as same operation between two `uint` with the same bit patterns), and I'm hoping the rest of the text of this PR (after review) will make that clear.

I was hoping the example in the second sentence would illustrate the "it's not always intuitive" view, but I'm happy to add to it.

source/language/classical.rst
steleman: > angle[4] b = a << 2;  // Produces pi/2, which is "0100"
> angle[4] c = a >> 2;  // Produces pi/4, which is "0010"

Left shift and right shift are the same exact arithmetic operation - in this case, division?
Left shifting an angle by 2 yields `pi` / 2, but right shifting an angle by 2 yields `pi` / 4?

jakelishman: Yeah exactly, the intent is that bit-shifting is defined the same way as it is for a `uint` which has the same bit pattern as `angle` (whether simulated or direct).  So bit-shift right by `n` is the same as integer division by `2 ** n`, and bit-shift left by `n` is the same as integer division by `2 ** (-n)`.

(Just for ease of discussion, since the value of `a` has disappeared from the context on GitHub: it's defined as `angle[4] a = 9 * (pi / 8);` which has the effective bit pattern `"1001"`.)

psivaraj: > Would me putting the exact type (angle[4]) be enough for you? I'd wanted to avoid explicitly writing an assignment, because that introduces the possibility of an explicit cast, whereas here I wanted to be super clear we're talking about the type and value of the expression.

Ah got it, I think it's fine as it is then.

source/language/classical.rst
jakelishman: Yeah, for sure it's the same thing with two `uint`.  I think with `uint`, though, people who want to use the comparison operators are generally working with numbers that definitely won't overflow.  With `angle` here, the overflow will get hit all the time.

I've added an extra little bit in 986fade.

blakejohnson (approved):

PR #364, opened 2022-06-06
Add support for new pragmas and annotations

jakelishman:
### Summary

This includes new tokenisation modes to support the "to end of line"
behaviour of the new pragmas and annotations.  The AST is updated to
support attaching arbitrary annotations to any statement.

### Details and comments

@jheckey: this is unchanged since I showed you the diff last week, I've just rebased it on top of the main branch.


source/openqasm/openqasm3/printer.py
```diff
@@ -776,13 +815,18 @@ def visit_ClassicalAssignment(
...
+        if node.command is not None:
+            self.stream.write(" ")
+            self.stream.write(node.command)
         self._end_line(context)
-        with context.increase_scope():
-            for statement in node.statements:
-                self.visit(statement, context)
+
+    @_maybe_annotated
+    def visit_Pragma(self, node: ast.Pragma, context: PrinterState) -> None:
```
jheckey, line 828: Interesting corner that I hadn't considered - annotated pragmas.

jheckey (commented):
LGTM.

source/openqasm/openqasm3/printer.py
jakelishman: I'm happy to update this PR to either allow or permit this.  At the moment, `pragma` is in `statement` in the ANTLR grammar, which automatically means it can be annotated, but I could also re-arrange that rule to forbid this, if your working group would prefer.

jakelishman: I've updated this so that `ast.Pragma` can no longer take annotations and updated the parser to make `pragma` a global-only statement.  I didn't make any changes to the spec of the text - I don't know if you'd also like to do that in a separate PR.

jheckey (approved):

levbishop (approved):

jakelishman:
Tests are failing because the for-loop PR merged, and this will just need its own for-loop examples updating.  I'll do that.

jakelishman:
This is updated and ready for AWS to check over the Python changes.

PR #366, opened 2022-06-14
Flesh out compile-time expressions and `const` types

jakelishman:
### Summary

Formalise rules for compile-time `const` expressions, and change array access
modifier to `readonly` from `const`.

These updates are generally expected to be consistent with what
implementations already do, but formalised completely into the language
specification.

This also formalises the idea of the expressions that are guaranteed to
be evaluated at compile-time, and thus which expressions are guaranteed
to be valid to use as widths in type declarations, and in qubit
dereferencing.  This change does not attempt to solve the issue of the
allowed "dynamic dereferencing" of qubits, but fully defines the
problem: a reference to a qubit is dynamic if the type of the indexing
value is not `const`.  Implementations of OpenQASM 3.0 are permitted to
treat this as an error.

This change also fleshes out the section on built-in "scientific
calculator" functions on constant expressions, adding specific rules for
how the particular overload is chosen.

The change of the array access modifier in subroutines is to avoid unnecessary,
misleading conflict with the type modifier `const` that generally means
"compile-time constant".

### Details and comments

This is in draft for initial comments from the types-and-casting WG first (since this is my attempt to formalise in writing what we've discussed and agreed upon).  The majority of the `const` expression rules here should be largely consistent with what current implementations are already doing in terms of compile-time constant folding, but formalises the rules, to give users a well-defined base point.

This PR _does not include_ two potential extensions to this scheme that have also been discussed:
- `const` type for `for`-loop variables, requiring the loop to be unrolled at compile time;
- `const def` functions, for user-defined compile-time macros.

I will write up the first of these (`for`-loop unrolling if the loop variable has `const` type) in a separate PR on top of this as a potential discussion topic.  That may well not be suitable for the 3.0 release, however.

We have no pressing need for user-defined compile-time macros in the 3.0 spec.

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -65,11 +65,44 @@ declares a reference to a quantum bit. These qubits are referred
 to as "virtual qubits" (in distinction to "physical qubits" on
 actual hardware; see below). The statement ``qubit[size] name;``
 declares a quantum register with ``size`` qubits.
-Sizes must always be constant positive integers. The label ``name[j]``
-refers to a qubit of this register, where
-:math:`j\in \{0,1,\dots,\mathrm{size}(\mathrm{name})-1\}` is an integer.
+Sizes must always be :ref:`compile-time constant <const-expression>` positive
+integers.
 Quantum registers are static arrays of qubits
 that cannot be dynamically resized.
+
+The label ``name[j]`` refers to a qubit of this register, where
+:math:`j\in \{0,1,\dots,\mathrm{size}(\mathrm{name})-1\}` is an integer.
+
+.. note::
+
+   To be compliant with the base OpenQASM 3.0 specification, an implementation
+   is only required to allow this "quantum-register indexing" with a
+   :ref:`compile-time constant value <const-expression>` (those with ``const``
+   types).  Implementations are permitted to treat indexing into a quantum
+   register with a value of non-\ ``const`` type as an error.  Consult your
+   compiler and hardware documentation for details.
+
+.. code-block::
+
+   // Valid statements
+
+   include "stdgates.inc";
+
+   qubit[5] q1;
+   const uint SIZE = 4;
+   uint runtime_u = 2;
+   qubit[SIZE] q2;  // Declare a 4-qubit register.
+
+   x q1[0];
+   z q2[SIZE - 2];  // The index operand is of type `const uint`.
+
+
+   // Validity is implementation-defined.
+
+   x q1[runtime_u];
+   // Indexing with a value with a non-`const` type (`uint`, in this case) is
+   // not guaranteed to be supported.
+
 
 The keyword ``qreg`` is included
 for backwards compatibility and will be removed in the future.
@@ -103,8 +136,8 @@ circuits. Physical qubits must not be declared and they are, as all the qubits, 
    // CNOT gate between physical qubits 0 and 1
    CX $0, $1;
 
-Classical types
----------------
+Classical scalar types
+----------------------
 
 Classical bits and registers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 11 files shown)


jakelishman:
(This is marked "draft" just because it's in the first stages of review with the types-and-casting group first.  I'll move it out of draft and tag for TSC once the internal review is complete.)

source/language/types.rst
```diff
@@ -271,64 +304,173 @@ be true and 0 will be false.
...
+   const int[16] i2 = int[16](runtime_u);  // Casting runtime values is not `const`.
+
+
+.. _const-expression-nonconst:
+
+The resultant type of any expression involving a value that is not ``const`` is
+not ``const``.  The output type of a call to a subroutine defined by a ``def``,
+or a call to a subroutine linked by an ``extern`` statement is not ``const``,
+although compilers are permitted to elide these calls at run-time if they can
+prove that doing so will not change the result of the program.  In these cases,
```
psivaraj, line 413: **question**: I don't quite understand this clause. Mind explaining further?

jakelishman: It might be better to remove that clause.  I'd meant that given
```qasm
def f(int a) -> int {
    return a;
}
```
a compiler would be allowed to transform `f(5)` into `5`, even through the non-`const` `def`.  When I wrote it, I was worried that the previous statement implied that that sort of classical optimisation (inlining, in this case) was forbidden.  Reading it again, I think that was an unfounded fear, and text about allowed classical transformations would go in a different part of the spec if we _were_ going to write about them anyway.

mhillenbrand: I also read the language in this PR as not introducing any constraints on the optimizations a compiler may perform. So, removing this clause would be fine with me.

mbhealy: I think leaving this clause in is reasonable.

jakelishman: I think the statement itself is sound, but it's probably best to put it somewhere else - I think it fits better in a (currently non-existent) section about what classical optimisations and transformations compilers are permitted to make to the source code.  "Provably does not change the outcome of the program" is (I think) a pretty uncontroversial starting point, although there's definitely some nuance around not changing the outcome of a program that involves quantum randomness!

jakelishman:
I've just rebased this over `main`, due to minor code-only conflicts caused by the merge of #370.  The text is identical to before the rebase, and is still waiting for TSC review.

source/language/types.rst
```diff
@@ -368,54 +510,112 @@ their argument expression in parentheses:
...
+      | pow      | ``int``, ``uint``                 | ``int``                              | :math:`\texttt{pow(a, b)} = a^b`.      |
       |          |                                   |                                      |                                        |
       |          | ``float``, ``float``              | ``float``                            | For floating-point and complex values, |
       |          |                                   |                                      | the principal value is returned.       |
       |          | ``complex``, ``complex``          | ``complex``                          |                                        |
       +----------+-----------------------------------+--------------------------------------+----------------------------------------+
-      | rotl     | ``bit[n]``, (``int`` or ``uint``) | ``bit[n]``                           |                                        |
+      | rotl     | ``bit[n]``, ``int``               | ``bit[n]``                           | Rotate the bits in the representation  |
```
blakejohnson, line 545: why drop `uint`?

source/language/types.rst
```diff
@@ -368,54 +510,112 @@ their argument expression in parentheses:
...
+      |          |                                   |                                      | rather than being set to zero.  The    |
+      |          |                                   |                                      | width of the output is set equal to    |
+      |          |                                   |                                      | the width of the input.                |
+      |          |                                   |                                      |                                        |
+      |          |                                   |                                      | ``rotl(a, n) == rotr(a, -n)``.         |
       +----------+-----------------------------------+--------------------------------------+----------------------------------------+
-      | rotr     | ``bit[n]``, (``int`` or ``uint``) | ``bit[n]``                           |                                        |
+      | rotr     | ``bit[n]``, ``int``               | ``bit[n]``                           | Rotate the bits in the representation  |
```
blakejohnson, line 556: Same question as I asked for `rotl`.

source/language/types.rst
```diff
@@ -368,54 +510,112 @@ their argument expression in parentheses:
...
+
+   const float[64] f4 = pow(i1, -2);
+   // Value 0.0625, expression has type `const float`.  The second,
+   // `(float, float) -> float`, overload is chosen, because `-2` (type
+   // `const int`) cannot be implicitly promoted to `const uint`, but both
+   // input types can be implicitly promoted to `float`.  The `complex` overload
+   // is not attempted, because it has lower priority.
+
+   const bit[8] b2 = rotl(b1, 3);
+   // Value "0101_0001", expression has type `const bit[8]` (the
```
blakejohnson, line 603: Maybe also add an example for what happens with a negative integer?

blakejohnson (commented):
Looks good. I just have a question about `rotr` and `rotl` on signed types.

blakejohnson:
@mbhealy volunteered to remove the `int`/`uint` change to the rotation operations.

source/language/types.rst
jakelishman: This table was written before the new implicit casting rules. `uint` autopromotes to `int` now if it's in range, and since these are unsized compile-time calls, that conversion will always succeed. I think explicitly saying "`int` or `uint`" here means we have to specify an order that the overloads are tried, but there wouldn't actually be any difference in them, so it seemed simpler to remove it and just use the implicit cast.

(I must have forgotten I did this, or I would have written it in the commit message.)

kevinhartman:
Should we update `classical.rst` as well if changing the supported operand types of `rotl` and `rotr`?

(FWIW, I'm not sure I like the idea of `uint` promoting to an `int` automatically, but that sounds like a sailed ship, and I don't have much context anyway.)

mbhealy:
> @mbhealy volunteered to remove the `int`/`uint` change to the rotation operations.

After looking into this more deeply, and after a short discussion in last week's meeting, there was some agreement among attendees that the idea of rotating an int value seems a bit odd. I think limiting rotations to bit and uint types only would make the most sense. Rotation on ints (with stationary or mobile sign bits) is still possible, but doesn't seem like something that needs direct language support/convenience features.

source/language/types.rst
mbhealy: Updated to uint and bit only in https://github.com/openqasm/openqasm/pull/366/commits/307fa094e8be792a845623101323999a452ab261

source/language/types.rst
mbhealy: Updated to uint and bit only in https://github.com/openqasm/openqasm/pull/366/commits/307fa094e8be792a845623101323999a452ab261

source/language/types.rst
mbhealy: No longer possible after https://github.com/openqasm/openqasm/pull/366/commits/307fa094e8be792a845623101323999a452ab261

mbhealy:
> @mbhealy volunteered to remove the `int`/`uint` change to the rotation operations.

Updated to uint and bit only in https://github.com/openqasm/openqasm/pull/366/commits/307fa094e8be792a845623101323999a452ab261

blakejohnson (dismissed):

jakelishman:
I've just pushed a5ab1f7 that includes the changes discussed in the working group meeting today about `rotl` and `rotr`, namely:
- there are two overloads for each: `(bit[n] target, int distance) -> bit[n]` and `(uint[n] target, int distance) -> uint[n]` for all compile-time known `n`.
- the `distance` argument (number of places to rotate) can be signed, and a negative number is equivalent to calling the other function with a positive number (i.e. `rotl(a, -4)` is the same as `rotr(a, 4)`).
- the `target` argument only makes sense as `bit[n]` or `uint[n]`.  It was never meant to be possible to rotate signed integers, and I think leaving the names out of the parameter specification caused some confusion here.

blakejohnson (approved):

Issue #367, opened 2022-06-17
OpenPulse-specific syntax present in OpenQASM 3.0 contexts


jakelishman:
### Summary

There are some places in our documentation where OpenPulse-specific constructs are used outside of `cal` and `defcal` blocks.  We either need to:
- change these examples to put those parts inside `cal` blocks
- update the text of the spec to allow calibration-grammar constructs to be present in additional situations.

This is so in the future, if a new `defcalgrammar` is developed outside of the OpenQASM 3 effort directly, we're better defining where its additional grammar constructs are actually valid.  I think it's most likely that the examples probably just want putting inside `cal` blocks, for the most part.

### Details

On the OpenPulse pages of the specification, there are some cases where OpenPulse types are used outside of `cal` or `defcal` blocks.  For example, one of the [Capture Instruction](https://openqasm.com/language/openpulse.html#capture-instruction) examples says:

```qasm
// Use a boxcar function to generate IQ data from raw waveform
extern boxcar(waveform input) -> complex[float[64]];
// Use a linear discriminator to generate bits from IQ data
extern discriminate(complex[float[64]] iq) -> bit;

cal {
    // Define the ports
    extern port m0;
    extern port cap0;
}

// ... and so on ...
```

The first `extern` uses the `waveform` type as an input, which isn't a valid type in OpenQASM 3 (it's in OpenPulse).  I suspect the `extern` functions just want moving into the `cal` block - this is how the [calibrations in practice](https://openqasm.com/language/pulses.html#calibrations-in-practice) section presents such examples.

The [page on general pulse grammars](https://openqasm.com/language/pulses.html) also has wildcard physical qubits matched as `$<identifier>` (e.g. `$q` instead of `q`).  That seems a bit inconsistent to me (why not use `q`? The `$` usually represents a specific hardware qubit, but that's not the case here - it's a stand-in for any qubit.), but not really an issue.

jakelishman:
The consensus from the TSC meeting on 2022-06-17 was that the `extern` function definitions I mentioned here ought to be within `cal` blocks.  There is tentative agreement to change the wildcard identifier in `defcal` statements from `$<identifier>` to simply `<identifier>` on the basis of consistency with the rest of the language.

I will make two separate PRs:
- add the `cal` statement to the ANTLR reference parser (and update `defcal` statements to correctly tokenise the unknown grammars, since they coincide)
- change the text of the spec to drop the initial `$` from wildcard identifiers (but not general hardware qubits)

[...]

PR #371, opened 2022-06-20
Drop dollar sign from pulse wildcard identifier

jakelishman:
### Summary

This changes the wildcard identifier pattern from `$<identifier>` to
just being the identifier (for example `$q` to `q`).  Precise hardware
qubits such as `$0` are not affected.  The extra dollar sign did not
carry any information, since the wildcard applied to _all_ qubits, and
caused a strange discrepancy between what was a valid identifier in
different situations.

### Details and comments

This is as discussed in the TSC meeting on 2022-06-17, though we also recognised that there was originally a reason this `$` was present, and people might have remembered stronger opposition to making this change since that meeting.

Fixes a part of #367.

### DIFF
--- a/examples/defcal.qasm
+++ b/examples/defcal.qasm
@@ -8,8 +8,8 @@ defcal x $1 {
   play drive($1), gaussian(...);
 }
 
-defcal rz(angle[20]:theta) $q {
-  shift_phase drive($q), -theta;
+defcal rz(angle[20]:theta) q {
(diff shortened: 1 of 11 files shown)


blakejohnson (approved):
Changes to the language of the spec look fine.

levbishop (approved):
Looks good, I'd like to get a sign-off from the OpenPulse implementers

taalexander (approved):

Issue #372, opened 2022-06-20
Example `pong.qasm` contains invalid OpenQASM 3


jakelishman:
The example file `pong.qasm` looks like it may have been left behind after several iterations of updates to the specification, and I'm not at all sure what it's meant to be doing.  This currently passes the tests because of our broken `defcal` handling in the reference ANTLR grammar lets the parser eat the first `defcalStatement` match all the way to the last brace of the file (even though it should have ended much sooner), since this is the only way ANTLR can build a valid parse tree.

My immediate concerns for OpenQASM 3 are for lines 24 to 28: https://github.com/openqasm/openqasm/blob/bb35c56cc652255115d6500de1a6642f3bb2714f/examples/pong.qasm#L24-L28

The `glue[1]-delay[0]` construct is unfamiliar to me - certainly it's the only instance of the word `glue` in the entire repository, including historic commits.  These lines were introduced in #58, and haven't been relevantly changed since.  Line 28 also appears to use a `defcal`'d gate call as an expression item, which I'm also not familiar with.

The OpenPulse sections also look invalid/out-of-date to me, but I don't know OpenPulse well enough to talk about that.  They are https://github.com/openqasm/openqasm/blob/bb35c56cc652255115d6500de1a6642f3bb2714f/examples/pong.qasm#L8-L21

This example needs rewriting into valid OpenQASM 3, or removing entirely.  This issue is blocking a fix to the ANTLR grammar for calibration blocks, because fixing that unmasks the errors in this file and causes test failures.

jakelishman:
`t1.qasm` and `defcal.qasm` also have invalid OpenQASM 3 - the former is missing types on its `for` loop variables, and the latter has an extraneous colon in a defcal argument list - but these are trivial fixes.  Both are masked in the current tests by the same faulty behaviour of the reference ANTLR grammar.

I have a fix for the ANTLR grammar written and ready to deploy, but I can't make it yet because the tests are failing.

ajavadia:
glue is what we used informally before `stretch`, so this is definitely left-over from an ancient era.

This particular `pong` example is basically Fig 5 from https://arxiv.org/pdf/2104.14722.pdf. It's for the use case where you want to define a specific "dynamically corrected gate". The pdf has a code example of how this should be done using openqasm. Is that valid using the most recent spec?

jakelishman:
[...]

PR #373, opened 2022-06-20
Fix calibration grammar handling in ANTLR

### LINKED ISSUE #246: Add a new token to match the body of defcal

### LINKED ISSUE #368: Reference OQ ANTLR parser does not support `cal` statements
The reference ANTLR grammar does not have a parsing rule that matches `cal` statements.  This is a bug - it should know to tokenise the block into a single token and delegate the correct lexing and parsing to the defined calibration grammar parser.

jakelishman:
### Summary

This teaches the ANTLR grammar to recognise `cal` statements (which were
previously entirely absent), and to allow symbolic expressions in place
of classical argument definitions in the argument lists of `defcal`
statements (as stated in the spec).

More completely, the tokenisation of blocks in the calibration updated
has been completely rewritten so that ANTLR tokenises everything between
a balanced `{}` into one single token.  The new requirement that the
brace block is balanced fixes a family of bugs where ANTLR would extend
the (supposedly non-greedy) `LBRACE .*? RBRACE` match to "eat" parser
errors following a `defcal` block.

### Details and comments

This will currently fail the tests because of the issues described in #372.  The example files `pong.qasm`, `t1.qasm` and `defcal.qasm` need correcting in a different PR before this can merge.

This will have some minor merge conflicts with #371, which will be trivial to update after that merges.

Fix #246
Fix #368


jakelishman:
Since #376 merged, this is now unblocked, and can be reviewed for code-only changes (@aspcompiler and @psivaraj, if you want / have time to).

blakejohnson (approved):
Approved as a code-only change.

PR #374, opened 2022-06-20
Correct surrounding OQ 3 code in OpenPulse examples

### LINKED ISSUE #367: OpenPulse-specific syntax present in OpenQASM 3.0 contexts

jakelishman:
### Summary

In some OpenPulse (or general pulse-grammar) examples where the embedded
of the calibration language inside OpenQASM 3 was being demonstrated,
the surrounding OQ 3 code was invalid.  This was mostly due to OpenPulse
types being given in `extern` statements; this would be an OpenPulse
extension, and so needs to be in `cal` block.

Examples which were _entirely_ in the pulse language are unchanged;
there's no advantage in wrapping everything in a `cal` statement, and it
just adds more noise for readers.

### Details and comments

Along with #371, fix the other part of #367.

Since #371 is now merged, this fixes #367.

### DIFF
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -404,12 +404,12 @@ discriminated using user-defined boxcar and discrimination ``extern``\s.
 
     defcalgrammar "openpulse";
 
-    // Use a boxcar function to generate IQ data from raw waveform
-    extern boxcar(waveform input) -> complex[float[64]];
-    // Use a linear discriminator to generate bits from IQ data
-    extern discriminate(complex[float[64]] iq) -> bit;
-
     cal {
+        // Use a boxcar function to generate IQ data from raw waveform
+        extern boxcar(waveform input) -> complex[float[64]];
+        // Use a linear discriminator to generate bits from IQ data
+        extern discriminate(complex[float[64]] iq) -> bit;
+
         // Define the ports
         extern port m0;
         extern port cap0;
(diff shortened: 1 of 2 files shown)


jakelishman:
From TSC meeting on 2022-07-08: this is just fixing a couple of invalid syntax points in the examples, and people were happy to merge it.

PR #384, opened 2022-08-17
Void type

mbhealy:
### Summary
The specification, examples, and grammar already allowed subroutines and externs to not return anything. This PR formalizes that the return type when no return value is provided is `void`, an unrealizable and uninstantiable type. 

### Details and comments
The grammar, parser, and associated tests have also been updated.

### DIFF
--- a/source/language/subroutines.rst
+++ b/source/language/subroutines.rst
@@ -16,16 +16,11 @@ defines a quantum bit argument named ``ancilla``.
 
 Subroutines return up to one value of classical type, signified by the
 ``return`` keyword. If there is no return value, the empty ``return``
-keyword may be used to immediately exit from the subroutine. Subroutines that
-do not return a value may be declared with no output signifiers::
+keyword may be used to immediately exit from the subroutine, which implicitly
+returns the :ref:`void type <void-type>`. Subroutines that
+do not return a value must be declared with no output signature::
 
     def name(parameters) { body }
-
-or may be declared to return the :ref:`void type <void-type>`::
-
-    def name(parameters) -> void { body }
-
-These two forms of subroutine definition are exactly equivalent.
 
 Qubit declarations are not allowed within subroutines as those declarations are global.
 
--- a/source/language/subroutines.rst
+++ b/source/language/subroutines.rst
@@ -25,6 +25,8 @@ or may be declared to return the :ref:`void type <void-type>`::
 
     def name(parameters) -> void { body }
 
+These two forms of subroutine definition are exactly equivalent.
+
 Qubit declarations are not allowed within subroutines as those declarations are global.
 
 A subroutine is invoked with the syntax ``name(parameters)`` and may be assigned
--- a/source/language/subroutines.rst
+++ b/source/language/subroutines.rst
@@ -14,9 +14,16 @@ the subroutine by reference or name, while classical types are passed in by valu
 All arguments are declared together with their type. For example, ``qubit ancilla``
 defines a quantum bit argument named ``ancilla``.
 
-The subroutines return up to one value of classical type, signified by the
-``return`` keyword. If there is no return type, the empty ``return``
-keyword may be used to immediately exit from the subroutine.
+Subroutines return up to one value of classical type, signified by the
+``return`` keyword. If there is no return value, the empty ``return``
+keyword may be used to immediately exit from the subroutine. Subroutines that
+do not return a value may be declared with no output signifiers::
+
+    def name(parameters) { body }
+
+or may be declared to return the :ref:`void type <void-type>`::
+
+    def name(parameters) -> void { body }
 
 Qubit declarations are not allowed within subroutines as those declarations are global.
 
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -180,6 +180,20 @@ types for the purposes of casting.
    // Declare a machine-precision float.
    float my_machine_float = 2.3;
 
+.. _void-type:
+
+Void type
+~~~~~~~~~
+
+Subroutines and externs that do not need to return a value may be declared to
+return ``void``. The ``void`` type is unrealizable and uninstantiable, and thus
+cannot be attached to an identifer.
+
+.. code-block::
+
+   extern updateMetadata(bit[10]) -> void;
+   def modifyArray(mutable array[int[8], 10] arr_arg) -> void { ... }
+
 .. _angle-type:
 
 Angles
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -185,9 +185,9 @@ types for the purposes of casting.
 Void type
 ~~~~~~~~~
 
-Subroutines and externs that do not need to return a value may be declared to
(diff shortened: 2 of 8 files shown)


source/language/subroutines.rst
```diff
@@ -14,9 +14,16 @@ the subroutine by reference or name, while classical types are passed in by valu
...
+Subroutines return up to one value of classical type, signified by the
+``return`` keyword. If there is no return value, the empty ``return``
+keyword may be used to immediately exit from the subroutine. Subroutines that
+do not return a value may be declared with no output signifiers::
+
+    def name(parameters) { body }
+
+or may be declared to return the :ref:`void type <void-type>`::
+
+    def name(parameters) -> void { body }
```
jakelishman, line 26: It might be worth adding an extra sentence after this just to clarify that "no specified return" and `-> void` are exactly equivalent (assuming that they are).

source/language/types.rst
```diff
@@ -180,6 +180,20 @@ types for the purposes of casting.
    // Declare a machine-precision float.
    float my_machine_float = 2.3;
 
+.. _void-type:
+
+Void type
+~~~~~~~~~
+
+Subroutines and externs that do not need to return a value may be declared to
```
jakelishman, line 188:
```suggestion
Subroutines and externs that do not return a value may be declared to
```
(to avoid any potential implication that return types could be variable)

source/language/types.rst
```diff
@@ -180,6 +180,20 @@ types for the purposes of casting.
...
+return ``void``. The ``void`` type is unrealizable and uninstantiable, and thus
+cannot be attached to an identifer.
```
jakelishman, line 190: Shall we say "or used as a cast operator" as well for completeness?  It's already implied by "uninstantiable", so the addition may be unnecessary.

jakelishman (commented):
Minor comments aside, I'm happy with the technical content of this PR.  I'm curious what the motivation for it was, though?  We don't have pointer types or first-class functions, so I can't immediately see the need to be able to name the unit type.

The consistency is nice, but I'm not sure `-> void` is any clearer than the absence of a return type.  As we use the keywords `def` or `extern` to introduce subroutine declarations, it's also superfluous grammatically (unlike in C, say).

steleman:
> Minor comments aside, I'm happy with the technical content of this PR. I'm curious what the motivation for it was, though? We don't have pointer types or first-class functions, so I can't immediately see the need to be able to name the unit type.
> 
> The consistency is nice, but I'm not sure `-> void` is any clearer than the absence of a return type. As we use the keywords `def` or `extern` to introduce subroutine declarations, it's also superfluous grammatically (unlike in C, say).

`void` does not denote, or signify, the absence of a type. Quite the opposite.

Also, there is no such thing as the _absence of a return type_ for functions. All functions return a type. There are not two distinct categories of functions - ones that return a type,  and others that don't.

If a function ever were not to return a type, it would never return. 

Returning `void` from a function is indicated either by the special statement
```
return;
```

with no operand, or, by the absence of a `return` statement, in which case the function implicitly returns `void` after the last statement.

Just like when writing

```
def function_with_void_return(int i, double d) {
  // code goes here
}
```

the absence of an explicit return type after the right paren does not mean that the function does not return. It only means that the function returns `void`.

By definition, `void` is an incomplete type that does not have a size, cannot perform arithmetic and cannot be either an `lvalue` or an `rvalue`.

A `void` type and a `pointer-to-void` type are two different things that have very little in common. Since OQ3 does not have pointers of any kind, it cannot have pointers to void either.

A function that _appears_ not to return anything is a function that returns this special `void` type that cannot be materialized.

jakelishman:
By "absence of a return type", I was implying "absence of a specified return type".  I understand the distinction between zero types and unit types.

My additional comments about pointers and first-class functions were to contrast OQ 3 to other languages.  As I also said, OQ 3 doesn't have such objects, which are some of the reasons usually used to justify spelling out a unit type.  Since OQ 3 doesn't have them and in the one circumstance `void` could be valid in this PR, it's also the "default" (by omitting the `-> <type>`), I'm just curious as to what other reasons were discussed about why we might want to add the extra keyword.

steleman:
> I'm just curious as to what other reasons were discussed about why we might want to add the extra keyword.

Because the current Spec does not explain what is being returned from a function that does not _appear to_ have a return type. Which implies there exists an assumption that functions that do not specify a return type do not return anything.

That assumption is incorrect. As I have already explained, functions must return _something_. For functions that do not specify a return type, the `void` type is the implicit _something_ being returned.

mbhealy:
> For functions that do not specify a return type, the `void` type is the implicit _something_ being returned.

This is really the essence of the PR in my mind. We are specifying something concretely that was hinted at before, just to ensure that there is no ambiguity possible.

And yes, Jake, I didn't make it clear enough in the first draft that the non-specified version is implicitly returning void, I'll amend the PR today to address that.

jakelishman:
That's fair.  Just to be clear, I'm not questioning implicit returns of a void type, and I'm fine with the additional text surrounding such functions to remove ambiguity.  I'm only questioning if we need to add explicit `-> void` syntax.  Apologies if that wasn't as clear as it could have been.  I don't feel strongly about it either way.

steleman:
> I'm only questioning if we need to add explicit `-> void` syntax.

Nope. No syntax changes are necessary.The compiler will take care of figuring out that it's a `void` return type.

source/language/subroutines.rst
mbhealy: Added in https://github.com/openqasm/openqasm/pull/384/commits/73136b3c45e5e6381e771d93c071e8758761a19e

source/language/types.rst
mbhealy: Removed in https://github.com/openqasm/openqasm/pull/384/commits/73136b3c45e5e6381e771d93c071e8758761a19e

mbhealy:
I would be fine removing the syntax changes if desired.

mbhealy:
At the TSC meeting today we decided to leave the syntax changes as is. The current version is ready for review by all.

jakelishman (dismissed):
Thanks Stefan, yeah, that's as I had understood it.

From the working group meeting today: Bettina (@bettinaheim) pointed out that it's easier to release a previously reserved keyword than it is to reserve something that wasn't already.  On that basis, I'm happy to move ahead with `-> void` as optional syntax.  I'm happy with all the other text in the PR.

mbhealy:
Per the discussion at last week's TSC meeting, I have removed the optional void return signature. All grammar and parser changes have been reverted and the specification text has been updated accordingly. The `void` keyword is now reserved for possible future use, so I added it back to the lexer, but there are currently no valid uses of it.

mbhealy:
This PR is now ready for final review/approval by the WG members and TSC.

stevenheidel (approved):

jakelishman (approved):

levbishop (approved):
LGTM

PR #386, opened 2022-08-22
Make the semantic of `gphase();` clearer. 

nakrakiiya:
### Summary

Make the semantic of `gphase();` clearer. See https://github.com/openqasm/openqasm/issues/383 .
This definition of `gphase()` also allows accumulation.

### Details and comments

The semantic of `gphase();` in Amazon Braket is just add a global phase gate to the containing scope. No matter which qubit is applied to, the result state of the circuit is the same. In this case, an n-qubits global phase gate is more reasonable.

If someone just want the state of a specific qubit, this design cannot distinguish which qubit `gphase();` applies to.

The *"scope"* is also more specific. Only gate, subroutine, and global scope have qubits declaration/definition. If we only limit to the containing scope, for example, the `for` scope, the semantic can also be *"do nothing"* because no qubits declared/defined in the `for` scope.

### DIFF
--- a/source/language/gates.rst
+++ b/source/language/gates.rst
@@ -132,7 +132,15 @@ From a physical perspective, the gates :math:`e^{i\gamma}U` and :math:`U` are eq
 phase :math:`e^{i\gamma}`. When we add a control to these gates, however, the global phase becomes a relative phase
 that is applied when the control qubit is one. To capture the programmer's intent, a built-in global phase gate
 allows the inclusion of arbitrary global phases on circuits. The instruction ``gphase(γ);`` adds a global phase
-of :math:`e^{i\gamma}` to the scope containing the instruction. For example
+of :math:`e^{i\gamma}` to the gate, subroutine, or global scope which containing the instruction. This is equivalent to 
+adding an n-qubits global phase gate 
+
+.. math::
+
+   \text{gphase(γ)} := e^{i\gamma} I_{2^n}
+
+, where :math:`I_{2^n}` is an identity matrix with size :math:`2^n`, and n is the number of qubits declared in the scope.
+For example
 
 .. code-block::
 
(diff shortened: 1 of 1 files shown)


source/language/gates.rst
```diff
@@ -132,7 +132,15 @@ From a physical perspective, the gates :math:`e^{i\gamma}U` and :math:`U` are eq
 phase :math:`e^{i\gamma}`. When we add a control to these gates, however, the global phase becomes a relative phase
 that is applied when the control qubit is one. To capture the programmer's intent, a built-in global phase gate
 allows the inclusion of arbitrary global phases on circuits. The instruction ``gphase(γ);`` adds a global phase
-of :math:`e^{i\gamma}` to the scope containing the instruction. For example
+of :math:`e^{i\gamma}` to the gate, subroutine, or global scope which containing the instruction.
```
stevenheidel, line 135: I'm not sure if it was ever spelled out, but I believe `gphase` was only intended to appear in a gate definition. Maybe @blakejohnson can clarify

jakelishman: cc @ajavadia and @levbishop, since Blake's on holiday.

levbishop: In my mind `gphase` can appear anywhere. However, it's only visible if it gets `ctrl` or `negctrl` modified. Thus, it can make a difference if used in a `gate` body, since `gates` can become controlled. Subroutines and global scopes can't be control modified so the global phase isn't observable from those contexts.

An inline-controlled `gphase` such as `ctrl @ gphase(theta) q;` is not actually applying a global phase and is observable in any scope.

stevenheidel (approved):

jakelishman:
From TSC meeting 2022-09-02: there is a question of what `gphase` would mean inside a subroutine, and whether it's worth needing to specify any special rules for it.  In particular, it's not obvious what the behaviour would be if the same qubit is passed as different arguments.

There was approximate agreement (without a quorum) to:
- accept this PR as-is
- (Lev) largely rewrite `gates.rst` to simplify all of its text, including this PR
- add a separate PR to explicitly forbid qubit references to alias the same qubit in subroutine calls

We need further assent/comment from @pschindler, @blakejohnson or @ajavadia before we can merge this, though.

blakejohnson:
Is this now subsumed by #393 ?

levbishop:
> Is this now subsumed by #393 ?
Yes my intention is that #393 morally includes this revision.

jakelishman:
Closing in favour of #393.

PR #388, opened 2022-09-09
Increment the OpenQASM3 reference parser version to 0.2.0

aspcompiler:
### Summary

Increment the version of reference parser so that we can publish a new package.

### Details and comments

I tagged my local branch with `py/v0.2.0`. Not sure if the tag is transferred to the PR. If not, we need to tag separately.


jakelishman (approved):
Yeah, let's go for it!

Issue #389, opened 2022-09-12
Printer should not add an extra space when printing barrier without argument


jcjaskula-aws:
### OpenQASM version:

3

### Python version:

3.7 to 3.10

### Operating system:

Mac

### What is the current behavior?

`Printer` should not add an extra space when printing `barrier` without argument. Currently, it outputs `barrier ;`

### Steps to reproduce the problem

One just print the QuantumBarrier Node with an empty list.

### What is the expected behavior?

Printing `barrier;`

### Suggested solutions (Optional)

See #387. 

Also node.qubits could also be optional instead of passing an empty list.

jakelishman:
We can add that change to the printer pretty trivially, the same as for `delay`, since the spec says that barrier can take no arguments.

I'm not a huge fan of changing the type of the attribute `QuantumBarrier.qubits` to `Optional[List[...]]` - it forces every downstream consumer to specifically handle the `None` type, which has an identical meaning to `[]`.  It feels more type-safe and like better data normalisation to me to avoid the nullable type in favour of the existing empty representation.

If you just mean the constructor and not the attribute, then I've no problem with adding default handling (so `QuantumBarrier()` is valid and produces a node with `qubits=[]`).

jcjaskula-aws:
I initially meant the attribute but I don't have a strong conviction in the way I would like to modify it. 

Personally, `barrier()` means "place a barrier" (so by default on all the frames/qubits). `barrier(list)` means "place a barrier on all the frames/qubits in the list". If the list is empty, it would mean that the barrier on no qubit. For instance, it would not make sense to have a barrier on all frames if the list constructed programmatically by [f for f in frames if is_readout(f)]) is empty. Obviously, one could just check if the list is not empty and accordingly apply the barrier.

jakelishman:
I'm happy for us to make this constructor (and a few other places which have optional arguments) have default implementations in the constructors.

In the AST, a `Barrier` node (necessarily) corresponds to a `barrier` statement in an OpenQASM 3 program.  In that, the semantics of the `barrier` instruction are that a `barrier` on no qubits is the same as a barrier on all of them; there's no language support for writing the word `barrier` followed by an explicitly empty list.  The AST needs to follow the semantics of the language - we can only change its data representation based on the text of the spec.  A lot of the point of the AST is to normalise the representation of the OQ3 program, and having multiple ways of representing the same data is kind of contrary to that.

PR #390, opened 2022-09-12
Fix space in barrier without argument

jcjaskula-aws:
This commit add a space in barrier only when it is called with arguments.

### Summary
Fixes issue #389

### Details and comments
Similar to fix for delay[duration];


jakelishman (approved):
Looks good to me, thanks!

PR #391, opened 2022-09-16
Forbid qubit aliasing in subroutine calls

jakelishman:
### Summary

This restriction was discussed in the TSC meeting on 2022-09-02, to avoid complications in the definition of `gphase` within subroutine scopes, and to ease implementations.  It may be reasonable to lift the restriction in a later language version.

### DIFF
--- a/source/language/subroutines.rst
+++ b/source/language/subroutines.rst
@@ -14,7 +14,7 @@ the subroutine by reference or name, while classical types are passed in by valu
 All arguments are declared together with their type. For example, ``qubit ancilla``
 defines a quantum bit argument named ``ancilla``.
 
-A given qubit can passed at most once in any subroutine call.  Different
+A given qubit can be passed at most once in any subroutine call.  Different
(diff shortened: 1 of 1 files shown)


source/language/subroutines.rst
```diff
@@ -14,6 +14,10 @@ the subroutine by reference or name, while classical types are passed in by valu
 All arguments are declared together with their type. For example, ``qubit ancilla``
 defines a quantum bit argument named ``ancilla``.
 
+A given qubit can passed at most once in any subroutine call.  Different
```
mbhealy, line 17:
```suggestion
A given qubit can be passed at most once in any subroutine call.  Different
```

mbhealy (commented):
One suggestion, otherwise LGTM

source/language/subroutines.rst
jakelishman: Thanks - this is what happens when I try to make two PRs in 3 minutes!

mbhealy (approved):
LGTM

PR #392, opened 2022-09-16
Reserve switch keywords

jakelishman:
### Summary

From TSC meeting 2022-09-02, we anticpate expanding the OpenQASM 3 language with these keywords, but they may not fully be in time for OpenQASM 3.0.  It's easier to allow something previously forbidden than the opposite, so we are taking this step proactively.

### Details and comments

This PR may be obsoleted by one that more completely adds `switch` and `case` support.


mbhealy (approved):
LGTM

PR #397, opened 2022-10-12
Update example for indexes

stevenheidel:
### Summary

One of the indexing examples isn't consistent with the definition, according to my read of the spec.

### Details and comments

Here's the definition:

image

`a = 0`, `b = 6`, and `c = 1` (by default). The largest integer m such that `a + mc <= b` is `m=6`, so there will be seven entries in the output array.


jakelishman (approved):
Yeah, this looks right to me too thanks!

PR #400, opened 2022-10-14
Load ANTLR files based on dynamic runtime version

jakelishman:
### Summary

This adds the capability for the reference package to have multiple generated versions of the ANTLR parser files simultaneously, separated by version information.  The ANTLR runtime only needs to match the major (4) and minor version.  Generated files should be put in
`openqasm3/_antlr/_<major>_<minor>`.  The `_antlr` package modifies the meta-path to redirect imports to itself into the correct concrete version.

CI is split up into "build" and "test" components, since the build is now fairly complex.  This will also make writing a release pipeline almost trivial, as the publishable artifacts are available immediately.

### Details and comments

I set up CI to build a version of the package with ANTLR files generated from versions 4.7 to 4.11.  4.7 was first released about 6 years ago, and 4.11 is the most recent release.  I think this should be enough for everything somebody might want to install at the same time.


aspcompiler:
LGTM. I manually test it and it worked great! Please update the version to 0.4.0 since 0.3.0 was already published.

aspcompiler (approved):

aspcompiler (approved):

jakelishman:
My intent is to follow this PR with another one that puts a deployment pipeline and deployment instructions in place, that will include procedures for doing a release.  It'd probably be good for you, me and Prasahnt to sit down in the next couple of weeks and agree on policies / maintenance - I'm keen to start moving Qiskit towards relying on this package in the near-ish future as well.

jakelishman:
(For now, I've just bumped the version numbers on `main` directly rather than doing it in this unrelated PR.)

aspcompiler (approved):

PR #403, opened 2022-10-21
Use array literal syntax in raw waveform example

braised-babbage:
Two small spec tweaks:
1. use array literal syntax in the raw waveform example
2. deprecate the parenthetical remark about https://github.com/openqasm/openqasm/pull/301 which is now closed

I'm not actually clear on whether the raw waveform syntax is still up in the air, so I'm happy to adjust the language further.

### DIFF
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -238,7 +238,7 @@ Waveforms
 
 Waveforms are of type ``waveform`` and can either be:
 
-- An array of complex samples (note this syntax is still under [active development](https://github.com/Qiskit/openqasm/pull/301) and is subject to change.) which define the points for the waveform envelope
+- An array of complex samples which define the points for the waveform envelope.
 - An abstract mathematical function representing a waveform. This will later be
   materialized into a list of complex samples, either by the compiler or the
   hardware using the parameters provided to the ``extern`` declared waveform template.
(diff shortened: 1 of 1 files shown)


source/language/openpulse.rst
```diff
@@ -265,7 +265,7 @@ samples provides optimization opportunities that wouldn't be available otherwise
    :force:
 
    // arbitrary complex samples
-   waveform arb_waveform = [1+0im, 0+1im, 1/sqrt(2)+1/sqrt(2)im];
+   waveform arb_waveform = {1+0im, 0+1im, 1/sqrt(2)+1/sqrt(2)im};
```
jakelishman, line 268: This is technically a change to the OpenPulse specification (based purely on the text of this patch) - the OpenPulse `waveform` type isn't inherently linked to the OpenQASM 3 `array` type, so there's no hard requirement that their literals actually need to have the same syntax.  I don't know precisely what state the OpenPulse specification is in, nor what implementors are doing.  I think the `openpulse-python` project is implicitly assuming that the waveform literal is the same as the `array` literal, from the way it inherits from the `qasm3Lexer.g4` tokens, but given this text, that's not actually correct.

Let me throw this over to @openqasm/wg-openpulse to comment on what the right answer here is, and for somebody to shepherd any other changes needed - I suspect the right answer is that the examples _do_ need updating as in this PR.

(On a semi-related note, the complex-number definitions in this example will likely fail to parse, and I don't think there's any explicit support for it in the OpenPulse spec.  Certainly it would be invalid OQ3 since #358, and OpenPulse usually just reuses the OpenQASM 3 rules.)

braised-babbage:
Good point re: OpenPulse vs OpenQASM. In retrospect it's not clear to me that using the array literal syntax for waveforms is appropriate either, so I'm going to close this issue and open a different one in the openpulse reference parser repo.

PR #404, opened 2022-10-27
Remove version string from most examples, two have been left intentionally

mbhealy:
### Summary
As discussed at the TSC meeting on October 14th, this PR removes the OPENQASM version string from most of the examples

### Details and comments
The version string is optional, and leaving it out is the general default. It was originally included to enable any potential breaking changes to be included, but since we have been able to maintain backward compatibility it is mostly unnecessary, and the TSC decided that compilers should/can decide for themselves which parser to use based on the features in the translation unit being parsed.

### DIFF
--- a/examples/adder.qasm
+++ b/examples/adder.qasm
@@ -2,7 +2,6 @@
  * quantum ripple-carry adder
  * Cuccaro et al, quant-ph/0410184
  */
-OPENQASM 3;
 include "stdgates.inc";
 
 gate majority a, b, c {
--- a/examples/alignment.qasm
+++ b/examples/alignment.qasm
@@ -3,7 +3,6 @@
  * specify design intent on gate alignment, without
  * being tied to physical qubits and gates.
 */
-OPENQASM 3.0;
 include "stdgates.inc";
 
 stretch g;
--- a/examples/dd.qasm
+++ b/examples/dd.qasm
@@ -2,7 +2,6 @@
  * This example demonstrates the use of referential delays
  * and time alignment.
 */
-OPENQASM 3.0;
 include "stdgates.inc";
 
 stretch a;
--- a/examples/gateteleport.qasm
+++ b/examples/gateteleport.qasm
@@ -1,5 +1,4 @@
 // Repetition code gate teleportation
-OPENQASM 3;
 include "stdgates.inc";
 
 // declarations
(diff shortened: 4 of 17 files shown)


jakelishman (approved):
This looks like what we agreed on to me.  For posterity, the example files not touched by this PR are:
- `arrays.qasm`: retains its `OPENQASM 3.0;` statement.
- `cphase.qasm`: doesn't have a version statement
- `defcal.qasm`: doesn't have a version statement
- `teleport.qasm`: retains its `OPENQASM 3;` statement.

jakelishman:
We perhaps could annotate the version statement in the `teleport.qasm` example to explicitly say that the `OPENQASM 3;` statement is optional, so we're not only saying it implicitly by its absence in other examples?  We technically were doing that already, since two files didn't have it (and it being optional is in the spec), but that still led to some questions.

mbhealy:
Good idea, will add some comments

jakelishman:
Oh, for further context for others: the [current version of the OQ3 paper on arXiv](https://arxiv.org/abs/2104.14722v2) is where the confusion arises from.  It says (sec. 3)

> Top-level circuits in OpenQASM2 must begin with the statement `OpenQASM 2.x;`, possibly preceded by one or more blank lines or lines of comments [...] Top-level circuits in OpenQASM3 begin similarly, *e.g.*, with a statement `OpenQASM 3.x;`

[sic: both sentences are actually false on the basis of their improper capitalisation of the `OPENQASM` token, but that's beside the point]

The living spec in this repository, and hosted at openqasm.com has always instead said that the version statement is optional in OpenQASM 3 programs.  If we have a chance to update the paper, it would be good to correct this obvious conflict.

mbhealy:
> We perhaps could annotate the version statement in the `teleport.qasm` example to explicitly say that the `OPENQASM 3;` statement is optional, so we're not only saying it implicitly by its absence in other examples? We technically were doing that already, since two files didn't have it (and it being optional is in the spec), but that still led to some questions.

Added in https://github.com/openqasm/openqasm/pull/404/commits/b7d112f4ffcf01d0cdded559013e3dd563c947b0

jakelishman (approved):

jakelishman:
From TSC 2022-10-28: this matches expectations, so merging.

Issue #398, opened 2022-10-13
Defcals for delays


PhilReinhold:
### What is the expected behavior?

defcal delay[time] $1 {
    delay[time] frame_1;
    delay[time] frame_2;
}

stevenheidel:
This was discussed at the TSC meeting on October 14th. In summary this is a great idea for mapping qubits to frames.

We had previous discussed adding  a duration argument to the `defcal` definition but dismissed it at the time as having no interesting use case. This however is a compelling use case.

The suggestion is to move forward with the syntax you suggested:
```
defcal delay[time] $1 {
    delay[time] frame_1;
    delay[time] frame_2;
}
```
and to keep the delay invocation as is. We would welcome a PR making this change to the grammar and spec!

PR #405, opened 2022-10-27
Add support for defcal delay with square brackets to grammar

PhilReinhold:
### Summary

Closes issue #398 

### Details and comments

This add support for the following syntax:

```
defcal delay[duration_name] qubit_name { ... }
```

When parsing this the resulting ast node is a `CalibrationDefinition` where `name` is "delay" and `arguments` has a single `ClassicalArgument` entry with type `DurationType` and name `duration_name`. That is to say, the node parses as if the user had written 

```
defcal delay(duration duration_name) qubit_name { ... }
```

### DIFF
--- a/source/grammar/qasm3Parser.g4
+++ b/source/grammar/qasm3Parser.g4
@@ -203,7 +203,7 @@ arrayReferenceType: (READONLY | MUTABLE) ARRAY LBRACKET scalarType COMMA (expres
 
 designator: LBRACKET expression RBRACKET;
 
-defcalTarget: MEASURE | RESET | DELAY | Identifier;
+defcalTarget: MEASURE | RESET | DELAY LBRACKET Identifier RBRACKET | Identifier;
 defcalArgumentDefinition: expression | argumentDefinition;
 defcalOperand: HardwareQubit | Identifier;
 gateOperand: indexedIdentifier | HardwareQubit;
--- a/source/language/pulses.rst
+++ b/source/language/pulses.rst
@@ -27,8 +27,11 @@ instruction sequence on *physical* qubits, e.g.
 
    defcal rz(angle[20] theta) $0 { ... }
    defcal measure $0 -> bit { ... }
+   defcal delay[time] $0 { ... }
    defcal measure_iq q -> complex[float[32]] { ... }
 
+defcals may be defined for the qubit ``measure``, ``reset``, ``delay`` operators as
+well as for arbitrarily named gates.
 We distinguish gate and measurement definitions by the presence of a
 return value type in the latter case, analogous to the subroutine syntax
 defined earlier. Furthermore, the return value type does not need to return a
--- a/source/openqasm/openqasm3/parser.py
+++ b/source/openqasm/openqasm3/parser.py
@@ -331,6 +331,20 @@ class QASMNodeVisitor(qasm3ParserVisitor):
             self.visit(ctx.returnSignature().scalarType()) if ctx.returnSignature() else None
         )
 
+        defcal_target: qasm3Parser.DefcalTargetContext = ctx.defcalTarget()
+        if defcal_target.DELAY():
+            duration_ident = defcal_target.Identifier()
+            span = get_span(duration_ident)
+            arguments.append(
+                add_span(
+                    ast.ClassicalArgument(
+                        type=add_span(ast.DurationType(), span=span),
+                        name=add_span(ast.Identifier(duration_ident.getText()), span=span)
+                    ),
+                    span=span,
+                )
+            )
+
         return ast.CalibrationDefinition(
             name=self.visit(ctx.defcalTarget()),
             arguments=arguments,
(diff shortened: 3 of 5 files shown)


source/grammar/qasm3Parser.g4
```diff
@@ -203,7 +203,7 @@ arrayReferenceType: (READONLY | MUTABLE) ARRAY LBRACKET scalarType COMMA (expres
 
 designator: LBRACKET expression RBRACKET;
 
-defcalTarget: MEASURE | RESET | DELAY | Identifier;
+defcalTarget: MEASURE | RESET | DELAY LBRACKET Identifier RBRACKET | Identifier;
```
jakelishman, line 206: Technical bikeshedding: I would still consider `delay` to be the target of the `defcal`, and the duration specifier to be a separate subrule of the statement.  Perhaps we could instead factor it as
```antlr4
defcalStatement:
    DEFCAL defcalTarget
    (LBRACKET defcalDuration RBRACKET)?
    (LPAREN defcalArgumentDefinitionList? RPAREN)?
    // ...
```
and leave `defcalTarget` alone?  Then in the AST visitor we'd error on any program that supplied the `defcalDuration` with an invalid target instead.  That leaves open the possibility of other identifiers being able to take a duration in their argument definition as well (not a language feature we're looking to add in the 3.0 release, but it's been discussed a fair amount in relation to a 3.1+ release).

The value of the rule `defcalDuration` to me depends on what we decide about the text of the spec first - the textual spec is our source of truth, and then this grammar is meant to be an indicative implementation of that.

(The reason I'm bringing it up when the outcome is identical right now is just because this ANTLR grammar is also meant to demonstrate how someone should _think_ about the parse tree of the language, as a counterpart to the written spec.)

source/language/pulses.rst
```diff
@@ -27,8 +27,11 @@ instruction sequence on *physical* qubits, e.g.
 
    defcal rz(angle[20] theta) $0 { ... }
    defcal measure $0 -> bit { ... }
+   defcal delay[time] $0 { ... }
    defcal measure_iq q -> complex[float[32]] { ... }
 
+defcals may be defined for the qubit ``measure``, ``reset``, ``delay`` operators as
+well as for arbitrarily named gates.
```
jakelishman, line 34: I think we'll need to expand this to explain much more about the new `[duration]` syntax, for example where it's valid (only `delay`, or any identifier? - without speaking for everyone, my impression is that there's only support in the TSC for `delay` supporting the time argument in 3.0, nothing else).

There's also a question of what the definition should look like here.  In `defcal`s of parametrised gates, there's two options for a parameter definition - it can either be an `expression` (which must evaluate to a type of `const angle[_]`), or it's a `type identifier` declaration (like `angle[12] theta`).  I think there's a decision that needs to be made here along similar lines:

1. have the only allowed thing be `"[" identifier "]"`, which causes `identifier` to be a parameter that's implicitly of type `duration` within the `defcal` body (subject to however that crosses into the particular pulse grammar)
2. have this be `"[" "duration" identifier "]"`, with the type explicit.
3. the same as point 2, but also allow `"[" expression "]"`, where `expression` must evaluate to a type `const duration`.

Form 1 is what's currently proposed, but may well be at odds with how types are explicit in (almost) every other part of the language, include the other parts of `defcal` statements.  Form 2 is perhaps more consistent with how parameters are usually declared.  Form 3 would perhaps be the most similar to the other parameters in `defcal`s are declared, and would permit things like specifying a `defcal delay[10ms]` that's different at compile time to a more general`defcal delay[duration t]` - not so useful for `delay`, but makes more sense in the context of future extensions of this syntax to gates.

(My immediate preference in OQ 3.0 is for form 2, but it's not my decision - it's for the TSC and for other implementors to weigh in on.)

source/openqasm/openqasm3/parser.py
```diff
@@ -331,6 +331,20 @@ def visitDefcalStatement(self, ctx: qasm3Parser.DefcalStatementContext):
...
+            span = get_span(duration_ident)
+            arguments.append(
+                add_span(
+                    ast.ClassicalArgument(
+                        type=add_span(ast.DurationType(), span=span),
+                        name=add_span(ast.Identifier(duration_ident.getText()), span=span)
+                    ),
+                    span=span,
+                )
+            )
```
jakelishman, line 346: This implicitly overloads the `argument` field of the node in the AST, making
```
defcal delay[t]
```
appear very similar to another symbol that could have been defined as (invalid typing in the argument list aside)
```
defcal my_symbol(duration t)
```

To me, the "gate parameter list" and the "duration specifier" would be better off as two logically separate abstract parts of the syntax: subject to what's decided in the text of the specification, I'd be in favour of adding a `.duration` field to the AST node `CalibrationDefinition` instead of overloading the parameter list like this.

jakelishman (commented):
Thanks for the PR!

This was briefly covered in the TSC meeting today (2022-10-28), though most of the TSC were absent.  To move forwards, particularly with discussion of what the new feature might look like, we'll really need to flesh out the textual parts of its description in the spec.  The text is meant to be the source of truth - the surrounding code is all just meant to be an indicative implementation.

It would also be good to add some lines into the examples in the `examples` folder in the repo root (`defcal.qasm` would be a fine place, for example).

blakejohnson:
Possible topic for further discussion in a TSC meeting?

source/language/pulses.rst
jwoehr: @PhilReinhold can you please update this PR in light of above discussion?

jwoehr: Thanks, Phil

PR #412, opened 2022-12-02
Document with contributors

jakelishman:

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -41,18 +41,21 @@ Zachary Schoenfeld
 
 # Contributors via GH interaction / WG participation
 
-Matthew Amy
-Phil Reinhold
-Dor Israeli
-Niel de Beaudrap
-Derek Bolt
-Peter Karalekas
-Lauren Capelluto
-BẢO Bach Gia
-Jialin Dou
-Erik Davis
-Vincent Pascuzzi
+Alex McCaskey
 Bill Ticehurst
+BẢO Bach Gia
+Derek Bolt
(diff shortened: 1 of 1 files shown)


contributors.md
```diff
@@ -0,0 +1,61 @@
...
+Luciano Bello
+Matthew Treinish
+Michael Healy
+Noah Wood
+Pranav Gokhale
+Prasahnt Sivarajah
+Salvadore de la Puente
+Thomas Alexander
+Travis Scholten
+Yunong Shikj
```
k4rtik, line 37: @jakelishman I believe Yunong's last name is "Shi".

jakelishman: oh whoops thanks - that's what I originally typed, then fidgeted around in vim without exiting insert mode!

blakejohnson (dismissed):
What did you pick as a cutoff date for gather contributors by commits? I believe your list corresponds to people that have made a commit in the OpenQASM 3 "era".

jakelishman:
I ran some variant of `git shortlog --no-merges 2.0...` - i.e. everything on `main` after the commit tagged `2.0`, and then very mildly filtered to avoid double-counting people who didn't have consistent names in their commits, or who were already counted for other contributions.  I used that tag, rather than an explicit date.

jakelishman:
I have subsumed this into #413, so it's all in the same place for posterity for the election.  There was never any particular intent to merge this PR - it was just a convenient way of receiving comments.

PR #414, opened 2022-12-02
RFC: Use extern constructors for arbitrary waveforms in the OpenPulse spec

braised-babbage:
### Summary

This proposes a revision to the OpenPulse spec, in which the special square bracket syntax `[0.1+0.1im, 0.2+0.2im]` is replaced by a notion of variadic waveform constructor, e.g. `arb(0.1+0.1im, 0.2+0.2im)`. h/t @stevenheidel 

Note: This PR presents an alternative to the effort associated with [this MR](https://github.com/openqasm/openpulse-python/pull/12) for waveform literal handling in the OpenPulse reference parser. Some more information about design tradeoffs associated with waveform literals vs array literals may be found in that link.

### Details and comments

There is a clear need for arbitrary waveforms in OpenPulse, but it's not clear that having a special syntax for them is entirely justified. Likewise, there are complications with overloading array literal syntax for this purpose. 

A significant upside to the square bracket syntax is that it makes writing a certain kind of arbitrary waveform (e.g. defined by a list of complex samples) very comfortable. However, implementors may find it useful to support more than one sort of arbitrary waveform. For example
1. In addition to allowing an explicit list of complex IQ values, it may be useful to allow for a list of (implementation specific) integer DAC codewords.
2. There may be some other variadic constructors which are meaningful, e.g. constructing a piecewise constant or piecewise linear waveforms.

By using the same `extern` notion as is used by parametric waveforms, the above could be handled gracefully, e.g.
1. `dacwaveform(32768, 32767, 32766, 32765, ...)`
2. `step(16ns, 0.1, 0.2, 0.3, 0.4)` and `ramp(16ns, 0.1, 0.2, 0.3, 0.4)`

I have tried to give sufficient room in the spec for implementors to freely experiment with these notions. In particular, I have stressed that the available set of waveform constructors, and their arguments, is implementation specific.

The primary difficulty with the changes in this PR is that OpenQASM does not have any notion of variadic functions, and so even the type signature of the `extern` constructors would be out of spec. I do not think this is a significant problem for implementors, since the actual constructor calls are syntactically legal, and presumably there is already some special treatment of waveform constructors, relative to ordinary function calls. But if we go ahead with the changes in this PR, we need to be mindful of future implications should OpenQASM or OpenPulse wish to introduce first-class variadic functions.

### DIFF
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -237,7 +237,7 @@ Waveforms
 ---------
 
 OpenPulse introduces a new type, ``waveform``. Within the language itself, waveforms are
-somewhat opaque object. They may be constructed and used as arguments for operators
+somewhat opaque objects. They may be constructed and used as arguments for operators
 which accept waveforms (e.g. the ``play`` instruction below).
 
 Ultimately, waveforms are realized as a sequence of samples which define the points of
--- a/source/language/openpulse.rst
+++ b/source/language/openpulse.rst
@@ -236,36 +236,45 @@ Manipulating frames based on the state of other frames is also permitted:
 Waveforms
 ---------
 
-Waveforms are of type ``waveform`` and can either be:
+OpenPulse introduces a new type, ``waveform``. Within the language itself, waveforms are
+somewhat opaque object. They may be constructed and used as arguments for operators
+which accept waveforms (e.g. the ``play`` instruction below).
 
-- An array of complex samples (note this syntax is still under [active development](https://github.com/Qiskit/openqasm/pull/301) and is subject to change.) which define the points for the waveform envelope
-- An abstract mathematical function representing a waveform. This will later be
-  materialized into a list of complex samples, either by the compiler or the
-  hardware using the parameters provided to the ``extern`` declared waveform template.
+Ultimately, waveforms are realized as a sequence of samples which define the points of
+the waveform envelope. We adopt a uniform convention that waveforms are defined via
+waveform constructors, which are ``extern`` functions of two varieties:
+- Parametric waveform constructors take a fixed number of arguments and represent an
+  abstract mathematical function, such as a ``gaussian``.
+- Arbitrary waveform constructors take a variable number of arguments, representing
+  explicit sample values. Note that while OpenQASM does not currently allow for
+  user-defined variadic functions, implementors may explicitly handle arbitrary waveform
+  constructors in a compiler.
 
-A value of type ``waveform`` may be defined either by explicitly constructing the complex samples
-or by calling one of the waveform template functions provided by the target device.
-Note that each of these extern functions takes a type ``duration`` as an argument,
-since waveforms must have a definite duration.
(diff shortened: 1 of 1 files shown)


braised-babbage:
I'd add that while I have tried to carve out space for implementors to consider various waveform constructors, I'm not fundamentally against having some standards here (that is part of the point of a spec, after all). So I'm happy to hear if anybody has opinions on this.

braised-babbage:
One thing this is missing: a proposal for how to express variadic extern declarations, e.g. `extern arb(complex, ...)`

braised-babbage:
I will give a quick update on my own thinking here. We have an OpenPulse implementation and have found it preferable for all waveforms to be defined by `extern` constructors. One problem with `waveform wf = {...}` as a definition is that the mathematical interpretation depends on some extra information (e.g. a sample rate). 

We instead use an explicit array declaration with the samples and then a constructor like the following

```
array[complex, 20] wf_samples = { ... };
waveform wf = discrete(20ns, wf_data);
```

In this particular example, we define a waveform of duration `20ns` samples at every 1ns.

As a result, our implementation has the following conventions

1. We reject waveform declarations with initializer array expressions, like `waveform wf = {...}`. 
2. There's no need for any kind of variadic extension to the language. The equivalent, `discrete(20ns, sample1, ..., sampleN)` is not supported, and not needed.

As a result I will likely be closing this MR, but I do think the OpenPulse spec stands to benefit from reconsidering what is allowed in a waveform declaration. The conventions I mention above reflect our own needs, where timing information is required early in the compilation pipeline (before we know the frames / channels on which each waveform is played), but these conventions may be suitable for others as well.

hodgestar:
I'm very much in favour of the spirit of this change to the OpenPulse spec.

Describing a waveform by a list of samples is often awkward. It requires knowing the sampling rate of the hardware playing the samples, which is for the most part a detail left to the compiler.

Perhaps as importantly it leaves unspecified what the value of the waveform between the sample points should be. Both the control and the actual physics of the QPU are real time processes, so the value of the waveform between sample points is important. In particular, the spectrum of the output signal can very significant, and the current sampled waveforms don't specify it at all.

Even when only simulating the hardware, it can be important for solvers to know the value of the output signal between sampling points.

Having a small list of well defined interpolation functions that take as input sample points and a duration would both simplify the specification and describe the requested output waveform better.

Rather than variadic arguments, I quite like the simpler proposal to just pass the sample points as a single array argument, as in:

```
waveform wf = discrete(20ns, wf_data);
```

There are quite a variety of interpolation functions (e.g. https://docs.scipy.org/doc/scipy/reference/interpolate.html) but it would still be nice to perhaps to pick one or two that users of OpenPulse can rely on and that implementers can be expected to implement.

On a slightly separate note, it would also be nice to recommend a bit more clearly which waveform shape functions implementers are expected to implement.

braised-babbage:
This specific PR can be closed -- as mentioned in the comments above, I'm no longer in favor of the changes here, and leaving this as a lingering draft can cause confusion.

PR #418, opened 2022-12-07
Allow DurationType as scalar type for arrays

braised-babbage:
### Summary

This PR allows for array types with duration scalars.

I don't know whether the previous omission of `DurationType` was intentional or not. However,
1. There does not appear to be any discussion in [the spec](https://openqasm.com/language/types.html#arrays) explicitly restricting the set of scalar types. So I think this should be syntactically valid, even if particular implementors do not currently support it.
2. We  allow `for` loops over durations, e.g. `for duration delta in {0us, 1.0us, 2.0us, 3.0us} { ... }`. Currently the parser handles via a special rule for set expressions, but [the spec indicates](https://openqasm.com/language/classical.html#for-loops)  that this is array-literal syntax. There's no reason, in principle, why we should accept this but reject the equivalent where a user declares a duration array and loops over the index range.

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -704,6 +704,12 @@ can be accessed, using the following general syntax:
    multiDim[0, 0] = 0.0; // multiDim == {{0.0, 1.2}, {2.1, 2.2}, {3.1, 3.2}}
    multiDim[-1, 1] = 0.0; // multiDim == {{0.0, 1.2}, {2.1, 2.2}, {3.1, 0.0}}
 
+The first argument to the ``array`` type constructor is the base type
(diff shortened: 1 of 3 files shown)


blakejohnson (approved):
Seems like a clear oversight.

jakelishman:
By the same argument, we should include `stretch` as an allowed data type as well, but I think that's relatively clearly never going to be valid in any implementation.  Perhaps we can update the text of the spec to explicitly spell out the allowed types (definitely excluding `stretch`), while we're at it?

braised-babbage:
@jakelishman I've updated the spec to list allowed and disallowed array base types.

blakejohnson (approved):

PR #426, opened 2022-12-14
Cite Wikipedia as external reference on Unicode

### LINKED ISSUE #425: Choose a different external link to unicode character categories
Currently, the link to a reference on unicode character categories
https://github.com/openqasm/openqasm/blob/38c4138e0ea29ef73f121e517f53943a28ef0689/source/bibliography.bib#L390-L395
is very spammy and full of distracting quasi-duplicitous adverts, but has little content.

Another reference would be better. Wikipedia for example.

jlapeyre:
The previously linked site had a very large number of intrusive advertisements and little information.

Closes #425

### Summary

Use a better reference for the external link to a reference on unicode characters.

### Details and comments

Use wikipedia. Use wikipedias page-citation generator.

### DIFF
--- a/source/bibliography.bib
+++ b/source/bibliography.bib
@@ -387,12 +387,15 @@
   publisher={Addison-Wesley Reading}
 }
 
-@misc{noauthorUnicodeNodate,
-	title = {Unicode {Character} {Categories}},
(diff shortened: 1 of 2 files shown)


blakejohnson (approved):
Sure

PR #428, opened 2022-12-14
In scan for TimeUnit fragment allow two unicode representations of mu

### LINKED ISSUE #422: Ambiguity with character code for microsecond
### OpenQASM version:

bcd53484

### Python version:

3.10.5

### Operating system:

fedora

The lexer recognizes `chr(181) + 's'` as a `TimeUnit` representing microsecond.
The lexer does not recognize `chr(956) + 's'` as any `TimeUnit`.

However, backslash + mu [TAB] in ipython produces `chr(956)`. (The same in Julia).
The qss parser also uses`chr(956)` in the `TimeUnit` for microsecond.

Try to parse first `1µ`, then `1μ`. The first works, the second throws an error.

That's a good question.

Perhaps allow both characters to represent the character mu in microsecond.

jlapeyre:
Previously only one unicode character for mu was recognized as the literal for microsecond. Now two are recognized. The new one appears frequently in Python. It also is used in qss-qasm.

Add tests for the three ways to specify microsecond.

Closes #422

### Summary

Recognize two different characters representing mu in the Lexer. Write them as unicode literals in order to remove ambiguity, since the glyphs are typically indistinguishable.

### DIFF
--- a/source/grammar/qasm3Lexer.g4
+++ b/source/grammar/qasm3Lexer.g4
@@ -147,7 +147,8 @@ FloatLiteral:
     // 123.456, 123. or 145.32e+1_00
     | DecimalIntegerLiteral DOT DecimalIntegerLiteral? FloatLiteralExponent?;
 
-fragment TimeUnit: 'dt' | 'ns' | 'us' | 'µs' | 'ms' | 's';
+// Each of the two unicode literals represent the greek letter mu.
+fragment TimeUnit: 'dt' | 'ns' | 'us' | '\u{00B5}s' | '\u{03BC}s' | 'ms' | 's';
 // represents explicit time value in SI or backend units
 TimingLiteral: (DecimalIntegerLiteral | FloatLiteral) TimeUnit;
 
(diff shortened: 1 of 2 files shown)


blakejohnson (approved):

jakelishman (commented):
Please can this come with an update to the text of the spec to make the explicit Unicode points clear as well?

k4rtik:
According to [Wikipedia](https://en.wikipedia.org/wiki/Mu_(letter)#Character_encodings), there seems to be a difference between `\mu` and `\micro` in Unicode. The lexer already supports `us` as an alternative for writing microseconds (if I am reading it correctly).

Another solution to this issue is to explicitly state that the character is `\micro`, i.e., U+00B5. Though this PR's solution does not seem problematic as the conflict between the two characters is going to be restricted to time unit parsing.

jakelishman:
From a pragmatic perspective, it'd be good to make sure that people who use TeX-to-unicode plugins for their text editors have an easy way to insert the correct symbol, since I'm guessing (I don't use one myself) that's the intended usecase for this over just using "us".  I would imagine that `\mu` will produce a mu-like letter, and `\micro` isn't a standard TeX macro, even though the `micro` codepoint is probably semantically more appropriate.

jlapeyre:
This should not be merged yet. See #440.

(Is there a don't merge label? I am unable to add labels)

blakejohnson:
To be revisited after #477.

blakejohnson:
Superseded by #477 .

PR #433, opened 2022-12-23
fix(spec/classical): incorrect backslash character

sno2:
Are you changing the specification? This PR needs to be approved by the TSC members.

Yes

Are you changing the grammar to adjust to the specification?

The grammar already matches this behavior.  See [line 133 in `source/grammar/qasm3Parser.g4`](https://github.com/openqasm/openqasm/blob/38c4138e0ea29ef73f121e517f53943a28ef0689/source/grammar/qasm3Parser.g4#L133).

### Summary

The character seems to be accidentally placed in this table entry because the Bitwise Or operator is used later in the reference without the backslash.

### Details and comments
Current Table View


blakejohnson (approved):
Thanks for catching this.

PR #435, opened 2022-12-24
feat(comments): clarify include path syntax

### LINKED ISSUE #434: Clarify Validity of Bit String-like paths in `include`s

Hello, this is the current text for the `include` statements:

> The statement include "filename"; continues parsing filename as if the contents of the file were inserted at the location of the include statement. This statement can only be used at the global scope.
> 
> ```openqasm
> // First non-comment is a version string
> OPENQASM 3.0;
> 
> include "stdgates.qasm";
> 
> // Rest of QASM program
> ```

However, this does not specify whether bit string-like strings should be valid.  Per the syntax described for bit strings, the following could be valid or invalid depending on if an implementation tokenized the path as a bit string and validates the path as a string literal.

```openqasm
include "0011";
```

Although, these are valid paths on Windows, for example.  Therefore, I think we should adjust the spec to state that any sort of string is allowed for this value.

sno2:
Close #434 

The current definition of include paths may not allow for paths that are valid bit strings (e.g. `001`). This updates the specification and reference ANTLR parser to allow such cases as valid.

✅ I have added the tests to cover my changes.
✅ I have updated the documentation accordingly.
✅ I have read the CONTRIBUTING document.

Are you changing the specification? This PR needs to be approved by the TSC members.

Yes

Are you changing the grammar to adjust to the specification?

The specification is the source of truth and any grammar updates should also coincide with accepted specification updates in the same PR.

Yes

See #434 for summary, details, and comments.


blakejohnson (changes requested):
I think you have correctly identified a bug in the ANTLR lexer. However, the spec already seems clear enough that `include` statements should accept any valid filename in the following string. @jakelishman wanted to suggest an alternative fix.

jakelishman (approved):
I've pushed up my changes - the principal difference is that we handle this within the lexer rather than the parser.  This makes it simpler for direct consumers of the ANTLR-generated parse tree to handle the output, which is shown here by the `openqasm3` package having been automatically upgraded to pass the new tests I wrote for it, without any change to the structure of either the ANTLR parser (only its lexer) or the `openqasm3` ANTLR parse-tree visitor.

I also fixed the same bug in the targets of `defcalgrammar`, since it was subject to the same thing.

blakejohnson (approved):

blakejohnson:
@jakelishman's version looks good to me.

PR #437, opened 2022-12-31
Make `IndexedIdentifier` an `Expression`

braised-babbage:
### Summary

This makes the AST `IndexedIdentifier` a subclass of `Expression`.

### Details and comments

`IndexedIdentifier` are lvalue expressions (can be used on the left-hand side of an assignment) like `a[{1, 2, 3}][0:1, 0:1]`.  There is no apparent reason for this to not subclass `Expression`.

It's worth noting what this PR does not change: the grammar. The grammar has an [`expression` rule](https://github.com/openqasm/openqasm/blob/main/source/grammar/qasm3Parser.g4#L121) which does not include `indexedIdentifier`. However, the AST class hierarchy is already not 1-1 with the grammar (e.g. `ArrayLiteral` is an `Expression` in the AST but not part of the `expression` rule in the grammar), and I think this is fine. Grammars deal with concrete syntax & must satisfy certain constraints (e.g. for ANTLR) that are not relevant at the AST level.

### DIFF
--- a/source/openqasm/openqasm3/ast.py
+++ b/source/openqasm/openqasm3/ast.py
@@ -461,7 +461,7 @@ class IndexExpression(Expression):
 
 
 @dataclass
-class IndexedIdentifier(QASMNode):
+class IndexedIdentifier(Expression):
     """An indentifier with index operators, such that it can be used as an
     lvalue.  The list of indices is subsequent index brackets, so in::
 


jakelishman:
I think this may run the risk of making the AST harder to consume - there's already an r-value representation of general indexing in `IndexExpression`, which does subclass `Expression`.  To me, adding this seems to give another way for the same valid OpenQASM 3 program to parse into two different AST representations, which I think we should try and minimise (not sure if it's possible to entirely eliminate).  If I were thinking in a more dynamic-language perspective (e.g. Python or Ruby) where assignment can be done to a general expression, the subclass would make a lot of sense, but OQ3 is far more restrictive so having a grammar/AST-level l-value/r-value split in these does seem reasonable to me.

What benefits do you have in mind for making it subclass?

(As a historical note: where there's particular weirdnesses in the AST, they often have their root in the older version of the ANTLR grammar, before we did a complete rewrite in #351.  That's not to say we should keep them around if they're actively harmful, I mostly just bring it up for reference.  The old grammar used to try and enforce a lot of semantic restrictions in the ANTLR-generated components, and one of the aims of the rewrite was to push more semantic concerns into custom human-written analysis passes.)

braised-babbage:
There's no particular motivation for this PR beyond noticing that the current AST hierarchy makes some type relations nonobvious (e.g. I would have naively guessed that an lvalue is a valid `Expression`, as this is how these things often work). Still, I am sympathetic to your concerns about downstream impact, since this would be a breaking change. It's not something I feel particularly strongly about, so I will withdraw the PR.

PR #439, opened 2023-01-13
Update governance document with details of new TSC

jakelishman:
### Summary

This contains the election results from the election ending on 2023-01-13.

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -18,8 +18,9 @@ This governance structure is scoped to just OpenQASM itself and not Qiskit as a 
 
 ## Structures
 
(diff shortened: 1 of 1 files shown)


jwoehr:
It's probably a good opportunity to answer the question embedded in the document, "How do we stagger the initial terms of the TSC?", since now the TSC has performed one iteration of staggering the terms.

jakelishman:
Good spot, thanks Jack - I just removed the question, since even asking it is obsolete now anyway.

blakejohnson (approved):

PR #441, opened 2023-01-27
Add AST node type HardwareQubitIdentifier

jlapeyre:
#### Before this PR
The ANTLR lexer and ANTLR parser distinguish between two kinds of identifiers with two tokens `Identifier` and `HardwareQubit`.

However before this PR, both of these tokens are merged into a single `ast` node, `ast.Identifier`, in the openqasm3 module. For example:
https://github.com/openqasm/openqasm/blob/7c98e1b2e88a033be247c1edb1926d048f18ffb9/source/openqasm/openqasm3/parser.py#L634-L635

https://github.com/openqasm/openqasm/blob/7c98e1b2e88a033be247c1edb1926d048f18ffb9/source/openqasm/openqasm3/parser.py#L789-L792

https://github.com/openqasm/openqasm/blob/7c98e1b2e88a033be247c1edb1926d048f18ffb9/source/openqasm/openqasm3/parser.py#L862-L865

#### Change introduced by PR

This PR instead associates `qasm3Parser.HardwareQubit` with a new type of node `ast.HardwareQubitIdentifer`. As before `qasm3Parser.Identifier` is associated with `ast.Identifier`.

#### Why

Before this PR, this semantic distinction must be made immediately by any downstream consumer in order to do anything useful. This is done by checking if the property `Identifier.name` begins with the character `$`.

Instead the semantic distinction should be done as early as possible without losing flexibility.

### DIFF
--- a/source/openqasm/openqasm3/ast.py
+++ b/source/openqasm/openqasm3/ast.py
@@ -59,6 +59,7 @@ __all__ = [
     "ForInLoop",
     "FunctionCall",
     "GateModifierName",
+    "HardwareQubitIdentifier",
     "IODeclaration",
     "IOKeyword",
     "Identifier",
@@ -247,6 +248,20 @@ class Identifier(Expression):
     Example::
 
         q1
+
+    """
+
+    name: str
+
+
+@dataclass
+class HardwareQubitIdentifier(Expression):
+    """
+    An identifier referencing a hardware (physical) qubit.
+
+    Example::
+
+        $0
 
     """
 
@@ -510,7 +525,7 @@ class QuantumGate(QuantumStatement):
     modifiers: List[QuantumGateModifier]
     name: Identifier
     arguments: List[Expression]
-    qubits: List[Union[IndexedIdentifier, Identifier]]
+    qubits: List[Union[IndexedIdentifier, Identifier, HardwareQubitIdentifier]]
     duration: Optional[Expression] = None
 
 
@@ -551,7 +566,7 @@ class QuantumPhase(QuantumStatement):
 
     modifiers: List[QuantumGateModifier]
     argument: Expression
-    qubits: List[Union[IndexedIdentifier, Identifier]]
+    qubits: List[Union[IndexedIdentifier, Identifier, HardwareQubitIdentifier]]
 
 
 # Not a full expression because it can only be used in limited contexts.
@@ -565,7 +580,7 @@ class QuantumMeasurement(QASMNode):
         measure q;
     """
 
-    qubit: Union[IndexedIdentifier, Identifier]
+    qubit: Union[IndexedIdentifier, Identifier, HardwareQubitIdentifier]
 
 
 # Note that this is not a QuantumStatement because it involves access to
@@ -604,7 +619,7 @@ class QuantumReset(QuantumStatement):
         reset q;
     """
 
-    qubits: Union[IndexedIdentifier, Identifier]
+    qubits: Union[IndexedIdentifier, Identifier, HardwareQubitIdentifier]
 
 
 @dataclass
@@ -1003,7 +1018,7 @@ class DelayInstruction(QuantumStatement):
     """
 
     duration: Expression
-    qubits: List[Union[IndexedIdentifier, Identifier]]
+    qubits: List[Union[IndexedIdentifier, Identifier, HardwareQubitIdentifier]]
 
 
 @dataclass
@@ -1056,7 +1071,7 @@ class AliasStatement(Statement):
     """
 
     target: Identifier
-    value: Union[Identifier, Concatenation]
+    value: Union[Identifier, Concatenation, HardwareQubitIdentifier]
 
 
 @dataclass
--- a/source/openqasm/openqasm3/parser.py
+++ b/source/openqasm/openqasm3/parser.py
@@ -129,6 +129,11 @@ def _visit_identifier(identifier: TerminalNode):
     return add_span(ast.Identifier(identifier.getText()), get_span(identifier))
 
 
+def _visit_hardware_qubit_identifier(hardware_qubit_identifier: TerminalNode):
+    return add_span(ast.HardwareQubitIdentifier(hardware_qubit_identifier.getText()),
+                    get_span(hardware_qubit_identifier))
+
+
 def _raise_from_context(ctx: ParserRuleContext, message: str):
     raise QASM3ParsingError(f"L{ctx.start.line}:C{ctx.start.column}: {message}")
 
@@ -632,7 +637,7 @@ class QASMNodeVisitor(qasm3ParserVisitor):
                 unit = ast.TimeUnit["dt"]
             return ast.DurationLiteral(value=float(value), unit=unit)
         if ctx.HardwareQubit():
-            return ast.Identifier(ctx.HardwareQubit().getText())
+            return ast.HardwareQubitIdentifier(ctx.HardwareQubit().getText())
         raise _raise_from_context(ctx, "unknown literal type")
 
     @span
@@ -788,7 +793,7 @@ class QASMNodeVisitor(qasm3ParserVisitor):
     @span
     def visitGateOperand(self, ctx: qasm3Parser.GateOperandContext):
         if ctx.HardwareQubit():
-            return ast.Identifier(name=ctx.getText())
+            return ast.HardwareQubitIdentifier(name=ctx.getText())
         return self.visit(ctx.indexedIdentifier())
 
     @span
@@ -861,7 +866,7 @@ class QASMNodeVisitor(qasm3ParserVisitor):
     @span
     def visitDefcalOperand(self, ctx: qasm3Parser.DefcalOperandContext):
         if ctx.HardwareQubit():
-            return ast.Identifier(ctx.HardwareQubit().getText())
+            return ast.HardwareQubitIdentifier(ctx.HardwareQubit().getText())
         return _visit_identifier(ctx.Identifier())
 
     def visitStatementOrScope(
--- a/source/openqasm/openqasm3/printer.py
+++ b/source/openqasm/openqasm3/printer.py
@@ -301,6 +301,9 @@ class Printer(QASMVisitor[PrinterState]):
     def visit_Identifier(self, node: ast.Identifier, context: PrinterState) -> None:
         self.stream.write(node.name)
 
+    def visit_HardwareQubitIdentifier(self, node: ast.HardwareQubitIdentifier, context: PrinterState) -> None:
+        self.stream.write(node.name)
+
     def visit_UnaryExpression(self, node: ast.UnaryExpression, context: PrinterState) -> None:
         self.stream.write(node.op.name)
         if properties.precedence(node) >= properties.precedence(node.expression):
--- a/source/openqasm/openqasm3/properties.py
+++ b/source/openqasm/openqasm3/properties.py
@@ -15,6 +15,7 @@ _PRECEDENCE_TABLE = {
     # Identifiers/literals are the top, since you never need to put brackets
     # around a single literal or identifier.
     ast.Identifier: 14,
+    ast.HardwareQubitIdentifier: 14,
     ast.BitstringLiteral: 14,
     ast.BooleanLiteral: 14,
     ast.DurationLiteral: 14,


jakelishman:
I'm not certain this is the right way to go - I think it could be, but it depends on whether you think about this as being a syntactical distinction or a typing one.  I personally think of the idea of hardware qubits as just being a typing distinction, where the `$` introductory sigil is needed because these symbols are implicitly defined, and it supplies the type.

In that view, I'm not certain that making a distinction in the class of the AST node makes sense; more properly (to me), it would be attached as the typing information to the `Identifier` node itself.  This AST doesn't include typing information for _any_ node right now.  I'm not clear that separating this as a new class makes any difference to consumers; they still have to verify the typing information when consuming the tree, but now we have a split between two different types of identifier classes that need to be made to interact with whatever symbol table setup they have.

For example: we're not suggesting making a "GateOrSubroutineIdentifier" node, even though that those are other identifiers that have specific constraints on where they can be semantically valid.  To push it further, we won't have a `VirtualQubitIdentifier` either, because that's typing information (which to me is the same thing that's going on with `HardwareQubitIdentifier`).

jlapeyre:
The motivation for this PR is that things like supporting physical qubits when converting OC3 to Qiskit (https://github.com/Qiskit/qiskit-qasm3-import/pull/2) would be cleaner and easier if we made some kind of semantic distinction.

Is there a reasonable scenario where you'd want to undo this distinction and merge the two elements again into `Identifier`? I didn't try hard to come up with an example. If the answer is no, then I think the the distinction should be preserved somehow.

There is good reason to call these elements literals rather than identifiers. If we want to design for the eventuality that OC becomes broadly adopted, I don't think it's a good idea to cater too closely to the needs of the low levels of the hardware stack. But I'd be interested to see the argument for not treating them as literals. I'm not committed to any interpretation at present.

In fact, the qss-qasm parser currently does [this](https://github.com/jlapeyre/openqasm/blob/b0df0c9a9cfa7678ca01c1c054f685da2efb68b2/source/grammar/differences-from-spec.md?plain=1#L39)
```
Larger set of allowed identifiers for physical qubits: `'$' (([a-zA-Z_]* [0-9]+) | [a-zA-Z_]+)`
```
([Here is a list of all difference wrt the spec](https://github.com/jlapeyre/openqasm/blob/ibmqasm/source/grammar/differences-from-spec.md))


It would be worth looking at how other languages deal with things like this. I can't imagine they don't make the semantic distinction at the level of element type in ASTs or IRs. I think most people would agree that parsing the identifier after it is part of an AST to find out  what it means is not the right approach.

> For example: we're not suggesting making a "GateOrSubroutineIdentifier" node

But this is different. All of these other identifiers have the same syntax.

jakelishman:
> There is good reason to call these elements literals rather than identifiers.

That depends on your perspective - I mentioned that I started off thinking of them as literals like you, but I do very much buy the argument Thomas and his team make in the linked PR about why they make sense as declarable identifiers, and they're one of the few teams who have an actual implementation of OpenQASM 3.  Treating it as an identifier in an AST with complete typing- and declaration-site (or lack thereof) information lets us satisfy the needs of both schools anyway, so I would definitely lean in favour of that.

> Is there a reasonable scenario where you'd want to undo this distinction and merge the two elements again into `Identifier`?

Let me be careful here, because there's two things at play.  For the syntactic difference, _yes_ we should erase it, because that's the difference between a parse tree (what ANTLR generates) and an AST (what the package generates, and what downstream consumers in the compiler stack want). As I understand it (could well be wrong), `qss-qasm` has this same split - the distinction you mention is during lexing/parsing, but if you look at an AST node (say `ASTReset`), it's built from an `ASTIdentifier` which carries type information.

What we're throwing away here is the _typing_ information, which we shouldn't.  It so happens that in this particular case we get the typing information from the parse tree, but for every other `Identifier` we would read it semantically from the declaration site.  We actually have both pieces of information available and we "throw away" (perhaps better: don't calculate) both in this package.

My argument is that the proper way to do what you want (which _is_ a valid concern) is to expose type information on all `Identifier` nodes (actually all `Expression` nodes, and more besides).  This PR is a half measure: it retains some part of the parse tree, because that can technically be used to infer the typing information slightly easier, but isn't extensible to handling the rest of typing information.  My way gets you what you want (type information), and what I want (a more abstract tree).

steleman:
> > There is good reason to call these elements literals rather than identifiers.
> 
> That depends on your perspective - I mentioned that I started off thinking of them as literals like you, but I do very much buy the argument Thomas and his team make in the linked PR about why they make sense as declarable identifiers, and they're one of the few teams who have an actual implementation of OpenQASM 3. Treating it as an identifier in an AST with complete typing- and declaration-site (or lack thereof) information lets us satisfy the needs of both schools anyway, so I would definitely lean in favour of that.
> 
> > Is there a reasonable scenario where you'd want to undo this distinction and merge the two elements again into `Identifier`?
> 
> Let me be careful here, because there's two things at play. For the syntactic difference, _yes_ we should erase it, because that's the difference between a parse tree (what ANTLR generates) and an AST (what the package generates, and what downstream consumers in the compiler stack want).

I am fuzzy as to what _parse tree_ means.

`qss-qasm` does not generate, or emit, any _parse trees_. In fact, `qss-qasm` doesn't generate any trees of any kind. LALR(1) parsing is based on stacks, not trees.

`qss-qasm` emits an `AST` with an associated semantic analysis. Part of the semantic analysis is done during lexing, another part during parsing, and yet another part after the AST has been generated - during an as-of-yet unimplemented and subsequent semantic analysis pass that I am in the process of writing. 

> As I understand it (could well be wrong), `qss-qasm` has this same split - the distinction you mention is during lexing/parsing, but if you look at an AST node (say `ASTReset`), it's built from an `ASTIdentifier` which carries type information via the symbol table.

Nope. Not at all.

An `Identifier` does not denote, record, identify, or describe a `Type`.

`Identifiers` are just human-readable (and human-understandable) mnemonic devices invented for the sole purpose of making the programmer's life easier.

The one and only thing that an `Identifier` represents - in abstract - is an association between a `Type` and a `memory location`. An `Identifier` does not represent a `Type`, and `Types` exist independently of `Identifiers`. `Types` are defined by the Language Specification. `Identifiers` are invented by the programmer.

Some specific `Identifiers` may be promoted to `Reserved Keywords`. For example, `pi`. In that case, the `Reserved Keyword` contains, and describes, its `Type` information. That semantic information is described and conveyed by the Language Specification (read: _convention_), not by the `Identifier` itself.

For example (in C):

```
typedef struct _giraffeS {
  unsigned height;
  unsigned weight;
  unsigned length_of_neck;
  unsigned number_of_splotches;
} giraffe_t;

// in some other translation unit:

giraffe_t Rosie = { 48, 1000, 18, 53 };
int Binky = 3;

// ...
```

The `Identifier` `Rosie` doesn't tell us anything about its `Type` (`giraffe_t`) anymore than `Binky` does (`int`). In fact, the following two declarations:

```
giraffe_t TZAlKC2P0zXaBjbSwdp = { 48, 1000, 18, 53 };
int KlrmMgA1g5XfWGcg9hn = 3;
```

are functionally equivalent to the previous two.

For all we care, the compiler is free to throw away the `Identifier` invented by the programmer, and use some other means of representing the association between the `Type` and the `memory location`.

> `ASTReset`, is built from an `ASTIdentifier` which carries type information via the symbol table.

Nope.

`ASTResetNode` is not _built from an `ASTIdentifierNode` which carries its type information_. The `Type` of `ASTResetNode` is built from itself, and is contained within, and is determined by, the `ASTResetNode` itself. Nothing else.

The fact that `ASTResetNode` contains an `ASTTypeEnum` that records  its `Type` is yet another convenient device artifact, invented for two purposes:

- making the compiler developers' lives easier.
- avoidance of dealing with C++'s `typeinfo` subsystem, which is much slower than `enum` (`integer`) arithmetic (`typeinfo` involves function calls, and they add up in terms of CPU cost). 

The fact that the `ASTResetNode`'s `ASTIdentifierNode` contains a pointer into its Symbol Table location is, again, another convenient device, again _invented solely for making the compiler developer's life easier_.

I could have written `qss-qasm`'s AST generation in a completely different way, and without these helpful artifacts. Using a Symbol Table was entirely optional. Using `ASTTypeEnum`s instead of `C++ typeinfo` was also entirely optional.

For example, I could have just recorded memory addresses and their associated `Type` information, as provided by `typeinfo`. Functionally, this hypothetical implementation would be 100% equivalent to the current one. It's just that it would be much more difficult to deal with for downstream consumers of the AST, and _much_ slower.

jakelishman:
Thanks for the detailed explanation Stefan.  I think there's some minor terminology differences we've used, but the architecture I'm advocating for is largely the same as `qss-qasm` - I was using it as an example of what I think this PR should aim for.

When I was saying "type information", I was meaning the information represented by `ASTTypeEnum`. I recognise that's not the only equivalent way to store that information (since this PR gives another), but I agree it's a very convenient one, and it's the style I was advocating too (albeit coming from much less experience than you).

In the last sentence you flagged, I meant that I thought you could use the `ASTIdentifier` node to access the type information attached to that memory location using the symbol table. I might not have got the precise details right, but the main point was that in the `qss-qasm` AST, you have one C++ class for all identifiers, regardless of the type attached to the memory location for each. I'm saying that that's the same architecture I would want here too.

I don't believe I said that `qss-qasm` generates a parse tree.

jlapeyre:
> two different types of identifier classes that need to be made to interact with whatever symbol table

We have two types of identifiers, represented by different regexes. This is different than two identifiers that satisfy a single regex, but that are assigned different types based on the context they occur in. In the former case, it looks practical to me to preserve the distinction between these different types of identifiers. But I agree that two types of identifier sharing a symbol table makes the design less clean. Having a separate table for hardware qubits might be a good solution.

It looks practical and clean to me to make the distinction in the lexer between identifiers for hardware qubits (or literals, or whatever semantics you attach later) and identifiers for everything else, and then preserve that distinction in all subsequent stages. (I'm eliding that we determine the semantics of some identifiers and they are consumed in going from the ANTLR parse tree to the AST)  If we don't want to do this, then perhaps the next cleanest thing would be to replace this

https://github.com/openqasm/openqasm/blob/7c98e1b2e88a033be247c1edb1926d048f18ffb9/source/grammar/qasm3Lexer.g4#L138-L139

with
```antlr
Identifier: (FirstIdCharacter GeneralIdCharacter*) |'$' [0-9]+; 
```
And hardware qubits would be recognized at a later stage with a regex that separates these two rules.

The mixed approach, which is the approach we have now, makes the least sense to me. That is:

#### Step 1  Make a distinction between identifiers and hardware qubits

here:
https://github.com/openqasm/openqasm/blob/7c98e1b2e88a033be247c1edb1926d048f18ffb9/source/grammar/qasm3Lexer.g4#L138-L139

https://github.com/openqasm/openqasm/blob/7c98e1b2e88a033be247c1edb1926d048f18ffb9/source/grammar/qasm3Parser.g4#L208-L209

#### Step 2 Remove the distinction between identifiers and hardware qubits

Shown in the code snippets linked in the first comment in this PR. Both `ctx.Identifer` and `ctx.HardwareQubit` are mapped to `ast.Identifer`. In the AST, hardware qubits are no longer recognized. This is what I meant by "throw away". We deliberately erase information. This would be ok if we didn't need it. But, in the next step we actually recompute it...

#### Step 3 Make again the distinction between identifiers representing virtual qubits and hardware qubits

This happens when translating the AST to something else, for example Qiskit circuits. We use the same regex that the lexer used in Step 1.

I think a better design would not require using the same regex twice on the same data.

jakelishman:
John, I think you've missed the main thrust of [my response to you](https://github.com/openqasm/openqasm/pull/441#issuecomment-1406757435).  I am not arguing to maintain the status quo.

If we wanted `HardwareQubit` to be a literal value, like `FloatLiteral` or `IntegerLiteral`, this PR would be perfect.  However, the discussion in #417 explains why there's good reason to want to treat them like declarable constants instead.  If so, then I think it's easier for consumers if it is treated in the AST like any other declarable thing (e.g. looks like `ast.Identifier`), _provided we also make the inferred type information available for identifier expressions_.

What I'm suggesting instead is that we add a type resolver to this package.  For example, in Python land [`astroid` exposes `lookup` on `Name`](https://github.com/PyCQA/astroid/blob/a31da6fe388ad98cca51b734e3201d78d717c68c/astroid/nodes/node_classes.py#L350-L367) that returns all assignments to that Python name (as defined by the scoping rules) and [`mypy`'s `Var` node in its typed AST has a direct `type` field](https://github.com/python/mypy/blob/7c14ebae6ca5d6ec39600e122b58a62afaa3ab02/mypy/nodes.py#L935-L977) that's filled during typing analysis.  In C land, [`clang`'s AST uses (e.g.) `Expr::getType()`](https://github.com/llvm/llvm-project/blob/df76ff98e8e68ef00e61574ce6ac688e6e12b9df/clang/include/clang/AST/Expr.h#L142) to expose the information (though I can't pretend to be anything more than a beginner on LLVM's source).  `qiskit-qasm3-import` has to do all its type resolution itself, but I never wanted that.  It was just a workaround because this package doesn't do it.  We still want the split in the ANTLR grammars, because that's how _this package_ infers that particular piece of typing information, and can then re-expose it in the more abstract form. 

---

I should highlight that this package uses the term `Identifier` atypically, which may well be causing confusion.  Here it's an AST node in the `Expression` tree, more akin to `Name` in `astroid` or `DeclRefExpr` in `clang`, except it doesn't come with any of the extra information those two ASTs provide.  This package doesn't have anything like `clang`'s `IdentifierInfo`, though it will likely need to gain something similar.

jlapeyre:
> What I'm suggesting instead is that we add a type resolver to this package.

Got it. I have some questions. But perhaps that belongs in an issue devoted to adding a type resolver.

levbishop:
Closed based on TSC discussion 2023-09-13, in order to leave options open for other directions eg changing the spec to maybe treat hardware/virtual qubitness as convention rather than  language syntax.

PR #447, opened 2023-02-16
Black fixes for #418

mbhealy:
### Summary
Merging of #418 seems to have caused the black linter to fail. This PR is the result of running black on the AST generator.
https://github.com/openqasm/openqasm/actions/runs/4187227948


mbhealy:
Apparently there are also pytest failures. Not sure why the previous PR did not trigger these.

jakelishman (approved):
Thanks Michael - I'll look into why the other PR didn't cause CI failures.  It might be because our `main` branch-protection rules aren't set strict enough on requiring PRs to be fully up-to-date with `main`.

Issue #446, opened 2023-02-16
[arrays]: The OQ3 Spec mentions `references` but does not define what they are


steleman:
### What is the expected enhancement?

The OpenQASM3 Spec for Arrays states the following:

```
All array parameters are passed as references [ ... ]
```

The `reference` Type is not defined anywhere in the current Spec.

For clarity and completeness, we should define the `reference` Type, and its properties.

jwoehr:
@steleman can you provide a model PR that would allow the spec to indicate more rigorously that array parameters are passed by reference as we all more or less understand the sense of things?

steleman:
```
Arrays are a special and degenerate type of a fixed-size tensor. Unlike tensors,
arrays cannot be resized or reshaped.

The size of an array is constant and immutable, and is recorded once, at
declaration time.

Arrays allocate a contiguous block of memory of suitable size and alignment to
accommodate the storage of its elements.

If the array declaration is direct-initialized via initializer list, its elements
are initialized to the values provided by the initializer list.  Otherwise, the
memory allocated is not initialized, and is assumed to be garbage.

Partial initialization via initializer list is allowed - i.e. the initializer
list is incomplete, and does not provide values for all the elements of the array.
In this case, the elements that are not initialized via initializer list are not
initialized, and their values are assumed to be garbage.

As a special case, a direct initializer list declaration of the form:

  array[<type>, <size>] <identifier> = { <X> };

will initialize all the elements of the array to <X>, where <X> is of a type
suitable for initializing an element of the array, and if and only if the array
element type supports direct initialization from such a type.

It is unspecified whether the memory allocation of an array is of static storage
duration, or it is obtained via dynamic allocation. This property is implementation
dependent.

If the array memory is obtained via dynamic allocation, the implementation is
responsible for re-claiming the memory when the array object goes out of scope
(garbage collection).

Arrays may be passed as parameters or arguments to functions.

When passed as argument to a function, an array shall not cause a copy spill on
the stack. Several mechanisms are available to implement this constraint:

  - Use the C model: an array implicitly decays to a pointer to its first
    element, and only this pointer is passed as argument. This pointer may
    cause a copy spill on the stack. This does not imply that pointer types
    are normative to the language. This implicit conversion from array to
    pointer and back may remain opaque to the user.

  - Use the Fortran model: every array is described by an Array Descriptor.
    An Array Descriptor is an opaque type that contains descriptive metadata
    describing its corresponding array, including size and memory address
[...]

PR #450, opened 2023-02-21
Added syntax derived from Stefan Teleman's #446

jwoehr:

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -704,11 +704,50 @@ can be accessed, using the following general syntax:
    multiDim[0, 0] = 0.0; // multiDim == {{0.0, 1.2}, {2.1, 2.2}, {3.1, 3.2}}
    multiDim[-1, 1] = 0.0; // multiDim == {{0.0, 1.2}, {2.1, 2.2}, {3.1, 0.0}}
 
-The first argument to the ``array`` type constructor is the base type
+The first argument to the ``array`` declaration is the base type
 of the array. The supported classical types include various sizes of ``bit``,
 ``int``, ``uint``, ``float``, ``complex``, and ``angle``, as well as
 ``bool`` and ``duration``. Note that ``stretch`` is not a valid array
 base type.
+
+Arrays cannot be resized or reshaped. Arrays are statically typed, and cannot
+implicitly convert to or from any other type.
+
+The size of an array is constant and immutable, and is recorded once, at
+declaration time.
+
+Arrays allocate a contiguous block of memory of suitable size and alignment to
+accommodate the storage of its elements.
+
+If the array declaration is direct-initialized via initializer list, its elements
+are initialized to the values provided by the initializer list.  Otherwise, the
+memory allocated is not initialized, and that memory's content is undefined.
+
+Partial initialization via initializer list is allowed - i.e. the initializer
+list is incomplete, and does not provide values for all the elements of the array.
+In this case, the elements that are not initialized via initializer list are not
+initialized, and their values are undefined.
+
+As a special case, a direct initializer list declaration of the form:
+
+ ``array[<type>, <size>] <identifier> = { <X> };``
+
+will initialize all the elements of the array to <X>, where <X> is of a type
+suitable for initializing an element of the array, and if and only if the array
+element type supports direct initialization from such a type.
+
+It is unspecified whether the memory allocation of an array is of static storage
+duration, or it is obtained via dynamic allocation. This property is implementation
+dependent.
+
+If the array memory is obtained via dynamic allocation, the implementation is
+responsible for reclaiming the memory when the array object goes out of scope
+(garbage collection).
+
+Arrays may be passed as parameters or arguments to functions.
+
+When passed as argument to a function, an array shall not cause a copy spill on
+the stack.
 
 Arrays *cannot* be declared inside the body of a function or gate. All arrays
 *must* be declared within the global scope of the program.
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -749,8 +749,7 @@ responsible for reclaiming the memory when the array object goes out of scope
 
 Arrays may be passed as parameters or arguments to functions.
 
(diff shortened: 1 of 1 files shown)


source/language/types.rst
```diff
@@ -704,12 +704,51 @@ can be accessed, using the following general syntax:
...
+As a special case, a direct initializer list declaration of the form:
+
+ ``array[<type>, <size>] <identifier> = { <X> };``
+
+will initialize all the elements of the array to <X>, where <X> is of a type
+suitable for initializing an element of the array, and if and only if the array
+element type supports direct initialization from such a type.
+
+It is unspecified whether the memory allocation of an array is of static storage
+duration, or it is obtained via dynamic allocation. This property is implementation
```
blakejohnson, line 740: I'm confused by the word "duration" here. Would it be okay to say "it is unspecified whether the memory allocation of the an array is static or obtained via dynamic allocation"?

blakejohnson (dismissed):
Other than one bit of language confusion, I am good with these changes.

source/language/types.rst
jwoehr: Added language per this discussion

source/language/types.rst
```diff
@@ -704,12 +704,54 @@ can be accessed, using the following general syntax:
...
+
+ ``array[<type>, <size>] <identifier> = { <X> };``
+
+will initialize all the elements of the array to <X>, where <X> is of a type
+suitable for initializing an element of the array, and if and only if the array
+element type supports direct initialization from such a type.
+
+It is unspecified whether the memory allocation of an array is of static storage
+duration, (meaning the allocation is alive for as long as the object is in scope and
+gets automatically reclaimed by the operating system when the object goes out of scope)
```
blakejohnson, line 741: Can we go with 'runtime' rather than 'operating system'? 
```suggestion
gets automatically reclaimed by the runtime when the object goes out of scope)
```

blakejohnson (approved):

source/language/types.rst
```diff
@@ -704,12 +704,54 @@ can be accessed, using the following general syntax:
...
+It is unspecified whether the memory allocation of an array is of static storage
+duration, (meaning the allocation is alive for as long as the object is in scope and
+gets automatically reclaimed by the runtime when the object goes out of scope)
+or it is obtained via dynamic allocation (meaning the allocation persists through the lifetime
+of the program, unless the program itself releases it back to the operating system).
+This property is implementation dependent.
+
+If the array memory is obtained via dynamic allocation, the implementation is
+responsible for reclaiming the memory when the array object goes out of scope
+(garbage collection).
```
levbishop, line 748: Since all arrays are declared in global scope maybe the real question is whether the compiler chooses to do lifetime analysis and only allocate the array (some time before) when it is first used and deallocate (some time after) the last usage.

source/language/types.rst
```diff
@@ -704,12 +704,54 @@ can be accessed, using the following general syntax:
...
+
+Arrays may be passed as parameters or arguments to functions.
+
+When passed as argument to a function, an array shall not cause a copy spill on
+the stack.
```
levbishop, line 753:
```suggestion
When an array, or slice of an array, is passed as argument to a function, the underlying array storage is referenced rather than copied.
```

source/language/types.rst
```diff
@@ -704,12 +704,54 @@ can be accessed, using the following general syntax:
...
+memory allocated is not initialized, and that memory's content is undefined.
+
+Partial initialization via initializer list is allowed - i.e. the initializer
+list is incomplete, and does not provide values for all the elements of the array.
+In this case, the elements that are not initialized via initializer list are not
+initialized, and their values are undefined.
+
+As a special case, a direct initializer list declaration of the form:
+
+ ``array[<type>, <size>] <identifier> = { <X> };``
```
levbishop, line 733: I can see value in having syntax for initializing all elements to the same value, but this special case syntax that overlaps with the partial-initialization syntax seems unnecessary. (Especially since I imagine the most common form of partial initialization would be to initialize just the first element of say a list of accumulated deltas, which usage is precluded by this special case).

But maybe its ok since you can think of this as *both* partial initialization (where the unspecified elements happen to be initialized to `X`) *and* as constant initialization of the whole array. Assuming we don't care about the cost of initializing more than was needed maybe it's ok. Otherwise, maybe:
```suggestion
As a special case, a direct initializer list declaration of the form:

 ``array[<type>=<X>, <size>] <identifier>;``
```
might be another option for constant initialization of the whole array?

braised-babbage: I'll chime in and say that I'm not a fan of supporting both partial initialization and the `array[<type>, <size>] <identifier> = { <X> };` initialization above, for the reason that I think the semantics of partial initialization should be basically monotonic (more data specified the initializer list means more elements get initialized). Or to turn this around: it would come as a surprise to me (if I hadn't read the spec) that `array[int, 5] = {3,2}` has more undefined elements than `array[int, 5] = {3}`.

source/language/types.rst
```diff
@@ -704,12 +704,53 @@ can be accessed, using the following general syntax:
...
+will initialize all the elements of the array to <X>, where <X> is of a type
+suitable for initializing an element of the array, and if and only if the array
+element type supports direct initialization from such a type.
+
+It is unspecified whether the memory allocation of an array is of static storage
+duration, (meaning the allocation is alive for as long as the object is in scope and
+gets automatically reclaimed by the runtime when the object goes out of scope)
+or it is obtained via dynamic allocation (meaning the allocation persists through the lifetime
+of the program, unless the program itself releases it back to the operating system).
+This property is implementation dependent.
```
braised-babbage, line 744: It's not clear to me why this needs to be said. There's nothing here that I disagree with, but I just don't see how to operationalize it -- I don't know of any other part of the spec that talks about storage classes or memory management. It seems like a user just has no way of knowing the difference here, and so probably wouldn't care.

blakejohnson (approved):

PR #456, opened 2023-03-15
Added Row Major array representation requirement to types.rst.

steleman:
### Summary

The current OpenQASM3 Spec does not address the in-memory ordering representation of arrays.
This PR adds a one line sentence stating that OpenQASM3 arrays will be stored in Row Major order.

### Details and comments

Storage ordering of (N-Dimensional) arrays has a direct impact on program performance.  There are two ways arrays can be stored in memory: Row Major order and Column Major order:

https://en.wikipedia.org/wiki/Row-_and_column-major_order

Storing arrays in Row Major order provides significant advantages over Column Major order. It enables certain optimizations in the compiler that are not possible in Column Major order.

https://cs.stackexchange.com/questions/71985/row-major-vs-column-major-order-2d-arrays-access-in-programming-languages

https://www.wyzant.com/resources/answers/620254/what-are-the-benefits-of-row-major-order-over-column-major-order-when-proce

For example, Fortran stores arrays in Column Major order. Fortran programs that want to avail themselves of vectorization in `for` loops must convert their 2-dimensional arrays to Row Major order before such optimizations can be applied. This, of course, entails overhead as well, and in some cases makes the vectorization entirely unprofitable -- whatever performance gain may have been achieved by vectorization is canceled out by the overhead of the array transpose.

```
 INTEGER array(2, 3), result(3, 2)
 array = RESHAPE((/1, 2, 3, 4, 5, 6/), (/2, 3/))
 ! array is  1  3  5
 !                 2  4  6
 result = TRANSPOSE(array)
 ! result is 1  2
 !                  3  4
 !                  5  6
 END
```

For the most idiomatic loops generally being written in programs:

```
array[int, 10, 10] ai;
for int i in [0 : 9]  {
    for int j in [0 : 9] {
        // Do something useful with ai[i, j]
    }
}
````

the compiler can take advantage of hardware prefetch, and preload elements of the array in cache, improving performance.

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -743,6 +743,8 @@ the shape and type of the assigned value must match that of the reference.
 Arrays may be passed to subroutines and externs. For more details, see
 :any:`arrays-in-subroutines`.
 
+OpenQASM arrays shall be represented in Row Major order.
+
 Types related to timing
 -----------------------
 

### CODE REVIEW

**Comment by blakejohnson** [comment]
As mentioned in last TSC meeting, if we do proceed I would really like more than just the one sentence added to the spec.

As for whether we *should* pursue this, I have to admit to still being on the fence. I still think it is not unreasonable for a language to be predominantly row-major or column-major. However, I also see how making such a declaration for OpenQASM would define things which we have previously treated as an implementation detail.

It also seems unlikely to have significant performance impact in some near-term architectures, some of which don’t have caches but just memory at a uniform r/w latency. These same architectures don’t have vector instructions where you can process multiple array loads and/or stores per clock cycle.

So, I think where I am today is that while I am not against adding this to the spec some day, I am not yet convinced that we should add it now.

**Comment by levbishop** [comment]
It seems to me that details of memory layout likely go beyond row/column major. Eg in some cases array indices might best be used to address a NUMA node eg for a distributed architecture with a cpu per qubit. As for prefetch, its normal for hardware to recognize other strides than 1 https://people.freebsd.org/~lstewart/articles/cpumemory.pdf

For now I lean toward experimenting on specifying memory layout by using annotations on the array declarations and if there is consensus that this can be made interoperable then it could be standardized in the future. Example:
```openqasm
@ibm.arraystorage DENSE-ROWMAJOR
array[float[32], 3, 2] multiDim;
```

blakejohnson:
I'm not sure if we got to consensus, but at today's TSC meeting it seemed like we would not pursue this and instead remove statements about dynamic vs static allocation of arrays in the spec.

jwoehr:
Possibly call to vote at next TSC

bkavcaduhq:
I can not found any information of **tuple** type in openqasm language, How about the tuple type in quantum computer language language?

blakejohnson:
We are going to go with an async TSC vote on this issue. @openqasm/openqasm-tsc please respond to this message with 👍 or 👎 .

blakejohnson:
@bettinaheim @dorisraeli @levbishop please record your vote on the comment above.

jwoehr:
Closed TSC 4 nay to 2 abstentions (see above)

PR #457, opened 2023-03-15
Monkeypatch docutils to fix bibliography output

### LINKED ISSUE #455: Navigation sidebar missing from index page
### OpenQASM version:

Live

### Python version:

NA

### Operating system:

NA

The sidebar navigation links are not visible from the index page of https://openqasm.com. Other subpages show the correct location for the navigation sidebar.

Visit  https://openqasm.com

Sidebar navigation is visible when visiting  https://openqasm.com.

I have tracked this down to a bug in the docutils package, which is required by sphinxcontrib-bibtex and occurs because of a missing opening div that is followed by a mismatched closing div. 

Details are here: https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/309
A patch is available here: https://sourceforge.net/p/docutils/patches/195/
Link to patchfile: https://sourceforge.net/p/docutils/patches/195/attachment/0001-Fix-previous_sibling-in-case-index-is-0.patch

The patch was merged into docutils dev about 6 months ago, but I have no idea when a new version of docutils will be released. We could either try to patch docutils using the file above during publishing of the webpage or directly insert the missing line `<div role="list" class="citation-list">\n` into the built html page after the line `<div class="docutils container" id="id1">`.

jakelishman:
### Summary

This applies the patch discussed in gh-455 dynamically on import of `docutils`, in order to fix the broken HTML output from the `..  bibliography` directive. We constrain `docutils` during our build process to fix it to a version that is known-good with this change applied.  The constraint may be dropped when docutils updates, as the relevant patch is merged to trunk.

### Details and comments

Fix #455.

### DIFF
--- a/.github/workflows/build-live-version.yml
+++ b/.github/workflows/build-live-version.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install -c constraints.txt -r requirements.txt
 
       - name: Generate HTML
         run: make html
--- a/.github/workflows/publish-live-version.yml
+++ b/.github/workflows/publish-live-version.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install -c constraints.txt -r requirements.txt
 
       - name: Generate HTML
         run: make html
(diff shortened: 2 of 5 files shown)


jakelishman:
Michael's suggestion of patching docutils turned out to be the easiest to do.  I did it dynamically from within Python rather than manually applying a patch file to the docutils source tree because this way requires less fiddling to find the installed location to patch.

jakelishman:
Here's a screenshot taken from the build artifact of the CI on this PR:

Screenshot 2023-03-15 at 17 36 07

mbhealy (approved):
LGTM

levbishop (approved):

blakejohnson (approved):

Issue #458, opened 2023-03-29
[set expressions]: Documentation needs clarification 


nelimee:
### Issue: 

In the official documentation [Register concatenation and slicing](https://openqasm.com/language/types.html#register-concatenation-and-slicing), an `index set` is defined as
> An index set can be specified by a single integer (signed or unsigned), a comma-separated list of integers contained in braces {a,b,c,…}, or a range.

In the same page of the official documentation [Array concatenation and slicing](https://openqasm.com/language/types.html#array-concatenation-and-slicing), we can read
> Arrays can be sliced just like quantum registers using index sets.

So, if I am understanding this correctly, it means that the following code should be correct:
```
array[int[8], 3] arr = {2, 3, 4};
array[int[8], 2] arrRef = arr[{0, 2}];
```

In the current grammar [qasm3Parser.g4#L164](https://github.com/openqasm/openqasm/blob/a7e02b52c7634a73f5b7a42ac8960f12fd3ce25a/source/grammar/qasm3Parser.g4#L164), we can read this comment:
> // `setExpression` is only valid when being used as a single index: registers
> // can support it for creating aliases, but arrays cannot.

### Expected enhancement?

So there are 2 options:
1. The text in the documentation is right. In which case the comment in the grammar should be removed / changed and the limitation of `setExpression is only valid when being used as a single index` should probably be removed.
2. The grammar is right. In which case the text in the documentation should be improved to reflect the limitations of `setExpression`.

Personally, I would even make `setExpression` valid everywhere a `rangeExpression` is valid, making the following correct:
```
array[int[8], 3, 3] arr     = {{1, 2, 3}, {4, 5, 6}, {7, 8, 9}};
array[int[8], 2, 2] arrRef  = arr[{0, 2}, {1, 2}]; // {{2, 3}, {8, 9}}
array[int[8], 2]    arrRef2 = arr[0, {1, 2}];      // {8, 9}
```
but there might be some intricacies that I do not see yet.

PR #460, opened 2023-04-03
Fixing documentation with respect to array indexing

nelimee:
Main issue: #458 

### Summary
The `comma-separated list of integers contained in braces {a,b,c,…}` way of indexing should not be used for arrays are arrays are meant to be only indexed by single integers and sliced by regular (i.e., not arbitrary) indices. 

This was not cleat at all in the documentation. 
This PR tries to fix this shortcoming.

### Details and comments

The current fix is minimal. I feel like the documentation would benefit from a re-writing for this part. I can see two ways:
1. presenting the different methods of indexing (single integer, integer set, slice) first, and then specifying which method is valid for each array-like types.
2. starting by presenting the most restricted type (arrays) and extending them in the next sections by writing something like "ints are considered as arrays of bits" and "registers (quantum or classical) are considered as arrays of (qu)bits with the additional capability of being indexed by integers sets `{a,b,c,…}`.

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -909,7 +909,9 @@ should be explicitly declared and assigned the concatenation.
    subroutine_call(first ++ third) // forbidden
    subroutine_call(selfConcat) // allowed
 
-Arrays can be sliced just like quantum registers using index sets. Slicing uses
+Arrays can be sliced just like quantum registers using a range ``a:b:c`` 
+and can be indexed using an integer but cannot be indexed by a a comma-separated 
+list of integers contained in braces ``{a,b,c,…}``. Slicing uses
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
Thanks for this - I agree that this is what the array spec part meant, and that any implied "discrete-set indexing" was a mistake, and didn't match the intended behaviour.

nelimee:
Just wanting to check: is there anything I should do that is currently blocking the PR, or is it considered pending until someone actually merge it?

jakelishman:
There's nothing you need to do.  Technically this is a change to the specification (though really it's just a bugfix), so it just needs approval from the technical steering committee.

blakejohnson (approved):
LGTM

PR #463, opened 2023-04-05
Switch statement

steleman:
### Summary

Adding `switch` statement to the OpenQASM3 Grammar.

### Details and comments

Added Grammar definition and examples for the OpenQASM3 `switch` statement.

### DIFF
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -431,7 +431,7 @@ it is called from.
 The Switch statement
 --------------------
 
-A ``switch`` statement is a form of flow control that provides for a predicated selection of statement(s) to be executed based on a discriminating controlling value. The discriminating controlling value can be either *explicit* - as it is the case for ``case`` statements - or *none of the above* - which is the case for ``default`` statements.
+A ``switch`` statement is a form of flow control that provides for a predicated selection of zero, one or more statements to be executed based on a discriminating controlling value. The discriminating controlling value can be either *explicit* - as it is the case for ``case`` statements - or *none of the above* - which is the case for ``default`` statements.
 
 A ``switch`` statement is not a loop. It does not iterate over a sequence of values.
 
@@ -452,19 +452,19 @@ An OpenQASM3 ``switch`` statement shall be the following grammar:
 
 - The ``switch`` keyword.
    
-- A right paren '(' literal.
+- A right paren ``(`` literal.
    
 - A ``controlling expression``.
    
-- A left paren ')' literal.
+- A left paren ``)`` literal.
    
-- A left brace '{' literal.
+- A left brace ``{`` literal.
    
 - A sequence of one or more ``case`` statements (defined below).
    
 - Either zero or one ``default`` statement(s) (defined below).
    
-- A right brace '}' literal.
+- A right brace ``}`` literal.
    
 
 The ``controlling expression`` of a ``switch`` statement shall be of integer type. Implicit conversions to an integer type are not allowed.
@@ -473,34 +473,34 @@ A ``case`` statement shall be the following grammar:
 
 - The ``case`` keyword.
 - An ``integer-constant-expression`` controlling label.
-- One semicolon literal (':').
-- An optional left-brace literal: '{'.
+- One semicolon literal (``:``).
+- An optional left-brace literal: ``{``.
 - A sequence of zero, one or more OpenQASM3 statements.
-- An optional right-brace literal: '}'.
-- An optional ``break`` statement followed by a semicolon literal.
+- An optional right-brace literal: ``}``.
+- An optional ``break`` statement followed by a semicolon (``;``) literal.
 
 A ``default`` statement shall be the following grammar:
 
 - The ``default`` keyword.
-- One semicolon literal (':').
-- An optional left-brace literal: '{'.
+- One semicolon literal (``:``).
+- An optional left-brace literal: ``{``.
 - A sequence of zero, one or more OpenQASM3 statements.
-- An optional right-brace literal: '}'.
-- An optional ``break`` statement followed by a semicolon literal.
+- An optional right-brace literal: ``}``.
+- An optional ``break`` statement followed by a semicolon (``;``) literal.
 
 A ``switch`` statement shall be in scope only within the scope where it is defined.
 
 The left and right braces of a ``switch`` statement shall not create brace-enclosed scope.
 
-Declarations or statements within a ``switch`` statement scope but outside of a ``case`` or ``default`` statement are ill-formed. The compiler shall raise an error diagnostic for such cases.
+Declarations or statements at ``switch`` statement scope but outside of a ``case`` or ``default`` statement are ill-formed. The compiler shall raise an error diagnostic for such cases.
 
-A ``case`` or ``default`` statement may create brace-enclosed scope, if such statements use the optional left and right brace literals.
+A ``case`` or ``default`` statement may create brace-enclosed scope, if and only if such statements use the optional left and right brace literals.
 
 ``case`` statements may create scope by using both left and right braces, or shall not use braces at all, in which case no scope is created.
 
 ``default`` statements may create scope by using both left and right braces, or shall not use braces at all, in which case no scope is created.
 
-A ``break`` statement shall be followed by a semicolon literal (;). ``break`` statements not followed by a semicolon literal shall raise an error diagnostic.
+A ``break`` statement shall be followed by a semicolon literal (``;``). ``break`` statements not followed by a semicolon literal shall raise an error diagnostic.
 
 A ``break`` statement followed by a semicolon terminates the ``case`` or ``default`` statement it belongs to.
 
@@ -522,7 +522,7 @@ The compiler shall issue a warning diagnostic for all fall-through ``case`` or `
 
 A ``switch`` statement shall contain at least one ``case`` statement. A ``switch`` statement with no ``case`` statements shall raise an error diagnostic.
 
-A ``switch`` statement is not required to contain a ``default`` statement. In this case, the compiler shall issue a warning diagnostic.
+A ``switch`` statement is not required to contain a ``default`` statement. If a ``default`` statement is absent, the compiler shall issue a warning diagnostic.
 
 Examples:
 
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -468,7 +468,7 @@ A ``case`` statement shall be the following grammar:
 - A sequence of zero, one or more OpenQASM3 statements.
 - A right-brace literal: ``}``.
 
-The ``integer-constant-list-expression`` is a sequence of one or more integer constants separated by comma ``,`` literals.
+The ``integer-constant-list-expression`` is a sequence of one or more integer `const` expressions separated by comma ``,`` literals.
 
 A ``default`` statement shall be the following grammar:
 
@@ -493,7 +493,7 @@ A ``case`` or ``default`` statement ending with a right-brace ``}`` terminates t
 
 A ``switch`` statement shall contain at least one ``case`` statement. A ``switch`` statement with no ``case`` statements shall raise an error diagnostic.
 
-A ``switch`` statement is not required to contain a ``default`` statement.
+A ``switch`` statement is not required to contain a ``default`` statement. If a ``switch`` statement does not contain a ``default`` statement and a runtime value is provided to the controlling expression that does not match any case, then the ``switch`` becomes effectively a no-op.
 
 Examples:
 
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -449,19 +449,12 @@ An OpenQASM3 ``switch`` statement shall use the following keywords:
 An OpenQASM3 ``switch`` statement shall be the following grammar:
 
 - The ``switch`` keyword.
-   
 - A right paren ``(`` literal.
-   
 - A ``controlling expression``.
-   
 - A left paren ``)`` literal.
-   
 - A left brace ``{`` literal.
-   
 - A sequence of one or more ``case`` statements (defined below).
-   
 - Either zero or one ``default`` statement(s) (defined below).
-   
 - A right brace ``}`` literal.
    
 
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -445,8 +445,6 @@ An OpenQASM3 ``switch`` statement shall use the following keywords:
 
 - ``default``
 
-- ``break``
-
 
 An OpenQASM3 ``switch`` statement shall be the following grammar:
 
@@ -472,21 +470,19 @@ The ``controlling expression`` of a ``switch`` statement shall be of integer typ
 A ``case`` statement shall be the following grammar:
 
 - The ``case`` keyword.
-- An ``integer-constant-expression`` controlling label.
-- One semicolon literal (``:``).
-- An optional left-brace literal: ``{``.
+- An ``integer-constant-list-expression`` controlling label.
+- A left-brace literal: ``{``.
 - A sequence of zero, one or more OpenQASM3 statements.
-- An optional right-brace literal: ``}``.
-- An optional ``break`` statement followed by a semicolon (``;``) literal.
+- A right-brace literal: ``}``.
+
+The ``integer-constant-list-expression`` is a sequence of one or more integer constants separated by comma ``,`` literals.
 
 A ``default`` statement shall be the following grammar:
 
 - The ``default`` keyword.
-- One semicolon literal (``:``).
-- An optional left-brace literal: ``{``.
+- A left-brace literal: ``{``.
 - A sequence of zero, one or more OpenQASM3 statements.
-- An optional right-brace literal: ``}``.
-- An optional ``break`` statement followed by a semicolon (``;``) literal.
+- A right-brace literal: ``}``.
 
 A ``switch`` statement shall be in scope only within the scope where it is defined.
 
@@ -494,35 +490,17 @@ The left and right braces of a ``switch`` statement shall not create brace-enclo
 
 Declarations or statements at ``switch`` statement scope but outside of a ``case`` or ``default`` statement are ill-formed. The compiler shall raise an error diagnostic for such cases.
 
-A ``case`` or ``default`` statement may create brace-enclosed scope, if and only if such statements use the optional left and right brace literals.
-
-``case`` statements may create scope by using both left and right braces, or shall not use braces at all, in which case no scope is created.
-
-``default`` statements may create scope by using both left and right braces, or shall not use braces at all, in which case no scope is created.
-
-A ``break`` statement shall be followed by a semicolon literal (``;``). ``break`` statements not followed by a semicolon literal shall raise an error diagnostic.
-
-A ``break`` statement followed by a semicolon terminates the ``case`` or ``default`` statement it belongs to.
-
-Declarations at brace-enclosed ``case`` or ``default`` scope shall be in scope only within the scope of their brace enclosing ``case`` or ``default`` statements, regardless of whether the enclosing statement contains a terminating ``break`` statement or not.
+A ``case`` or ``default`` statement creates brace-enclosed scope.
 
 Declarations of types that automatically acquire global scope in OpenQASM3 - such as gates, functions, arrays, qubits and defcals - are not allowed at ``case`` or ``default`` ``switch`` statement scope. Use of such declarations is ill-formed and requires a compiler diagnostic.
 
-Duplicate values for ``case`` statements controlling labels are not allowed. The compiler shall issue an error diagnostic in such cases.
+Duplicate values within any ``integer-constant-list-expression`` for controlling labels of ``case`` statements are not allowed. The compiler shall issue an error diagnostic in such cases.
 
-A ``case`` or ``default`` statement ending with a ``break`` statement followed by a semicolon terminates the execution of the ``switch`` statement. After executing all the statements of the ``case`` or ``default`` statement, control is then transferred to the first statement following the closing right brace of the enclosing ``switch`` statement.
-
-A ``case`` statement that does not contain a terminating ``break`` statement followed by a semicolon transfers control to the statement immediately following it.
-
-A ``default`` statement that does not contain a terminating ``break`` statement followed by a semicolon transfers control outside the enclosing ``switch`` statement, to the first statement immediately following.
-
-This is sometimes referred to as "fall-through".
-
-The compiler shall issue a warning diagnostic for all fall-through ``case`` or ``default`` ``switch`` statements.
+A ``case`` or ``default`` statement ending with a right-brace ``}`` terminates the execution of the ``switch`` statement. After executing all the statements of the ``case`` or ``default`` statement, control is then transferred to the first statement following the closing right brace of the enclosing ``switch`` statement.
 
 A ``switch`` statement shall contain at least one ``case`` statement. A ``switch`` statement with no ``case`` statements shall raise an error diagnostic.
 
-A ``switch`` statement is not required to contain a ``default`` statement. If a ``default`` statement is absent, the compiler shall issue a warning diagnostic.
+A ``switch`` statement is not required to contain a ``default`` statement.
 
 Examples:
 
@@ -535,34 +513,43 @@ Examples:
 	int i = 15;
 
 	switch (i) {
-	case 1: {
+	case 1, 3, 5 {
 	  // OpenQASM3 statement(s)
 	}
-	  break;
-	case 2: {
+	case 2, 4, 6 {
 	  // OpenQASM3 statement(s)
 	}
-	  break;
-	case 3: {
+	case -1 {
 	  // OpenQASM3 statement(s)
 	}
-	  break;
-	case 5: {
+	default {
 	  // OpenQASM3 statement(s)
 	}
-	  break;
-	case 12: {
-	  // OpenQASM3 statement(s)
-	}
-	  break;
-	default: {
-	  // OpenQASM3 statement(s)
-	}
-	  break;
 	}
 	
-	
-2. A ``switch`` statement containing declarations at ``case`` statement scope, and a function call, also at ``case`` statement scope:
+
+2. A switch statement with binary literals in the ``case`` statements:
+
+.. code-block::
+
+	OPENQASM 3.0;
+
+	bit[2] b;
+	switch (int(b)) {
+	case 0b00 {
+	  // OpenQASM3 statement(s)
+	}
+	case 0b01 {
+	  // OpenQASM3 statement(s)
+	}
+	case 0b10 {
+	  // OpenQASM3 statement(s)
+	}
+	case 0b11 {
+	  // OpenQASM3 statement(s)
+	}
+	}```
+3. A ``switch`` statement containing declarations at ``case`` statement scope, and a function call, also at ``case`` statement scope:
 
 
 .. code-block::
@@ -580,64 +567,19 @@ Examples:
 
 	bit c1;
 
-	qubit[8] $0;
-	qubit $1;
+	qubit[8] q0;
 
 	switch (i) {
-	case 1: {
-	  j = k + foo(k, $0);
+	case 1 {
+	  j = k + foo(k, q0);
 	}
-	  break;
-	case 2: {
+	case 2 {
 	  float[64] d = j / k;
 	}
-	  break;
-	case 3: {
+	case 3 {
 	}
-	  break;
-	default: {
+	default {
 	}
-	  break;
-	}
-	
-	
-3. A ``switch`` statement containing fall-through ``case`` statements.
-
-
-.. code-block::
-
-	OPENQASM 3.0;
-
-	def foo(int i, qubit[8] d) -> bit {
-	  return measure d[i];
-	}
-
-	int i = 15;
-
-	int j = 1;
-	int k = 2;
-
-	float[64] d = 0.0;
-
-	bit c1;
-
-	qubit[8] $0;
-	qubit $1;
-
-	switch (i) {
-	case 1: {
-	  j = k + foo(k, $0);
-	}
-	  break;
-	case 2:
-	  d = j / k;
-	  break;
-	case 3:
-	case 5:
-	  c1 = foo(k, $0);
-	  break;
-	default:
-	  break;
 	}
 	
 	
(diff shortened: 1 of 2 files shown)


blakejohnson:
I need to do a detailed read through of this proposal. I support the general concept, though, as having switch statements should make it easier for compilers to infer jump tables in many circumstances. This is useful for reducing branch latency is near-term architectures.

source/language/classical.rst
```diff
@@ -427,6 +427,278 @@ Terminating the program early
 The statement ``end;`` immediately terminates the program, no matter what scope
 it is called from.
 
+
+The Switch statement
+--------------------
+
+An OpenQASM3 ``switch`` statement shall use the following keywords:
+    - ``switch``
```
taalexander, line 435: From this point on the list is rendering incorrectly. It may be due to an indentation error, [see](https://github.com/openqasm/openqasm/blob/2ed64cbf4a5cb8ebe29fae11d8a4f0a82f8ca790/source/language/classical.rst).

source/language/classical.rst
```diff
@@ -427,6 +427,278 @@ Terminating the program early
 The statement ``end;`` immediately terminates the program, no matter what scope
 it is called from.
 
+
+The Switch statement
+--------------------
+
```
taalexander, line 433: Can we please start the definition of the switch statement with a brief description of its functionality/purpose, ie.,  for `while` the equivalent is

> The statement while ( bool ) <body> executes program until the Boolean evaluates to false [[3]](https://github.com/openqasm/openqasm/blob/2ed64cbf4a5cb8ebe29fae11d8a4f0a82f8ca790/source/language/classical.rst#id11). Variables in the loop condition statement may be modified within the while loop body. The body can be either a single statement terminated by a semicolon, or a program block in curly braces {} of several statements:

source/language/classical.rst
```diff
@@ -427,6 +427,278 @@ Terminating the program early
...
+An OpenQASM3 ``switch`` statement shall use the following keywords:
+    - ``switch``
+    - ``case``
+    - ``default``
+    - ``break``
+
+An OpenQASM3 ``switch`` statement shall be the following grammar:
+
+      - the ``switch`` keyword
+      - a right paren '(' literal
```
taalexander, line 443: Can we please use `(`, ie, `` markup for this and other similar instances below as this is consistent with the rest of the document?

source/language/classical.rst
```diff
@@ -427,6 +427,278 @@ Terminating the program early
...
+The Switch statement
+--------------------
+
+An OpenQASM3 ``switch`` statement shall use the following keywords:
+    - ``switch``
+    - ``case``
+    - ``default``
+    - ``break``
+
+An OpenQASM3 ``switch`` statement shall be the following grammar:
```
taalexander, line 440: This specification is much more comprehensive then the other classical control statements (in a good way).

source/language/classical.rst
```diff
@@ -427,6 +427,278 @@ Terminating the program early
...
+	  A ``break`` statement followed by a semicolon terminates the ``case``
+      or ``default`` statement it belongs to.
+
+	  Declarations at brace-enclosed ``case`` or ``default`` scope shall
+      be in scope only within the scope of their brace enclosing ``case``
+      or ``default`` statements, regardless of whether the enclosing
+      statement contains a terminating ``break`` statement or not.
+
+	  Declarations of types that automatically acquire global scope in
+      OpenQASM3 - such as gates, qubits and defcals - are not allowed
```
taalexander, line 502: 👍

source/language/classical.rst
```diff
@@ -427,6 +427,278 @@ Terminating the program early
...
+	  A ``switch`` statement shall contain at least one ``case`` statement.
+      A ``switch`` statement with no ``case`` statements shall raise an
+      error diagnostic.
+
+	  A ``switch`` statement is not required to contain a ``default`` statement.
+      In this case, the compiler shall issue a warning diagnostic.
+
+  Examples:
+
+.. code-block::
```
taalexander, line 539: In front of each example can we please add a short sentence describing what it is demonstrating?

taalexander (changes requested):
Hi Stefan, this look great 😄, a few minor changes are requested.

source/language/classical.rst
steleman: Yes it looks like my merge got all messed-up so I will have to re-format and re-submit this anyway.

source/language/classical.rst
steleman: Done.

source/language/classical.rst
steleman: Done.

source/language/classical.rst
steleman: Done.

source/language/types.rst
```diff
@@ -743,6 +743,8 @@ the shape and type of the assigned value must match that of the reference.
 Arrays may be passed to subroutines and externs. For more details, see
 :any:`arrays-in-subroutines`.
 
+OpenQASM arrays shall be represented in Row Major order.
+
```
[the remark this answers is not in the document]
steleman: Removed.

levbishop:
I'm in favour of the general idea, but have some comments on the details.

### Fall-through
Allowing fall-through but additionally mandating a warning in such case seems inconsistent. I'll give examples of alternative rules for a task to match either {1,3,5} or {2,4,6} or {-1}.

Option 1:
Allow only fall-through in the situation that there is no other code in a case/default.
```openqasm
switch (i) {
case 1: // fall-through with no warning/error because the case is empty
case 3: 
case 5:  
// do something for positive odd case
break;
case 2:
case 4:
case 6:
//do something for positive even case
// no break here
case -1: // accidental fall-through raises error
//do something for special -1 flag
break
}
```

Option 2:
Make the different cases mutually exclusive with no fall-through, like in eg rust/python. Probably good idea to extend the case statement to take multiple matches, and require braces for each case to emphasize that they are exclusive:
```openqasm
switch (i) {
case 1, 3, 5 { // multiple matches
   // do something for positive odd case
}
case 2, 4, 5 {
    //do something for positive even case
}
case -1 {
    //do something for special -1 flag
}
}
```

Option 3: 
Have some syntax for explicit fallthrough eg  a `nobreak` keyword.
```openqasm
switch (i) {
case 1: 
nobreak;
case 3: 
nobreak;
case 5: 
// do something for positive odd case
break;
case 2: 
nobreak;
case 4:
nobreak;
case 6:
//do something for positive even case
// no break here
case -1: // accidental fall-through with no break/nobreak raises error
//do something for special -1 flag
break
}
```

Option 4:
Just don't *mandate* the warning in the fallthrough case. (Lint tools may still *choose* to raise additional warnings beyond minimum required by language spec).
```openqasm
switch (i) {
case 1: // fall-through with no warning/error because the case is empty
case 3: 
case 5:  
// do something for positive odd case
break;
case 2:
case 4:
case 6:
//do something for positive even case
// no break here
case -1: // accidental fall-through not required to warn/error - some tooling *may* warn
//do something for special -1 flag
break;
}
```

Option 5:
Some combination of the other options - multiple-match `case 1, 3, 5` plus explicit `nobreak`; or `nobreak` but only required in the case of non-empty `case` body.

I don't have strong opinion but lean toward option 4. It's familiar and C-like and as far as I know nowhere else in the openqasm spec do we *mandate* warnings. In this case I would also make the missing-default warning equally optional. Ie, `s/shall issue a warning diagnostic/may issue a warning diagnostic/g`

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+- A left brace ``{`` literal.
+   
+- A sequence of one or more ``case`` statements (defined below).
+   
+- Either zero or one ``default`` statement(s) (defined below).
+   
+- A right brace ``}`` literal.
+   
+
+The ``controlling expression`` of a ``switch`` statement shall be of integer type. Implicit conversions to an integer type are not allowed.
```
blakejohnson, line 470: Would it make sense to also allow bit array types in controlling expressions? i.e. so that one could have something like:
```
bit[3] r;
qubit[3] q;
for n in [0:2] {
   r[n] = measure q[n];
}
switch (r) {
   case "000":
   case "001":
      // ...
}

File: source/language/classical.rst [hunk 9]
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+- The ``default`` keyword.
+- One semicolon literal (``:``).
+- An optional left-brace literal: ``{``.
+- A sequence of zero, one or more OpenQASM3 statements.
+- An optional right-brace literal: ``}``.
+- An optional ``break`` statement followed by a semicolon (``;``) literal.
+
+A ``switch`` statement shall be in scope only within the scope where it is defined.
+
+The left and right braces of a ``switch`` statement shall not create brace-enclosed scope.
```
blakejohnson, line 493: Why this behavior?

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+A ``default`` statement that does not contain a terminating ``break`` statement followed by a semicolon transfers control outside the enclosing ``switch`` statement, to the first statement immediately following.
+
+This is sometimes referred to as "fall-through".
+
+The compiler shall issue a warning diagnostic for all fall-through ``case`` or ``default`` ``switch`` statements.
+
+A ``switch`` statement shall contain at least one ``case`` statement. A ``switch`` statement with no ``case`` statements shall raise an error diagnostic.
+
+A ``switch`` statement is not required to contain a ``default`` statement. If a ``default`` statement is absent, the compiler shall issue a warning diagnostic.
```
blakejohnson, line 525: This makes sense if the controlling type is required to be an integer. If we decide to allow for bit arrays, then exhaustive cases would be more common and this could be dropped.

blakejohnson (commented):
Seems like a great addition. As pointed out by others, the style here is quite a bit more prescriptive than other parts of the spec, but it also leaves little doubt about interpretation. So, probably a good thing.

My only question is whether to allow bit array types in the controlling statements rather than just integers?

source/language/classical.rst
steleman: It's no longer a `switch` statement. It's a collection of `for` loops inside a sequence of `if / else if / else` conditionals.

A `switch` statement is explicitly _not_ a loop.

The whole point of a `switch` statement is the construction of a LUT. That is, the `case` statement label value **_must_** be a known compile-time constant. If the controlling `case` label statement value is not a known compile-time constant, the compiler cannot construct a LUT. That defeats the entire purpose of the `switch` statement.

Also: `case` label values must be integer types, not strings.

Even if we considered the possibility of allowing `switch` on a `String` type, OpenQASM3 does not define what a `String` type is. That being the case, how can we `switch` on it?

The complexity of switching on strings is `O(N)` times number of `case` statements. The compiler will have to write a `strcmp(3C)` for every single `case` statement, and the complexity of `strcmp(3C) is `O(N)`. That's antithetic to the purpose of a `switch` statement.

- What happens in the case where the `string` case label value that's being switched on does not exist?
- How does the compiler check that all the possible `case` label statement values have been exhausted in the `switch` statement?

blakejohnson: We discussed at TSC meeting today that we can address my usability concern with the following pattern:
```
bit[3] b;
switch (int(b)) {
   case 0b000:
      // ...
   case 0b001:
      // ...
}
```
i.e. write the literals for the cases using OpenQASM's syntax for binary integer literals.

Therefore, we decided to table the idea of extending the allowed types in the controlling statement to include bit arrays. It is something we can revisit in the future.

blakejohnson:
At last TSC meeting we discussed that we wanted to narrow down the choices to what Lev labeled as Options 2 and 4. Option 2 basically adds syntax for case statements to match multiple values, while Option 4 just drops the required warning from @steleman's proposal. Plan is to vote on this at TSC meeting today.

levbishop:
As voted at 2023-06-21 TSC Meeting option 2 is chosen.

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
 The statement ``end;`` immediately terminates the program, no matter what scope
 it is called from.
 
+
+The Switch statement
+--------------------
+
+A ``switch`` statement is a form of flow control that provides for a predicated selection of zero, one or more statements to be executed based on a discriminating controlling value. The discriminating controlling value can be either *explicit* - as it is the case for ``case`` statements - or *none of the above* - which is the case for ``default`` statements.
```
blakejohnson, line 434:
```suggestion
A ``switch`` statement is a form of flow control that provides for a predicated selection of zero, one or more statements to be executed based on discriminating controlling values. The discriminating controlling values must be *explicit*, i.e. OpenQASM does not have the equivalent of a ``default`` statement found in some other languages.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+An OpenQASM3 ``switch`` statement shall use the following keywords:
+
+- ``switch``
+
+- ``case``
+
+- ``default``
+
+- ``break``
+
```
blakejohnson, line 449:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+- A ``controlling expression``.
+   
+- A left paren ``)`` literal.
+   
+- A left brace ``{`` literal.
+   
+- A sequence of one or more ``case`` statements (defined below).
+   
+- Either zero or one ``default`` statement(s) (defined below).
+
```
blakejohnson, line 466:
```suggestion
```

source/language/classical.rst
blakejohnson, line 449:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+- An optional left-brace literal: ``{``.
+- A sequence of zero, one or more OpenQASM3 statements.
+- An optional right-brace literal: ``}``.
+- An optional ``break`` statement followed by a semicolon (``;``) literal.
+
+A ``default`` statement shall be the following grammar:
+
+- The ``default`` keyword.
+- One semicolon literal (``:``).
+- An optional left-brace literal: ``{``.
```
blakejohnson, line 486:
```suggestion
- An ``integer-constant-list-expression`` controlling label.
- A left-brace literal: ``{``.
- A sequence of zero, one or more OpenQASM3 statements.
- A right-brace literal: ``}``.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+- A sequence of zero, one or more OpenQASM3 statements.
+- An optional right-brace literal: ``}``.
+- An optional ``break`` statement followed by a semicolon (``;``) literal.
+
+A ``switch`` statement shall be in scope only within the scope where it is defined.
```
blakejohnson, line 491:
```suggestion
- An ``integer-constant-list-expression`` controlling label.
- A left-brace literal: ``{``.
- A sequence of zero, one or more OpenQASM3 statements.
- A right-brace literal: ``}``.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+The left and right braces of a ``switch`` statement shall not create brace-enclosed scope.
+
+Declarations or statements at ``switch`` statement scope but outside of a ``case`` or ``default`` statement are ill-formed. The compiler shall raise an error diagnostic for such cases.
```
blakejohnson, line 495:
```suggestion
Declarations or statements at ``switch`` statement scope but outside of a ``case`` statement are ill-formed. The compiler shall raise an error diagnostic for such cases.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+``case`` statements may create scope by using both left and right braces, or shall not use braces at all, in which case no scope is created.
+
+``default`` statements may create scope by using both left and right braces, or shall not use braces at all, in which case no scope is created.
+
+A ``break`` statement shall be followed by a semicolon literal (``;``). ``break`` statements not followed by a semicolon literal shall raise an error diagnostic.
+
+A ``break`` statement followed by a semicolon terminates the ``case`` or ``default`` statement it belongs to.
+
+Declarations at brace-enclosed ``case`` or ``default`` scope shall be in scope only within the scope of their brace enclosing ``case`` or ``default`` statements, regardless of whether the enclosing statement contains a terminating ``break`` statement or not.
+
```
blakejohnson, line 508:
```suggestion
```

source/language/classical.rst
blakejohnson, line 449:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+   
+- A right brace ``}`` literal.
+   
+
+The ``controlling expression`` of a ``switch`` statement shall be of integer type. Implicit conversions to an integer type are not allowed.
+
+A ``case`` statement shall be the following grammar:
+
+- The ``case`` keyword.
+- An ``integer-constant-expression`` controlling label.
```
blakejohnson, line 475:
```suggestion
- An ``integer-constant-list-expression`` controlling label.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+- One semicolon literal (``:``).
+- An optional left-brace literal: ``{``.
```
blakejohnson, line 477:
```suggestion
- A left-brace literal: ``{``.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+- A sequence of zero, one or more OpenQASM3 statements.
+- An optional right-brace literal: ``}``.
+- An optional ``break`` statement followed by a semicolon (``;``) literal.
```
blakejohnson, line 480:
```suggestion
- A right-brace literal: ``}``.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+A ``default`` statement shall be the following grammar:
```
blakejohnson, line 482:
```suggestion
The ``integer-constant-list-expression`` is a sequence of one or more integer constants separated by comma ``,`` literals.

A ``default`` statement shall be the following grammar:
```

source/language/classical.rst
blakejohnson, line 486:
```suggestion
- A left-brace literal: ``{``.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+- The ``default`` keyword.
+- One semicolon literal (``:``).
+- An optional left-brace literal: ``{``.
+- A sequence of zero, one or more OpenQASM3 statements.
+- An optional right-brace literal: ``}``.
+- An optional ``break`` statement followed by a semicolon (``;``) literal.
```
blakejohnson, line 489:
```suggestion
- A right-brace literal: ``}``.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+A ``break`` statement shall be followed by a semicolon literal (``;``). ``break`` statements not followed by a semicolon literal shall raise an error diagnostic.
+
+A ``break`` statement followed by a semicolon terminates the ``case`` or ``default`` statement it belongs to.
+
+Declarations at brace-enclosed ``case`` or ``default`` scope shall be in scope only within the scope of their brace enclosing ``case`` or ``default`` statements, regardless of whether the enclosing statement contains a terminating ``break`` statement or not.
+
+Declarations of types that automatically acquire global scope in OpenQASM3 - such as gates, functions, arrays, qubits and defcals - are not allowed at ``case`` or ``default`` ``switch`` statement scope. Use of such declarations is ill-formed and requires a compiler diagnostic.
+
+Duplicate values for ``case`` statements controlling labels are not allowed. The compiler shall issue an error diagnostic in such cases.
```
blakejohnson, line 511:
```suggestion
Duplicate values within any ``integer-constant-list-expression`` for controlling labels of ``case`` statements are not allowed. The compiler shall issue an error diagnostic in such cases.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+A ``case`` or ``default`` statement ending with a ``break`` statement followed by a semicolon terminates the execution of the ``switch`` statement. After executing all the statements of the ``case`` or ``default`` statement, control is then transferred to the first statement following the closing right brace of the enclosing ``switch`` statement.
```
blakejohnson, line 513:
```suggestion
A ``case`` or ``default`` statement ending with right-brace ``}`` terminates the execution of the ``switch`` statement. After executing all the statements of the ``case`` or ``default`` statement, control is then transferred to the first statement following the closing right brace of the enclosing ``switch`` statement.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+A ``case`` statement that does not contain a terminating ``break`` statement followed by a semicolon transfers control to the statement immediately following it.
+
+A ``default`` statement that does not contain a terminating ``break`` statement followed by a semicolon transfers control outside the enclosing ``switch`` statement, to the first statement immediately following.
+
+This is sometimes referred to as "fall-through".
+
+The compiler shall issue a warning diagnostic for all fall-through ``case`` or ``default`` ``switch`` statements.
+
```
blakejohnson, line 522:
```suggestion
```

source/language/classical.rst
blakejohnson, line 525:
```suggestion
A ``switch`` statement is not required to contain a ``default`` statement.
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+1. A simple ``switch`` statement with ``case`` and ``default`` statements:
+
+.. code-block::
+
+	OPENQASM 3.0;
+
+	int i = 15;
+
+	switch (i) {
+	case 1: {
```
blakejohnson, line 538:
```suggestion
	case 1, 3, 5 {
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	  // OpenQASM3 statement(s)
+	}
+	  break;
+	case 2: {
```
blakejohnson, line 542:
```suggestion
	case 2, 4, 6 {
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	  break;
+	case 5: {
+	  // OpenQASM3 statement(s)
+	}
+	  break;
+	case 12: {
+	  // OpenQASM3 statement(s)
+	}
+	  break;
+	default: {
```
blakejohnson, line 558:
```suggestion
	default {
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	  // OpenQASM3 statement(s)
+	}
+	  break;
```
blakejohnson, line 561:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	}
+	
+
```
blakejohnson, line 564:
```suggestion

2. A switch statement with binary literals in the ``case`` statements:

.. code-block::

  OPENQASM 3.0;

  bit[2] b;
  switch (int(b)) {
  case 0b00 {
    // OpenQASM3 statement(s)
  }
  case 0b01 {
    // OpenQASM3 statement(s)
  }
  case 0b10 {
    // OpenQASM3 statement(s)
  }
  case 0b11 {
    // OpenQASM3 statement(s)
  }
  }
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	}
+	  break;
+	default: {
+	  // OpenQASM3 statement(s)
+	}
+	  break;
+	}
+	
+	
+2. A ``switch`` statement containing declarations at ``case`` statement scope, and a function call, also at ``case`` statement scope:
```
blakejohnson, line 565:
```suggestion
3. A ``switch`` statement containing declarations at ``case`` statement scope, and a function call, also at ``case`` statement scope:
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	int j = 1;
+	int k = 2;
+
+	bit c1;
+
+	qubit[8] $0;
+	qubit $1;
+
+	switch (i) {
+	case 1: {
```
blakejohnson, line 587:
```suggestion
	case 1 {
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	  j = k + foo(k, $0);
+	}
+	  break;
+	case 2: {
```
blakejohnson, line 591:
```suggestion
	case 2 {
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	  float[64] d = j / k;
+	}
+	  break;
+	case 3: {
```
blakejohnson, line 595:
```suggestion
	case 3 {
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	}
+	  break;
+	default: {
```
blakejohnson, line 598:
```suggestion
	default {
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	}
+
+	int i = 15;
+
+	int j = 1;
+	int k = 2;
+
+	bit c1;
+
+	qubit[8] $0;
```
blakejohnson, line 583:
```suggestion
	qubit[8] q0;
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	qubit $1;
```
blakejohnson, line 584:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+	switch (i) {
+	case 1: {
+	  j = k + foo(k, $0);
```
blakejohnson, line 588:
```suggestion
	  j = k + foo(k, q0);
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+	  for int i in [0 : 7] {
+		k = measure q[i];
+		r += k;
+	  }
+
+	  return r;
+	}
+
+	qubit[8] $0;
```
blakejohnson, line 662:
```suggestion
	qubit[8] q;
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+	int j = 30;
+	int i = foo($0);
```
blakejohnson, line 665:
```suggestion
	int i = foo(q);
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+
+	switch (i) {
+	case 1: {
```
blakejohnson, line 668:
```suggestion
	case 1, 2, 5, 12 {
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	}
+	  break;
+	case 2: {
+	}
+	  break;
+	case 3: {
+	  switch (j) {
+	  case 10: {
```
blakejohnson, line 676:
```suggestion
	  case 10, 15, 20 {
	  	  h q;	  
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	  case 15: {
+	  }
+		break;
+	  case 20: {
+	  }
+		break;
+	  default: {
+	  }
+		break;
+	  }
```
blakejohnson, line 688:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,279 @@ Terminating the program early
...
+	  break;
+	case 5: {
+	}
+	  break;
+	case 12: {
+	}
+	  break;
+	default: {
+	}
+	  break;
```
blakejohnson, line 699:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,197 @@ Terminating the program early
...
+
+- ``case``
+
+- ``default``
+
+
+An OpenQASM3 ``switch`` statement shall be the following grammar:
+
+- The ``switch`` keyword.
+
```
blakejohnson, line 452:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,197 @@ Terminating the program early
...
+- ``default``
+
+
+An OpenQASM3 ``switch`` statement shall be the following grammar:
+
+- The ``switch`` keyword.
+   
+- A right paren ``(`` literal.
+
```
blakejohnson, line 454:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,197 @@ Terminating the program early
...
+
+An OpenQASM3 ``switch`` statement shall be the following grammar:
+
+- The ``switch`` keyword.
+   
+- A right paren ``(`` literal.
+   
+- A ``controlling expression``.
+
```
blakejohnson, line 456:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,197 @@ Terminating the program early
...
+An OpenQASM3 ``switch`` statement shall be the following grammar:
+
+- The ``switch`` keyword.
+   
+- A right paren ``(`` literal.
+   
+- A ``controlling expression``.
+   
+- A left paren ``)`` literal.
+
```
blakejohnson, line 458:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,197 @@ Terminating the program early
...
+- The ``switch`` keyword.
+   
+- A right paren ``(`` literal.
+   
+- A ``controlling expression``.
+   
+- A left paren ``)`` literal.
+   
+- A left brace ``{`` literal.
+
```
blakejohnson, line 460:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,197 @@ Terminating the program early
...
+- A right paren ``(`` literal.
+   
+- A ``controlling expression``.
+   
+- A left paren ``)`` literal.
+   
+- A left brace ``{`` literal.
+   
+- A sequence of one or more ``case`` statements (defined below).
+
```
blakejohnson, line 462:
```suggestion
```

source/language/classical.rst
```diff
@@ -427,6 +427,197 @@ Terminating the program early
...
+- A ``controlling expression``.
+   
+- A left paren ``)`` literal.
+   
+- A left brace ``{`` literal.
+   
+- A sequence of one or more ``case`` statements (defined below).
+   
+- Either zero or one ``default`` statement(s) (defined below).
+
```
blakejohnson, line 464:
```suggestion
```

blakejohnson:
Spec has been updated to reflect my understanding of our agreement from the TSC meeting of June 21. @jakelishman has volunteered to updated the reference parser accordingly.

source/language/classical.rst
```diff
@@ -427,6 +427,190 @@ Terminating the program early
 The statement ``end;`` immediately terminates the program, no matter what scope
 it is called from.
 
+
+The Switch statement
+--------------------
```
hodgestar, line 432: Nitpick: If looks like the heading level for the `switch` statement is incorrect. I assume the intention is to have it within the `Looping and branching` section, but instead it is a sibling.

On the other hand, the heading level for `Looping and branching` also seems odd -- it's currently inside `Low-level classical instructions` which makes it not findable in the table of contents on the left.

Perhaps `Looping and branching` should be moved up a level? I'm happy to make a separate PR for that if people like.

PR #464, opened 2023-04-06
Fixing forStatement grammar, parser and AST

### LINKED ISSUE #462: [forStatement] ANTLR reference grammar is incorrect
## The issue

In [qasm3Parser.g4#L66](https://github.com/openqasm/openqasm/blob/05c0f9fd5211de972d55c8e4396f88cc251fa777/source/grammar/qasm3Parser.g4#L66)
```antlr
forStatement: FOR scalarType Identifier IN (setExpression | LBRACKET rangeExpression RBRACKET | Identifier) body=statementOrScope;
```
the grammar state that the iterable (i.e., the *thing* that will be iterated during the for-loop) is either:

- a `setExpression`
- a `rangeExpression`
- an `Identifier`

whereas the [official documentation on for-loop](https://openqasm.com/language/classical.html#for-loops) explicitly states that using `indexedIdentifier`s should be allowed:
> It is valid to use an indexing expression (e.g. my_array[1:3]) to arrive at one of the types given above.

## How to fix this

Replace `Identifier` by `indexedIdentifier` in the `forStatement` rule.

nelimee:
### Summary

This PR fixes #462.

For the moment, only the basic modifications have been performed. Type checking might have to be done in the parser or later in the AST when it is typed.

### DIFF
--- a/source/grammar/qasm3Parser.g4
+++ b/source/grammar/qasm3Parser.g4
@@ -63,7 +63,7 @@ includeStatement: INCLUDE StringLiteral SEMICOLON;
 breakStatement: BREAK SEMICOLON;
 continueStatement: CONTINUE SEMICOLON;
 endStatement: END SEMICOLON;
-forStatement: FOR scalarType Identifier IN (setExpression | LBRACKET rangeExpression RBRACKET | Identifier) body=statementOrScope;
+forStatement: FOR scalarType Identifier IN (setExpression | LBRACKET rangeExpression RBRACKET | expression) body=statementOrScope;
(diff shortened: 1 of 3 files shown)


nelimee:
A small note: this is the first time I make modifications to the parser, it would be beneficial to double-check attentively that I did not forget an important modification.

jakelishman (approved):
This looks fine to me as-is, thanks!  There's not really anything for us to test more than already exists, since the AST we're producing is untyped (from the perspective of the OQ3 type system - it's "typed" in Python-space).

PR #470, opened 2023-05-05
Add round-trip test for annotation parse/print

PhilReinhold:
### Summary

Annotations were not being printed at the correct indentation level. An easy bugfix since it just required correctly passing the context object. New test assures round-trip parse/print for annotations.


jakelishman (approved):
Super, thanks!

Issue #472, opened 2023-05-10
Clarify whether defcals can be defined using aliases


amilsted:
### What is the expected enhancement?

It isn't clear to me whether the following is supported:
```
let q0 = $0;

defcal X90 q0 {   // equivalent to `defcal X90 $0 { ... }` ?
    ...
}
```
It seems like a natural and useful thing to support, as naming physical qubits (rather than referring to them only by integer) can clarify code e.g. in cases where there is more than one type of qubit on a device. 

I could not tell from the spec whether this is allowed, so perhaps this should be clarified?

jwoehr:
> [Physical Qubits](https://openqasm.com/language/types.html#physical-qubits)
> While program qubits can be named, hardware qubits are referenced only by the syntax $[NUM]

I take "only" to mean "only" but I can see how the discussion of Aliasing can lead to ambiguity.

> [Aliasing](https://openqasm.com/language/types.html#aliasing)
>The let keyword allows quantum bits and registers to be referred to by another name as long as the alias is in scope.

I have opened Pull Request #473 to address this.

amilsted:
Gonna reference #471, which also makes the case for more flexible references to hardware qubits.

blakejohnson:
I would say that the `q0` referenced in the definition:
```
defcal X90 q0 { }
```
is not the same `q0` as in the surrounding scope. This is also true of how variables in `def` statements can shadow variables in their enclosing scope. e.g.
```
int a = 1;
def foo(int a) -> int {
   // in this scope 'a' references whatever was passed in, not the value of 'a' in the enclosing scope.
```

amilsted:
Ah, I hadn't realized that `defcal X90 q0 { ... }` is already valid and defines a generic defcal.

Perhaps we can reopen https://github.com/openqasm/openqasm/issues/471 and consider more flexible physical-qubit names like `$q0` instead?

PR #473, opened 2023-05-10
Clarify Issue #472 - Physical Qubits cannot be Aliased

jwoehr:
### Summary
Address #472 by clarifying that physical qubits cannot be aliased

### Details and comments
I have no opinion either way, but the standard _seems_ to indicate that physical qubits are not declared, therefore cannot be aliased. Clarifying the language in the reverse direction is equally acceptable.

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -58,7 +58,7 @@ Quantum types
 
 Qubits
 ~~~~~~
-
+physical-qubits
 There is a quantum bit (``qubit``) type that is interpreted as a reference to a
 two-level subsystem of a quantum state. The statement ``qubit name;``
 declares a reference to a quantum bit. These qubits are referred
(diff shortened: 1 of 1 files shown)


jwoehr:
@PhilReinhold I'm happy with either. I just made a change that seemed to clarify the present status. 
The spec says: 

> Physical Qubits
> While program qubits can be named, hardware qubits are referenced only by the syntax `$[NUM]`
> 

So I made this PR based on interpreting that to mean "no aliases".
If the TSC views it your way, I'll alter the PR to clarify in that sense.

bkavcaduhq:
ah, I think alias in spec mean to physical qubit can be named and a qubit have **unlimited state** and scope of defcal ..{} can not simulator to same the physical qubit so state in/out scope are so different

PR #475, opened 2023-06-13
Proposed appointment of Erik Davis to the TSC.

blakejohnson:
@stevenheidel has announced his resignation from the TSC. After discussion, it was decided that to avoid voter fatigue, it would be better to appointment someone to serve the remaining 5-6 months of Steven's term rather than running a snap election cycle now. This PR both amends the governance doc to give the TSC the authority to make such appointments, and proposed Erik Davis as appointee.

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -23,7 +23,7 @@ We will form a **Technical Steering Committee**. The TSC shall be composed of 6 
 * Dor Israeli (Quantum Machines)
 * Lev S Bishop (IBM Quantum)
 * Blake Johnson (IBM Quantum)
-* Steven Heidel (AWS)
+* Erik Davis (AWS)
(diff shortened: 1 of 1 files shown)


stevenheidel (approved):

levbishop (approved):

dorisraeli (approved):
🤩

blakejohnson:
Including me, that's 4 TSC votes for and none against. So, I will proceed to merge.

jakelishman:
I extended a GitHub invitation to @braised-babbage to join the OpenQASM org last night with the right permissions for a TSC member (as part of making sure enough people in AWS had permissions to deploy the Python AST).  I'd thought Erik was already part of the org, but apparently not.

PR #477, opened 2023-06-21
(Draft) Better Unicode Support

### LINKED ISSUE #422: Ambiguity with character code for microsecond

### LINKED ISSUE #440: Clarify unicode support
> The unicode tools went down so i can't quite fill in all the links right now. I'll come back and revise it when they're back...

Inspired by #428 I promised in the last TSC meeting to take a look at best practices for handling unicode equivalences in programming languages. As Jake suspected, like most things Unicode, it turns out to be a pretty involved topic.

### Identifiers: UAX31
Most unicode-related discussion seems to be around which identifiers should be considered valid. There is a whole unicode annex on the topic [UAX-31: UNICODE IDENTIFIER AND PATTERN SYNTAX](https://www.unicode.org/reports/tr31/) which goes through a lot of the possible decisions/related pitfalls and gives various compliance statements that an implementation may assert to give a uniform understanding of these issues across languages/implementations. 

The current OpenQasm spec [identifier section](https://openqasm.com/language/types.html#identifiers) is similar to a UAX31  compliance statement of the form:

- [UAX31-C1](https://www.unicode.org/reports/tr31/tr31-37.html#C1): The OpenQASM language conforms to version 37 of the Unicode® Standard Annex #⁠31 
- [UAX31-C2](https://www.unicode.org/reports/tr31/tr31-37.html#C2): It observes the following requirements:
  - [UAX31-R1-2](https://www.unicode.org/reports/tr31/tr31-37.html#R1-2). **Default Identifiers**: To determine whether a string is an identifier it uses [UAX31-D1](https://www.unicode.org/reports/tr31/tr31-37.html#D1) with the following **profile**:
    - `Start` := `[[:XID_Start]_]`
    - `Continue` := `[:XID_Continue:]`
    - `Medial` := `[]`
  - [UAX31-R1b](https://www.unicode.org/reports/tr31/tr31-37.html#R1b). **Stable Identifiers**: Once a string qualifies as an identifier, it does so in all future versions.

Relative to the current spec, while technically a breaking change, the difference is minor and unlikely to invalidate any code in the wild: 

- The difference for `Start` is just [removal of 22 codepoints](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5BA-Za-z_%5D%5B%3ALu%3A%5D%5B%3ALl%3A%5D%5B%3ALt%3A%5D%5B%3ALm%3A%5D%5B%3ALo%3A%5D%5B%3ANl%3A%5D-%5B%5B_%5D%5B%3AXID_Start%3A%5D%5D%5D&g=&i=) and [addition of 4 codepoints](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5B%5B_%5D%5B%3AXID_Start%3A%5D%5D-%5B%5BA-Za-z_%5D%5B%3ALu%3A%5D%5B%3ALl%3A%5D%5B%3ALt%3A%5D%5B%3ALm%3A%5D%5B%3ALo%3A%5D%5B%3ANl%3A%5D%5D%5D&g=&i=). 
- For `Continue` the difference is [removal of 18 codepoints](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[%5bA%2dZa%2dz%5f0%2d9%5d%5b%3aLu%3a%5d%5b%3aLl%3a%5d%5b%3aLt%3a%5d%5b%3aLm%3a%5d%5b%3aLo%3a%5d%5b%3aNl%3a%5d-%5b%3aXID%5fContinue%3a%5d]) and [addition of 3,130 codepoints](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5B%3AXID_Continue%3A%5D-%5B%5BA-Za-z_0-9%5D%5B%3ALu%3A%5D%5B%3ALl%3A%5D%5B%3ALt%3A%5D%5B%3ALm%3A%5D%5B%3ALo%3A%5D%5B%3ANl%3A%5D%5D%5D&g=&i=)

> **My thought**: I'm happy to assume there are good reasons for these differences that deep thinkers who authored that document have considered, and just modify OpenQASM to use `XID_Start`, `XID_Continue`.

### Security: UTS39 and UTR36
There are a class of so-called [Trojan Source](https://arxiv.org/pdf/2111.00169.pdf) attacks where source code can pass human review (and sometimes also automated analysis) but in the execution environment behave differently than the displayed form suggests. Other unicode documents that are relevant to our discussion are [UTR36: UNICODE SECURITY CONSIDERATIONS](https://www.unicode.org/reports/tr36/) and [UTS39: UNICODE SECURITY MECHANISMS](https://www.unicode.org/reports/tr39/), which has a [section about identifiers](https://www.unicode.org/reports/tr39/#Identifier_Characters). Sticking to characters with `Identifier_Status=Allowed` as defined there removes [27323 codepoints](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5B%3AXID_Continue%3A%5D-%5B%3AIdentifier_Status%3DAllowed%3A%5D%5D&g=Identifier_type&i=) from the set of `Continue` characters, for various reasons.

A problem with defining things this was is that there are not forward or backward-compatibility guarantees around the `Identifier_Status` and characters can have their status change in future revisions, which seems undesirable for a programming language stability. We could freeze a version of UTS39, but then we wouldn't get the benefit of updated protection against newly-discovered security issues. There is an interesting proposal [IPIC9](https://jonathanwarden.com/2016/07/24/unicode-identifiers-in-your-language/#IPIC9_An_Immutable_Profile_for_Identifiers_in_Code) that makes the case for this (see also the two follow-up posts with more reasoning about punctuation and immutability), but this never seem to catch on in the wild and ultimately I don't like it for general purpose programming source code.

> **My thought**: It's useful to read UTS39/UTR36 for other security considerations, but the `Identifier_Status` is too unstable for language spec use. It can factor into linter rules, syntax highlighting in programmer editors, etc. `Identifier_Status` maybe makes sense for things where the identifiers must be used directly, such as hashtags or international domain names, but for programming languages this can delegate to style guides and external tooling rather than the language spec and compiler. (And there are many other unicode security issues such as bidi, line breaking, confusable identifiers etc, that have to be handled in a more global way than simply at the identifier lexing stage, so such tooling will be necessary no matter what). This is the philosophy taken by most other languages: python, rust, etc

### Normalization

All the stuff I read basically agreed that the only sensible options are NFC and NFKC. The argument for NFC is that NFC normalization just gets rid of meaningless and invisible differences, and most input methods don't give any reasonable way to generate sequences that are different but NFC-equivalent. NFC normalization of `XID_Continue` eliminates [1098 code points](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AXID_continue%3A%5D%26%5B%3A%5EisNFC%3A%5D&g=&i=toNFC)  the ones that jump out being 

Ω 	[U+2126](https://util.unicode.org/UnicodeJsps/character.jsp?a=2126)	OHM SIGN	
K 	[U+212A](https://util.unicode.org/UnicodeJsps/character.jsp?a=212A)	KELVIN SIGN	
Å 	[U+212B](https://util.unicode.org/UnicodeJsps/character.jsp?a=212B)	ANGSTROM SIGN

which seems obviously correct to me.

NFKC normalization is less obvious. It eliminates [XXX codepoints]() Some of the normalization seems obviously correct like:
XXX
which seem much like the above choices around OHM SIGN etc. Some of the normalization seems wrong, like 
XXX
where the distinction is the whole point. 

I guess the point is that NFKC normalization was not built around the needs of programming languages (and what I understand of the unicode stability guarantees this is not something that can be "fixed" in the future). One argument against NFKC is that editor/command line search tools don't usually support NFKC-equivalent searching, so "find all the uses of the identifier by name of X" needs specific tool support. UTS55 below gives some examples of the kind of thing can go wrong there.

A rule of thumb (formalized in the draft UTS55) I've seen is that case-insensitive languages should use NFKC and case-sensitive languages should use NFC. 

A final decision is whether to use equivalent normalized comparisons (where the input stream is converted to the normalization form before further processing) or filtered normalization (where having codepoints outside of the normalization form in the input is an error). Choosing the latter avoids the searching-for-identifiers problem from above, but disallows formatting the identifiers the way you would like to, which is more of a problem for some languages than others. It seems pretty obvious that the only 3 combinations that make sense are:

- [UAX31-R4](https://www.unicode.org/reports/tr31/tr31-37.html#R4) **Equivalent Normalized Identifiers** using Normalization Form C (NFC).
- [UAX31-R4](https://www.unicode.org/reports/tr31/tr31-37.html#R4) **Equivalent Normalized Identifiers** using Normalization Form KC (NFKC).
- [UAX31-R6](https://www.unicode.org/reports/tr31/tr31-37.html#R6) **Filtered Normalized Identifiers** using Normalization Form KC (NFKC).

If we chose NFKC I wonder if it might be friendliest to actually go with a non-standard mixed equivalent/filtered normalization where the input is first NFC normalized and any remaining non-NFKC codepoints are treated as an error. Nevertheless...

> **My thought**: I don't love it, but we should go with 
> [UAX31-R4](https://www.unicode.org/reports/tr31/tr31-37.html#R4) **Equivalent Normalized Identifiers** using Normalization Form C (NFC). As a special case, since there are a few greek letters OpenQasm treats specially (pi, mu, etc) we should just clarify in the spec exactly which of their variants we accept as synonyms. Much as the proposed #428 does already. Any remaining ambiguities should be handled in style-guides and linters, etc.

### General programming language considerations: UTS55
Since I started looking into this, a draft proposal was published for a new document [UTS55: UNICODE SOURCE CODE HANDLING](https://www.unicode.org/reports/tr55/), which should be read alongside draft  [proposed updates to UAX31](https://www.unicode.org/reports/tr31/proposed.html)

This looks pretty comprehensive, but it will probably take a bit of effort to translate for our needs. Also it still leaves some decisions for the language designer. The [section 4.1.3 on nested languages](https://www.unicode.org/reports/tr55/#Nested-Languages) is likely relevant to `defcal` blocks.

> **My thought**:  This is fiddly stuff, and we can surely find issues to quibble with the recommendations here, but we unicode is fundamentally messy and unsatisfying and we aren't interested in being unicode experts, so outsourcing the decision-making here is probably the right choice so we should follow UTS55 once it publishes. Another reason to follow UTS55 is that a lot of the recommendations there spread the responsibilities among the compiler/spec and other tooling IDEs/text editors/pretty printers/linters etc so if we deviate from eg the line break rules of UTS55 then unless editors implement OpenQASM-specific linebreaking rules, invisible character highlighting rules..., there would likely be deviations between the editor and compiler's ideas about the extent of comments, etc.

 
### What do other languages do?

- [ISO C++](https://isocpp.org/files/papers/P1949R4.html#conformance-points-for-uax31), and [reference](https://en.cppreference.com/mwiki/index.php?title=cpp/language/identifiers)
- [ISO C Annex D](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3054.pdf#page=523)
- [Rust](https://rust-lang.github.io/rfcs/2457-non-ascii-idents.html)
- [Python reference](https://docs.python.org/3/reference/lexical_analysis.html#identifiers) introduced in [PEP3131](https://peps.python.org/pep-3131/)

All of these in there latest versions are based on UAX31-R1 default identifiers. They all use UAX31-R4: Equivalent Normalized Identifiers, with python choosing NFKC and the rest choosing NFC. (Actually its not 100% clear if C and C++ mean UAX31-R4 Equivalent Normalized Identifiers or UAX31-R6 Filtered Normalized Identifiers, the compliance statement says Equivalent but the commentary suggests Filtered.)

```python
μ𝛍µ𝝁𝜇𝞵𝝻 = 1
μμμμμμμ += 10 # the same variable by a different display name
print(μμμμμμμ) # 11
π𝞏𝛑ᴨ𝝅𝜋𝟉𝝿ϖ𝛡𝝕ℼ𝜛𝞹 = 1 # but not including п
πππᴨππππππππππ += 20
print(πππᴨππππππππππ) # 21
```

levbishop:
### Summary
Fixes #422 
Fixes #440
Replaces #428 

This tries to bring the spec and grammar more into compliance with the latest proposed [UAX31](https://www.unicode.org/reports/tr31/proposed.html) and draft [UTS55](https://www.unicode.org/reports/tr55/)

### Details and comments
ANTLR doesn't have great support for set intersection/difference operations on unicode character sets so I had to duplicate some fragments pulling the definitions from PropList.txt but these shouldn't ever  be changing.

I'm still thinking about how best to handle the rules for ignorable white space.

### DIFF
--- a/source/grammar/qasm3Lexer.g4
+++ b/source/grammar/qasm3Lexer.g4
@@ -130,12 +130,11 @@ OctalIntegerLiteral: '0o' ([0-7] '_'?)* [0-7];
 DecimalIntegerLiteral: ([0-9] '_'?)* [0-9];
 HexIntegerLiteral: ('0x' | '0X') ([0-9a-fA-F] '_'?)* [0-9a-fA-F];
 
-fragment ValidUnicode: [\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}]; // valid unicode chars
-fragment Letter: [A-Za-z];
-fragment FirstIdCharacter: '_' | ValidUnicode | Letter;
-fragment GeneralIdCharacter: FirstIdCharacter | [0-9];
-
+// Identifiers comply with UAX31-R1-2
+fragment FirstIdCharacter: [\p{XID_Start}_];
+fragment GeneralIdCharacter: [\p{XID_Continue}];
 Identifier: FirstIdCharacter GeneralIdCharacter*;
+
 HardwareQubit: '$' [0-9]+;
 
 fragment FloatLiteralExponent: [eE] (PLUS | MINUS)? DecimalIntegerLiteral;
@@ -155,14 +154,23 @@ TimingLiteral: (DecimalIntegerLiteral | FloatLiteral) TimeUnit;
 BitstringLiteral: '"' ([01] '_'?)* [01] '"';
 // allow ``"str"`` and ``'str'``
 StringLiteral
-    : '"' ~["\r\t\n]+? '"'
-    | '\'' ~['\r\t\n]+? '\''
+    : '"' NotUAX31NewlineDoubleQuote+? '"'
+    | '\'' NotUAX31NewlineSingleQuote+? '\''
     ;
 
+// UAX31 Whitespace handling
+fragment UAX31Newline: [\u000A\u000B\u000C\u000D\u0085\u2028\u2029];
+fragment NotUAX31Newline: ~[\u000A\u000B\u000C\u000D\u0085\u2028\u2029];
+fragment NotUAX31NewlineSingleQuote: ~['\u000A\u000B\u000C\u000D\u0085\u2028\u2029];
+fragment NotUAX31NewlineDoubleQuote: ~["\u000A\u000B\u000C\u000D\u0085\u2028\u2029];
+fragment UAX31HorizontalSpace: [ \t\u200E\u200F]; // Pattern_White_Space minus newlines
+fragment NotUAX31WhiteSpace: ~[\p{Pattern_White_Space}];
+fragment UAX31IgnorableSpace: [\u200E\u200F];  // Pattern_White_Space AND Default_Ignorable_Code_Point
+
 // Ignore whitespace between tokens, and define C++-style comments.
-Whitespace: [ \t]+ -> skip ;
-Newline: [\r\n]+ -> skip ;
-LineComment : '//' ~[\r\n]* -> skip;
+HorizontalSpace: UAX31HorizontalSpace+ -> skip;
+Newline: UAX31Newline+ -> skip;
+LineComment : '//' NotUAX31Newline* -> skip;
 BlockComment : '/*' .*? '*/' -> skip;
 
 
@@ -170,7 +178,7 @@ BlockComment : '/*' .*? '*/' -> skip;
 // integer/floating-point literals, so we use a special mode to ensure it's
 // lexed correctly.
 mode VERSION_IDENTIFIER;
-    VERSION_IDENTIFER_WHITESPACE: [ \t\r\n]+ -> skip;
+    VERSION_IDENTIFER_WHITESPACE: (UAX31HorizontalSpace|UAX31Newline)+ -> skip;
     VersionSpecifier: [0-9]+ ('.' [0-9]+)? -> popMode;
 
 
@@ -178,14 +186,14 @@ mode VERSION_IDENTIFIER;
 // rather than the default arbitrary-whitespace-based tokenisation.  This is
 // used by the annotation and pragma rules.
 mode EAT_TO_LINE_END;
-    EAT_INITIAL_SPACE: [ \t]+ -> skip;
-    EAT_LINE_END: [\r\n] -> popMode, skip;
+    EAT_INITIAL_SPACE: UAX31HorizontalSpace+ -> skip;
+    EAT_LINE_END: UAX31Newline -> popMode, skip;
 
     // The line content must be a non-empty token to satisfy ANTLR (otherwise it
     // would be able to produce an infinite number of tokens).  We could include
     // the line ending to guarantee that this is always non-empty, but that just
     // puts an annoying burden on consumers to remove it again.
-    RemainingLineContent: ~[ \t\r\n] ~[\r\n]*;
+    RemainingLineContent: NotUAX31WhiteSpace NotUAX31Newline*;
 
 
 // We need to do a little context-aware lexing when we hit a `cal` or `defcal`
@@ -195,12 +203,12 @@ mode EAT_TO_LINE_END;
 // manage the state of the lexer, so instead we need to do a little duplication
 // of the tokens, because ANTLR doesn't allow us to inherit rules directly.
 mode CAL_PRELUDE;
-    CAL_PRELUDE_WHITESPACE: [ \t\r\n]+ -> skip;
+    CAL_PRELUDE_WHITESPACE: (UAX31HorizontalSpace|UAX31Newline)+ -> skip;
     CAL_PRELUDE_COMMENT: (LineComment | BlockComment) -> skip;
     CAL_PRELUDE_LBRACE: LBRACE -> type(LBRACE), mode(CAL_BLOCK);
 
 mode DEFCAL_PRELUDE;
-    DEFCAL_PRELUDE_WHITESPACE: [ \t\r\n]+ -> skip;
+    DEFCAL_PRELUDE_WHITESPACE: (UAX31HorizontalSpace|UAX31Newline)+ -> skip;
     DEFCAL_PRELUDE_COMMENT: (LineComment | BlockComment) -> skip;
     DEFCAL_PRELUDE_LBRACE: LBRACE -> type(LBRACE), mode(CAL_BLOCK);
 
--- a/source/grammar/qasm3Lexer.g4
+++ b/source/grammar/qasm3Lexer.g4
@@ -147,7 +147,7 @@ FloatLiteral:
     // 123.456, 123. or 145.32e+1_00
     | DecimalIntegerLiteral DOT DecimalIntegerLiteral? FloatLiteralExponent?;
 
-fragment TimeUnit: 'dt' | 'ns' | 'us' | 'µs' | 'ms' | 's';
+fragment TimeUnit: 'dt' | 'ns' | 'us' | '\u{00B5}s' | '\u{03BC}s' | 'ms' | 's';
 // represents explicit time value in SI or backend units
 TimingLiteral: (DecimalIntegerLiteral | FloatLiteral) TimeUnit;
 
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -10,10 +10,33 @@ Types and Casting
 Identifiers
 -----------
 
-Identifiers must begin with a letter [A-Za-z], an underscore or an element from
-the Unicode character categories Lu/Ll/Lt/Lm/Lo/Nl :cite:`wikipediaUnicode`.
-The set of permissible continuation characters consists of all members of the
-aforementioned character sets with the addition of decimal numerals [0-9].
+Roughly, OpenQASM identifiers start with an alphabetic character or underscore and continue with alphanumeric and underscore.
+A precise statement of the Unicode compatibility is:
+
+- `UAX31-C1 <https://www.unicode.org/reports/tr31/tr31-37.html#C1>`_: The OpenQASM language conforms to version 37 of the Unicode® Standard Annex #⁠31
+- `UAX31-C2 <https://www.unicode.org/reports/tr31/tr31-37.html#C2>`_: It observes the following requirements:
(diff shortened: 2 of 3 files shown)


jlapeyre:
@levbishop , I see above "replaces #422". Shouldn't this instead be "replaces #428"?

source/language/types.rst
```diff
@@ -487,15 +510,16 @@ single-Unicode-character identifier.
...
-      +-------------------------------+--------+--------------+---------------------+
-      | :math:`\tau = 2\pi`           | tau    | τ            | 6.283185...         |
-      +-------------------------------+--------+--------------+---------------------+
-      | Euler’s number :math:`e`      | euler  | ℇ            | 2.7182818284...     |
-      +-------------------------------+--------+--------------+---------------------+
+      +-------------------------------+--------+----------------------------------+---------------------+
+      | Constant                      | ASCII  | Unicode                          | Approximate Base 10 |
+      +===============================+========+==================================+=====================+
+      | :math:`\pi`                   | pi     | - µ U+00B5 MICRO SIGN            | 3.1415926535...     |
+      |                               |        | - μ U+03BC GREEK SMALL LETTER MU |                     |
```
jakelishman, line 517: Those are some pretty unusual representations of pi.

source/grammar/qasm3Lexer.g4
```diff
@@ -147,45 +146,54 @@ FloatLiteral:
...
 TimingLiteral: (DecimalIntegerLiteral | FloatLiteral) TimeUnit;
 
 
 BitstringLiteral: '"' ([01] '_'?)* [01] '"';
 // allow ``"str"`` and ``'str'``
 StringLiteral
-    : '"' ~["\r\t\n]+? '"'
-    | '\'' ~['\r\t\n]+? '\''
+    : '"' NotUAX31NewlineDoubleQuote+? '"'
+    | '\'' NotUAX31NewlineSingleQuote+? '\''
```
jakelishman, line 158: Previously, `\r` and `\t` were also forbidden in string literals, which this PR now allows.  I think that's totally fine, especially since the spec at no point actually even _defines_ a string literal (as opposed to a "bitstring literal", except for a somewhat implicit definition in the "included files" bit), I just wanted to mention it.

source/language/types.rst
```diff
@@ -10,10 +10,33 @@ Types and Casting
...
+   - A sequence of one or more of any of the following characters shall be interpreted as a sequence of one or more end of line:
+      -  ``U+000A`` (line feed)
+      -  ``U+000B`` (vertical tabulation)
+      -  ``U+000C`` (form feed)
+      -  ``U+000D`` (carriage return)
+      -  ``U+0085`` (next line)
+      -  ``U+2028`` LINE SEPARATOR
+      -  ``U+2029`` PARAGRAPH SEPARATOR
+   - The ``Pattern_White_Space`` characters with the property ``Default_Ignorable_Code_Point`` shall be treated as ignorable format controls
+   - All other characters in ``Pattern_White_Space`` shall be interpreted as horizontal space.
```
jakelishman, line 37: This generally seems sensible from an abstract perspective.  Is it worth mentioning / calling out that we don't actually expect any near-term implementers to have fully implemented all the Unicode normalisation?

Also, this text technically permits the sequence `U+OOOA U+OOOA` to represented 1, 2 or really any natural number of "ends of line".  I don't know how much you care about trying to define something unambiguous that's sensible here, though, since even the old `\n` / `\r\n` / `\r` split isn't the easiest thing in the world to define.

jakelishman (commented):
I know I haven't been in meetings any time recently, so I might not have all the context, but since I was asked by Lev, here's a couple of comments on the text.

In terms of code of the ANTLR-based parser, if I were reviewing something for a product that I owned, I'd complain about a lack of testing here, and that the normalisation hasn't actually been implemented.  I don't necessarily think that that should actually stand in the way of the spec change, though, given how complex it is.

source/grammar/qasm3Lexer.g4
jlapeyre: String literals supporting a string data type will be essentially impossible to add later if desired (I think likely) anyway. Because of bitstring literals.

levbishop:
> @levbishop , I see above "replaces #422". Shouldn't this instead be "replaces #428"? If so, I'll close #428.

Yes. I meant "closes #422 and replaces #428"

source/language/types.rst
levbishop: oops i'll fix

PR #480, opened 2023-09-12
State that whitespace before paren is not allowed in subroutine call

jlapeyre:
In other words
```
f (x, y)   # Not allowed
```
```
f(x, y)   # allowed
```
It's better to make this explicit.  Implementers often allow this at first.


jlapeyre:
I'm not sure I support merging this. Leaning towards preferring not merging it.

One argument is that if `f` `(` `x` , and `)` are all tokens, then whitespace should not be arbitrarily disallowed in one place.

jlapeyre:
Closing without merging, after discussing in TSC meeting today.

There was some support for closing this PR without merging.

No one spoke in favor of merging it.

If we want to continue calling the function name, the parens, and the arguments separate tokens (or compound tokens), then disallowing whitespace between the first two tokens would be the only exception in the language to the rule that whitespace is ignored.

PR #482, opened 2023-10-04
Allow identifiers in `measureExpression`

braised-babbage:
### Summary

The current syntax for measurement precludes the usage of defcals with return values, as discussed in this issue: https://github.com/openqasm/openqasm/issues/481

This PR extends `measureExpression` to allow for general identifiers. For example, this supports the following

```
measure_iq q;
measure_iq q -> c[0];
c[0] = measure_iq q[0];
```

### Details and comments

This is the smallest change I could think of to support this functionality. There's one wart, which is that `measure_iq q;` is ambiguous and resolves to a gate application rather than a quantum measurement.

### DIFF
--- a/source/grammar/qasm3Parser.g4
+++ b/source/grammar/qasm3Parser.g4
@@ -155,7 +155,7 @@ expression:
 // not in the expression tree, but can contain elements that are within it.
 aliasExpression: expression (DOUBLE_PLUS expression)*;
 declarationExpression: arrayLiteral | expression | measureExpression;
-measureExpression: MEASURE gateOperand;
+measureExpression: ( MEASURE | Identifier ) gateOperand;
 rangeExpression: expression? COLON expression? (COLON expression)?;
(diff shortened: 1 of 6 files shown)


hodgestar:
This was discussed in today's TSC meeting and generally looked on favourably. To fully evaluate it we need:

- An update / clarification to the specification.
- To think about the best way to update the grammar and parser to produce the clearest AST with the least disruption.

braised-babbage:
I still plan on updating this, but will mention that one workaround is to use function calls and subroutines instead of measure expressions & defcals, at least when the qubits are known at compile time. For example,

```
// instead of measure_iq $0 -> c[0] 
// or c[0] = measure_iq $0
c[0] = measure_iq_q0(); 
```

levbishop:
Re-considered at 2024-10-16 TSC meeting, that this is still a broadly-desirable change, the original author does eventually plan to finish the needed changes to bring this out of draft but in the meantime if someone else has need for this functionality they are welcome to pick up the work.

PR #490, opened 2023-11-09
Allow pragmas outside of global scope

braised-babbage:
From [the spec on pragma](https://openqasm.com/language/directives.html#pragmas),

> Pragma directives start with pragma and continue to the end of line. The text after pragma is a single string, and parsing is left to the specific implementation. Implementations may optionally choose to support the older #pragma keyword as a custom extension. Pragmas should be processed as soon as they are encountered; if a pragma is not supported by a compiler pass it should be ignored and preserved intact for future passes. Pragmas should avoid stateful or positional interactions to avoid unexpected behaviors between included source files. **If the position is relevant to a pragma, an annotation should be used instead.**

The grammar places no restrictions on where pragmas may appear. However, the Python reference parser currently raises an exception if a pragma occurs outside of the global scope. I do not see the point of this restriction. If the meaning of a pragma is independent of its position, then one could simply allow them anywhere. 

This PR changes the parser to not treat pragmas outside of the global scope as an error. I have added a clarifying line to the spec indicating that this is allowed.

To slightly motivate why I prefer the flexibility of a "pragma anywhere" approach, I will say that we often work with OpenQASM programs which are produced by mechanically combining various AST fragments. If any of these AST fragments includes a pragma, then their composition is not so simple. For example, we may wish to produce a defcal "on-the-fly", and as it stands the restriction on `pragma` placement requires that any pragmas associated with the defcal get percolated to the program toplevel via some side channel. It would be much more convenient to allow defcal bodies to carry pragma statements. To give another example, we sometimes take a sequence of statements and wrap this in a loop (e.g. to scan over a parameter as part of a larger experiment) -- again life is simpler if pragmas don't have to bubble up to the toplevel.

### DIFF
--- a/source/language/directives.rst
+++ b/source/language/directives.rst
@@ -15,9 +15,10 @@ Pragma directives start with ``pragma`` and continue to the end of line. The
 text after ``pragma`` is a single string, and parsing is left to the specific
 implementation. Implementations may optionally choose to support the older ``#pragma``
 keyword as a custom extension.
-Pragmas should be processed as soon as they are encountered; if a
-pragma is not supported by a compiler pass it should be ignored and preserved
-intact for future passes.  Pragmas should avoid stateful or positional
+
+Pragmas may appear anywhere in the program. Pragmas should be processed as soon as they 
+are encountered; if a pragma is not supported by a compiler pass it should be ignored 
+and preserved intact for future passes.  Pragmas should avoid stateful or positional
 interactions to avoid unexpected behaviors between included source files. If the
 position is relevant to a pragma, an annotation should be used instead.
 
--- a/source/openqasm/openqasm3/parser.py
+++ b/source/openqasm/openqasm3/parser.py
@@ -215,8 +215,6 @@ class QASMNodeVisitor(qasm3ParserVisitor):
 
     @span
     def visitPragma(self, ctx: qasm3Parser.PragmaContext):
-        if not self._in_global_scope():
-            _raise_from_context(ctx, "pragmas must be global")
         return ast.Pragma(
             command=ctx.RemainingLineContent().getText() if ctx.RemainingLineContent() else None
         )
--- a/source/openqasm/tests/test_qasm_parser.py
+++ b/source/openqasm/tests/test_qasm_parser.py
@@ -1740,6 +1740,9 @@ def test_pragma():
     pragma command arg1 arg2
     #pragma otherwise_invalid_token 1a2%&
     pragma otherwise_invalid_token 1a2%&
+    {
+      pragma another_command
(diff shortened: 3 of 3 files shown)


blakejohnson (approved):
Makes sense to me.

jwoehr:
If a pragma appears outside global scope, is its impact scoped locally or globally?

jakelishman:
Jack: if I recall correctly, that question is the reason the spec originally only allowed pragmas at the global scope. I'm fairly sure it was discussed, and there wasn't at the time a known use-case for local pragmas, so they were made global-only in order to avoid needing to answer that question.  I think it was approximately assumed that annotations could serve a more local purpose.

That said, I don't feel strongly, and my recollection is that the original reason for not allowing it was a lack of compelling use-case more than a technical reason.  Provided we do define how a local pragma is to be interpreted (even if just to say "a compiler is free to interpret this entirely however it likes"), I'm fine with it.

jwoehr:
> **Provided we do define how a local pragma is to be interpreted** (even if just to say "a compiler is free to interpret this entirely however it likes"), I'm fine with it.

Yes, that is the important point. Is a local pragma local in effect or global?

braised-babbage:
> > **Provided we do define how a local pragma is to be interpreted** (even if just to say "a compiler is free to interpret this entirely however it likes"), I'm fine with it.
> 
> Yes, that is the important point. Is a local pragma local in effect or global?

My view is that a pragma should be a kind of global signal to the compiler, regardless of whether it appeared in some particular lexical scope. If at some point we introduce a notion of separate compilation, I would be happy to consider pragmas applying to a particular compilation unit.

Probably the most ubiquitous notion of "pragma" is the one from C/C++ where the effect is global but position matters (i.e. pragmas apply as the file is processed top-to-bottom). Some of these are truly used as global signals, and others are just turned on and off for some portion of the program. In this respect I think the OpenQASM notion of annotation is a nice alternative mechanism to communicate information to the compiler with a clear & bounded extent. But if pragmas apply only to the scope block they are in, they would not offer much beyond what we can do with annotations.

levbishop (changes requested):
I approve in principle after a reno and ideally a test

braised-babbage:
Based on the comments above and TSC discussion today (8/7/24), I am going to withdraw this PR, at least in its current form. My observation is that people interpreting the spec may come with various notions of what pragmas mean / are for, and we have relatively few concrete examples to pin discussions on. Because the effect of pragmas is independent of their position, there is a risk of confusion amongst human readers when pragmas appear some place other than at the top of a QASM file. There may still be room for spec clarifications, but that does not need to happen in this PR.

PR #495, opened 2023-11-21
Move 'Looping and branching' up one heading level

hodgestar:
### Summary

Previously the [Looping and branching](https://openqasm.com/language/classical.html#looping-and-branching) section was nested inside [Low-level classical instructions](https://openqasm.com/language/classical.html#low-level-classical-instructions). Control flow is an important and distinct part of OpenQASM that deserves to be visible in the table of contents and separated from other operations such as defining classical variables.

### Details and comments

I moved the sub-headings up one level too.

### DIFF
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -249,10 +249,10 @@ OpenQASM evaluates expressions from left to right.
 
 
 Looping and branching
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
This seems like a good idea to me - I can never find this section when I need it because of the document hierarchy.

PR #498, opened 2023-12-04
Issue #486: Clarify interpretation of physical qubits

### LINKED ISSUE #486: [proposal]: Clarify interpretation of physical qubits

From the documentation (https://openqasm.com/language/types.html#physical-qubits):

> While program qubits can be named, hardware qubits are referenced only by the syntax $[NUM]. For an n qubit system, we have physical qubit references given by $0, $1, …, $n-1. These qubit types are used in lower parts of the compilation stack when emitting physical circuits. Physical qubits must not be declared and they are, as all the qubits, global variables.

At AWS, we accept OpenQASM programs in the service, and we allow users to submit programs with physical qubits. As a developer interpreting OpenQASM semantics for our customers, I have the following questions and proposals:

1. I propose that we relax the requirement that physical qubit indices must be consecutive. There are some quantum devices that are not labeled with consecutive indices, such as Rigetti's Aspen devices.
2. What is the definition of physical circuits?
    1. My proposal, and assumption for the remainder of this issue: A physical circuit only has operations which are natively executable by the device without further decomposition. This includes that multi-qubit operations can be natively executed directly on that set of qubits, without additional SWAPs or routing between disconnected qubits.
    2. The documentation states that physical qubits are needed for representing physical circuits. Can physical qubits also appear beyond physical circuits? If the answer is no, that implies that the answer to the following questions — about mapping, routing, and non-native gates, respectively — is also no.
3. If a program is has physical qubits, can those qubits be remapped by a compiler? I presume no, or else virtual qubits would be more appropriate.
4. Can a two-qubit operation be applied to two qubits that aren’t directly connected? In other words, can routing be done for a program with physical qubits? I expect this to be valid OpenQASM code, but ultimately up to the compiler / hardware provider whether to accept it.
    1. For example, `CNOT $0, $2;` can be realized on a device with topology `0 <-> 1 <-> 2` by routing: `SWAP $1, $2; CNOT $0, $1; SWAP $1, $2;`
5. Can physical qubits be used with non-native gates? I expect this to also be valid, but again left to the discretion of the hardware provider whether to accept it.
    1. For example, `H $0`; can be realized on a device with a basis gate set of `{RX(pi/2), RZ(theta)}` by compiling to `RZ(0.5*pi) $0; RX(0.5*pi) $0; RZ(0.5*pi) $0;`

laurencap:
### Summary

Adds clarifications about physical qubits, and closes #486

### Details and comments

 - Separates virtual and physical qubit examples
 - Relaxes the consecutive numbering constraint
 - Adds definition for physical circuit: one where every circuit operation has a defcal
 - Physical qubits can be seen as a "no remap" directive (or else they would be virtual qubits)
 - Physical qubits can be used outside of physical circuits

Open questions
 - ~Is there a better place for physical circuit discussion? And is the physical circuit text sufficient as is, or should "executable circuit" be introduced as another term for physical circuit?~
 - ~Can physical qubits be used in a defgate?~

See more discussion on #486

### DIFF
--- a/releasenotes/notes/clarify-physical-qubits-f720f21a3d792c70.yaml
+++ b/releasenotes/notes/clarify-physical-qubits-f720f21a3d792c70.yaml
@@ -1,0 +1,6 @@
+---
+upgrade:
+  - |
+    Improve physical qubit documentation. Define what constitutes a valid
+    physical identifier, define physical circuits, and answer several questions
+    about what a compiler may do with physical qubits.
--- a/releasenotes/notes/clarify-physical-qubits-f720f21a3d792c70.yaml
+++ b/releasenotes/notes/clarify-physical-qubits-f720f21a3d792c70.yaml
@@ -2,5 +2,5 @@
 upgrade:
   - |
     Improve physical qubit documentation. Define what constitutes a valid
-    physical identifier, define physical circuits, and answer several questions
-    about what a compiler may do with physical qubits.
+    physical identifier, define physical circuits, and specify what
+    an implementer may do with physical qubits.
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -142,16 +142,16 @@ non-consecutive, depending on the device.
 Like virtual qubits, physical qubits are global variables, but unlike virtual qubits, they must
 not be declared.
 
-These qubit types are often used for ``defcal`` blocks because calibrations are typically valid
-only for a particular set of physical qubits (see also :ref:`pulse gates <pulse-gates>`).
+These qubit types are always used for ``defcal`` blocks. Calibrations are valid only for a
+particular set of physical qubits. See also :ref:`pulse gates <pulse-gates>`.
 
-.. TODO: Where is a better place for physical circuit discussion?
-.. TODO: Should we introduce executable circuit as another term for physical circuit?
+Physical qubits cannot be used in ``gate`` statements. See also
+:ref:`gate definitions <gate-statement>`.
 
 Physical qubits are also used in lower parts of the compilation stack when emitting physical
 circuits. A physical circuit is one which only references physical qubits, and every operation
-used in the circuit has an associated ``defcal``, which we can call hardware-native gates or
-measurements. A physical circuit can be directly executed on the target hardware.
+used in the circuit has an associated ``defcal``, which we refer to as hardware-native gates and
+measurements.
 
 .. code-block::
 
@@ -162,7 +162,6 @@ measurements. A physical circuit can be directly executed on the target hardware
 
 Physical qubit constraints
 ..........................
-.. TODO: Can physical qubits be used in a ``defgate``? My answer would be no.
 
 Physical qubits, by definition, reference particular hardware qubits. Circuit equivalence does not
 hold over permutations of physical qubit labels. Thus, physical qubits cannot be remapped by a
@@ -173,9 +172,10 @@ that would require routing or gate decomposition to run (i.e., does not have a `
 operation in the circuit) would by definition not be a physical circuit. However, physical qubits
 can still be used in such circuits.
 
-For example, a program defines the ``H`` gate with the ``gate`` statement. ``H`` is therefore a
-supported gate, but not a hardware-native gate. The compiler can decompose the statement ``H $0;``
-to hardware-native gates, while still respecting strict qubit mapping.
+For example, a program defines the ``H`` gate with the ``gate`` statement, without a corresponding
+``defcal`` of ``H``. ``H`` is therefore a supported gate, but not a hardware-native gate. The
+compiler can decompose the statement ``H $0;`` to hardware-native gates, while still respecting
+strict qubit mapping.
 
 It is possible to write a partially-constrained program with both physical and virtual qubits.
 Such programs are also non-physical circuits, and may or may not be supported by compilers or
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -116,17 +116,6 @@ significantly since there is no need for quantum memory management.
 However, it also means that users or compiler have to explicitly manage
 the quantum memory.
 
-.. _physical-qubits:
-
-Physical Qubits
-~~~~~~~~~~~~~~~
-
-While program qubits can be named, hardware qubits are referenced only
-by the syntax ``$[NUM]``. For an ``n`` qubit system, we have physical qubit
-references given by ``$0``, ``$1``, ..., ``$n-1``. These qubit types are
-used in lower parts of the compilation stack when emitting physical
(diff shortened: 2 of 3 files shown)


source/language/types.rst
```diff
@@ -135,8 +124,63 @@ circuits. Physical qubits must not be declared and they are, as all the qubits,
...
+.. code-block::
+
    // CNOT gate between physical qubits 0 and 1
    CX $0, $1;
+   // Define the pulse-level instruction sequence for ``h`` on physical qubit 0
+   defcal h $0 { ... }
+
+Physical qubit constraints
+..........................
+.. TODO: Can physical qubits be used in a ``defgate``? My answer would be no.
```
laurencap, line 165: Update: Erik gave a good argument for why it would be possible, so I'm inclined to revise this.

source/language/types.rst
```diff
@@ -135,8 +124,63 @@ circuits. Physical qubits must not be declared and they are, as all the qubits,
...
+
+While virtual qubits can be named, hardware qubits are referenced by the syntax ``$[NUM]``, where
+``NUM`` is a non-negative integer. Any integer included in the published device topology is a valid
+physical qubit identifier. Note that this implies that physical qubit identifier indices may be
+non-consecutive, depending on the device.
+
+Like virtual qubits, physical qubits are global variables, but unlike virtual qubits, they must
+not be declared.
+
+These qubit types are often used for ``defcal`` blocks because calibrations are typically valid
```
braised-babbage, line 145: Just to clarify, defined calibrations apply only to physical qubits (cf. https://openqasm.com/language/pulses.html). The OpenQASM spec allows for a family of calibrations to be defined using a shorthand syntax, like `defcal rz(angle theta) q { ... }` but technically the `q` here is not a virtual qubit -- it's a kind of formal parameter.

> One can define a defcal using a regular identifier for a qubit, which causes that calibration definition to be valid for all physical qubits. This is most likely to be useful for gates that are implemented virtually.

laurencap: Awesome, will update.

source/language/types.rst
```diff
@@ -135,8 +124,63 @@ circuits. Physical qubits must not be declared and they are, as all the qubits,
...
+compiler or hardware provider without opt-in from the programmer.
+
+Note that while physical circuits require physical qubits, the converse need not be true. A circuit
+that would require routing or gate decomposition to run (i.e., does not have a ``defcal`` for every
+operation in the circuit) would by definition not be a physical circuit. However, physical qubits
+can still be used in such circuits.
+
+For example, a program defines the ``H`` gate with the ``gate`` statement. ``H`` is therefore a
+supported gate, but not a hardware-native gate. The compiler can decompose the statement ``H $0;``
+to hardware-native gates, while still respecting strict qubit mapping.
```
levbishop, line 178: A gate can (and often will have both a `defcal` and a `gate` definition. The `defcal` gives the physical implementation and the `gate` specifies the ideal unitary. I think we fix keep this para like this

```suggestion
For example, a program defines the ``H`` gate with the ``gate`` statement,
without a corresponding ``defcal`` of ``H``. ``H`` is therefore a
supported gate, but not a hardware-native gate. The compiler can decompose the statement ``H $0;``
to hardware-native gates, while still respecting strict qubit mapping.
```

laurencap:
Addt comments from meeting:

- Blake
  - physical qubit in gate decl: Answer would be no. Interpreted as a mathematical equivalence
    - Erik: makes sense, hardware diffs can be represented in defcals
 - Thomas
    - useful to have non-integer physical ids?
   - in favor of being able to declare
   - Me: both good points, let's revisit in another PR

releasenotes/notes/clarify-physical-qubits-f720f21a3d792c70.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade:
+  - |
+    Improve physical qubit documentation. Define what constitutes a valid
+    physical identifier, define physical circuits, and answer several questions
+    about what a compiler may do with physical qubits.
```
laurencap, line 6:
```suggestion
    physical identifier, define physical circuits, and specify what
    an implementer may do with physical qubits.
```

levbishop (dismissed):
Lgtm

source/language/types.rst
```diff
@@ -135,8 +124,63 @@ circuits. Physical qubits must not be declared and they are, as all the qubits,
...
    // CNOT gate between physical qubits 0 and 1
    CX $0, $1;
+   // Define the pulse-level instruction sequence for ``h`` on physical qubit 0
+   defcal h $0 { ... }
+
+Physical qubit constraints
+..........................
+
+Physical qubits, by definition, reference particular hardware qubits. Circuit equivalence does not
+hold over permutations of physical qubit labels. Thus, physical qubits cannot be remapped by a
```
braised-babbage, line 167:
```suggestion
hold over permutations of physical qubit labels. Thus, physical qubits should not be remapped by a
```

braised-babbage (dismissed):
Looks good to me.

braised-babbage (approved):

blakejohnson (approved):
I'm good with these changes.

blakejohnson:
Failing checks. Will need to investigate.

source/language/types.rst
```diff
@@ -135,8 +124,63 @@ circuits. Physical qubits must not be declared and they are, as all the qubits,
...
+While virtual qubits can be named, hardware qubits are referenced by the syntax ``$[NUM]``, where
+``NUM`` is a non-negative integer. Any integer included in the published device topology is a valid
+physical qubit identifier. Note that this implies that physical qubit identifier indices may be
+non-consecutive, depending on the device.
+
+Like virtual qubits, physical qubits are global variables, but unlike virtual qubits, they must
+not be declared.
+
+These qubit types are always used for ``defcal`` blocks. Calibrations are valid only for a
+particular set of physical qubits. See also :ref:`pulse gates <pulse-gates>`.
```
jlapeyre, line 146: *These qubit types* (which is copied from the previous text) should be *This qubit type*. But I think it would be better to make this even more precise. It's not clear to me what the sentence means, in particular *used for*. Does it mean that the qubit parameters in a `defcal` declaration must be physical qubits? Or that the qubit arguments in a gate call that dispatches to a `defcal` "method" must be physical qubits? And/or that the qubits appearing in the body of a `defcal` declaration must be physical qubits?

I guess that it would be easier and more automatic to write precise and concise statements if we had an unambiguous language to talk about the language elements. For example, using the conventional programming language distinction between *parameter* and *argument*. The spec makes this distinction sometimes, but not always. Pursuing this observation belongs in another issue, but I think we could still make an improvement to this text.

blakejohnson:
Closed in favor of #505.

PR #499, opened 2023-12-12
Fix formatting in switch section, add extra example

braised-babbage:
This does two things:
- it fixes some small formatting issues in the recently added section on `switch` statements
- it adds an extra example in which the switch cases use a `const` expression

### DIFF
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -467,7 +467,7 @@ A ``case`` statement shall be the following grammar:
 - A sequence of zero, one or more OpenQASM3 statements.
 - A right-brace literal: ``}``.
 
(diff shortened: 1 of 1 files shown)


source/language/classical.rst
```diff
@@ -540,8 +566,9 @@ Examples:
 	case 0b11 {
 	  // OpenQASM3 statement(s)
 	}
-	}```
```
braised-babbage, line 543: The backticks here were a typo.

jakelishman (approved):
This looks good to me, and the additional example would for sure makes the intention clearer.

blakejohnson (approved):

PR #502, opened 2023-12-14
Fail workflow on documentation build failures

jakelishman:
### Summary

The documentation-build script did not propagate failures from individual lines, which meant that a failure in the script could still result in an attempted empty deployment, which deletes the whole website. This ensures that the bash script propagates the failure through to the workflow status, so the deployment does not trigger.

We also fix the failure that took out the website last time - a missing transitive dependency (which isn't our fault, but is easy to work around).

### Details and comments

Since I was at it, and apparently I hooked `shellcheck` in to vim at some point, I quoted the argument expansions it was moaning about.

I ran a dummy run of this commit with the deployment replaced with an artefact upload on my fork, which you can see ran successfully here: https://github.com/jakelishman/openqasm/actions/runs/7211750781.

### DIFF
--- a/buildPublishable.sh
+++ b/buildPublishable.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/bash
 
+set -eux -o pipefail
+
 ## This script builds the publishable version of the specification including
 ## all stable branches. Each is built successively in the normal way after
 ## a `git checkout` of the found stable branches. Each built stable version is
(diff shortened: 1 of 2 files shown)


buildPublishable.sh
```diff
@@ -16,20 +18,20 @@ echo "Searching for stable branches with repo path refs/${repoPath}/stable/"
...
 fi
 
 # initialize the destination folder
-mkdir -p ${destDir}/versions
+mkdir -p "${destDir}/versions"
 
 # Build the stable version list
-unset versionList
-for branch in `git for-each-ref --format='%(refname:short)' --sort=-refname refs/${repoPath}/stable/`; do
+versionList=""
```
jakelishman, line 33: with `set -u`, `unset versionList` would have caused a failure in the `[ -z "$versionList" ]` expansion.  I suppose I could have used `[ -z ${versionList:-} ]`, but I didn't think of it at first - happy to change it if that's more legible.

jakelishman:
The AST CI failures in this one are fixed by #503.

jlapeyre (approved):
Thanks

blakejohnson:
Thanks, @jakelishman

PR #503, opened 2023-12-14
Update Python versions and dependencies in AST GitHub Actions

jakelishman:
### Summary

This normalise the AST action to only run the style and lint checks once, on the newest Python version, so that differences between the linters and formatters that might arise from bugfixes not backported to the last supported package version for the oldest Python do not cause spurious failures between the two branches.  This notably happened with `black`, where the latest version compatible with Python 3.7 did not contain a bugfix that later versions did have access to.

Further, this updates the latest supported version of Python to 3.12, which matches reality, since previously we had left it at 3.10.

### DIFF
--- a/.github/workflows/tests-ast.yml
+++ b/.github/workflows/tests-ast.yml
@@ -15,18 +15,18 @@ jobs:
       fail-fast: false
       matrix:
         # Just using minimum and maximum to avoid exploding the matrix.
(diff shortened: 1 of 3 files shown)


jakelishman:
The docs-build failure in this PR is fixed by #502.

jlapeyre (approved):
Thanks

PR #508, opened 2024-01-16
Update contributing guidelines to require package and language specification versions to match

hodgestar:
### Summary

At the TSC meeting on 10 January 2024 those present discussed how to version our packages for the upcoming and future release.

Currently there is only one version of the language specification -- i.e. `3.0` -- and our package releases have been numbered `0.X`.

Those present proposed that for future releases the language specification and package version numbers should match.

This PR documents that proposal. Some additional work will be needed to implement it if the proposal is accepted. See details below.

### Details and comments

If this PR was accepted we would need to:

- [ ] Release version 3.0.0 (corresponding to commit 51c36946c687c8b17000962f6ce735ca1d1c9b3b on stable/3.0 branch and the same as PyPI release 0.5).
- [ ] Release version 3.0.1 (corresponding to commit f558486430cabeb600022085692ba17fb078e44b with commit message `Add anonymous scope blocks`).
- [ ] Release version 3.1.0rc0 (which adds the switch statement).
- [ ] Check that GitHub and PyPI releases match.
   
Questions:

- Any considerations I missed here or that we missed in the meeting?
- Any further clarifications needed to the documentation?
- Any other proposals for what release updates to do afterward we merge this?
- Votes from TSC?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,9 +80,27 @@ The development cycle for OpenQASM is managed in the open using Github for proje
 TODO: When preparing a new release changes for the released version should be identified
 in the release notes (See [issue #328](https://github.com/openqasm/openqasm/issues/328)).
 
-### Semantic Versioning
-The OpenQASM language uses [semantic versioning (semver)](https://semver.org/).
-All official releases are identified by a valid semver (See [Tags](#tags)).
+### Versioning
+
+The OpenQASM language specification and packages use [semantic versioning (semver)](https://semver.org/).
+
+The language specification version is defined by the major and minor version numbers.
+For example, versions `3.0.X` are all the same language. Patch numbers are used for
+errata and clarifications.
+
+All official package releases are identified by a valid semver, and marked with a
+Git tag (See [Tags](#tags)).
+
+The GitHub release, PyPI release, Git tag and OpenQASM language version should all
+match. For example, Git tag `3.0.1` would corresepond to PyPI release `3.0.1`, and
+OpenQASM language version `3.0`.
+
+Package patch releases may contain non-breaking improvements and changes, but must
+still correspond to the same language specification (as described above).
+
+Package releases with version `0.X` predate this policy and all correspond to language
+version `3.0`.
+
 The latest development branch (See [Branches](#branches)) is always identified
 by the semver `<next_major>.<next_minor>.0-dev` where `next_<major/minor>` are
 the target major/minor versions of the next release.
@@ -96,14 +114,15 @@ branch, the language specification can and will change (possible breaking)
 as new language features are introduced and refined.
 All efforts should be made to ensure that the development branch is maintained in
 a self-consistent state that is passing continuous integration (CI).
-Changes should not be merged unless they are verified by CI. TODO: The latest
-development specification of the language should be automatically published by CI
-to a fixed URL for easy access to the current development HEAD.
+Changes should not be merged unless they are verified by CI.
+TODO: The latest development specification of the language should be automatically
+published by CI to a fixed URL for easy access to the current development HEAD.
+
 * `stable/<major.minor>` branches:
 Branches under `stable/<major.minor>` are used to maintain released versions of the OpenQASM
 specification. They contain the version of the specification corresponding to the
 release as identified by its [semantic version](https://semver.org/). For example,
-stable/3.2 would be the specification version for major version 3
+`stable/3.2` would be the specification version for major version 3
 (corresponding to OpenQASM3) and minor version 2. On these branches, the language specification
 is considered stable. The only changes that may be merged to a stable branch are
 patches/bugfixes. When a patch is required when possible the fix should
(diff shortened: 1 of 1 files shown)


jakelishman:
Is the intent to apply package versions like `3.0.x` to the `openqasm3` package on PyPI?  If so, I'm not sure it's right - a parser's version isn't typically tied to a particular version of the language spec unless the parser _is_ the spec.

Parsers can often handle multiple versions of the language simultaneously, abstracting out syntax / behavioural differences so that consumers don't have to worry about it.  It also "uses" the major version field of SemVer, so the `openqasm3` can never make "major-version" changes, for example changing its AST or modifying a plugin mechanism, etc.

For example, `gcc` and `clang` aren't versioned by the latest C spec that they support, and the the `openqasm3` package is similarly explicitly non-normative about the language spec, so it shouldn't be tied to the version number.

More rolling-release languages like Rust started off life with their compiler as the normative implementation, so say `rustc` matches the version of Rust, but there's work underway to decouple that into a proper spec.  CPython is largely normative, but alternate implementations like PyPy have a "package" version and a "spec" version (e.g. 7.3.12 and 3.10.12 here):
```
jake@ninetales$ pypy3 --version
Python 3.10.12 (af44d0b8114cb82c40a07bb9ee9c1ca8a1b3688c, Jun 25 2023, 12:38:57)
[PyPy 7.3.12 with GCC Apple LLVM 14.0.3 (clang-1403.0.22.14.1)]
```

hodgestar:
@jakelishman Agreed in the case of compilers, interpreters, and generic parsers. Currently there are two difficulties with applying this logic here though:

- the specification and parser are in the same repository, so they're released together
- the parser only supports the version of the specification in the current commit (I think)

Given this, another point of view is that the repository contains the specification and the `openqasm3` package contains the reference implementation for that specification.

This repository could become just the specification, and the `openqasm3` package could become a parser that supported multiple versions, but that seems to imply at the very least splitting the repositories and thinking about how to support multiple versions in the parser.

Further thoughts? I don't personally mind either direction -- I'm still new here.

jakelishman:
I'd be fine to split the repos and move `openqasm3` somewhere else; I'm not convinced that `openqasm3` _should_ be in this repo, but at the time a couple of years ago, there was some interest in trying to make sure that additions to the AST didn't lag the spec by too much, so it just stayed here.  If I remember right, the package was also added before the `openqasm` organisation existed, so there wasn't a great alternative location for it.

> the specification and parser are in the same repository, so they're released together

I don't agree with this - the GitHub zip of the source on any given tag is going to contain both, but that is not necessarily the canonical release object, it's just something GitHub implicitly tags on.  `openqasm3` is not released by a `3.0.1` tag, it's only released by an `ast-py/v<version>` tag, which similarly does not imply a new spec version.   The release artefacts of Python packages are typically the sdist and any supplied wheels.  The release artefact of a spec release is the text of the spec.  We don't generate the latter as a separate object as part of the tagging process, but we could (it's a relatively trivial GitHub Action to write for a MWE - build the HTML and zip up the result).

> the parser only supports the version of the specification in the current commit (I think)

Yeah you're right here, though I was just giving one example of how the versions can be decoupled.  `openqasm3` doesn't define the OpenQASM 3 language, it just provides one particular programmatic way to interact with the language.  The API of the library is not related to the actual language (other than including one particular representation of it), and it's the API of that package that its version number should refer to.

For example, if we tie the version of `openqasm3` to the version of the language version, and say the `Program` AST node currently looks like:
```python
@dataclass
class Program(QASMNode):
    statements: List[Statement]
    version: Optional[str] = None
```
It would change nothing about the language were `openqasm3` to want to change its data structure to be
```python
@dataclass
class Program(QASMNode):
    statements: List[Statement]
    version: Optional[Version] = None

class Version:
    major: int
    minor: Optional[int] = None
```
(for example), but that would be a semver major API change that it could now never make until OpenQASM 4.0 (which would invalidate the name `openqasm3`!), but might be very welcomed by people trying to use the package.

The point is that the API of the `openqasm3` package (imo) ought to be free to communicate changes to itself via its version number.  If the semantic versioning of the language is applied to it, then the API is near-entirely frozen in its current state, and at best, the version number of the package has not very much meaning.

---

For sake of argument, let's say you buy my argument above that `openqasm3` and the spec should have separate version numbers (if you don't, then this bit is irrelevant).  If the concern is now predominantly that a "tag" on this repo feels like it implies a particular version of the `openqasm3` reference package, I'd say that this repo is (currently) acting like a monorepo, and it's not uncommon in monorepos to have separated tagging strategies.

For example, in Rust land, `crossbeam` is a suite of crates for concurrent programming in Rust, and [it uses a monorepo with several crates developed jointly](https://github.com/crossbeam-rs/crossbeam/).  Still, though, [the tags themselves distinguish which component they're tagging by using a prefix](https://github.com/crossbeam-rs/crossbeam/tags). In this repo, the versions of the parser are tagged with a prefix `ast-py/v`, and the spec is just unprefixed.

If we think the unprefixed nature of the spec tags is confusing, we could start tagging those as `spec/v<version>` and bundling a reduced release artefact along with them in the GH UI, if you think that would be clearer?

jakelishman:
> I'm still new here.

That doesn't make your opinion any less valid than mine! If nothing else, you've been in the TSC meetings recently and I haven't been for quite a while now.

braised-babbage:
The motivation behind having the spec versions and the Python package versions synced was to make the relationship between these clear, and to reduce the risk of finding ourselves in our our current situation. At present, the pypy package is stale and doesn't incorporate all of the bug fixes associates with 3.0 features, doesn't clearly correspond to any particular snapshot of the spec, and at the same time this repo has gone beyond OpenQASM 3.0 with new features `switch` and anonymous blocks. The significance of this depends a bit on what we mean by `openqasm3` being a "reference" for the spec. Maybe we should stop saying that.

But I do find @jakelishman's arguments persuasive, and I don't think solving the problems I mentioned above _requires_ agreement between spec and package versions. We can probably go a long ways by just adopting a more regular release process for the `openqasm3` package, and clearly documenting how python package versions relate to spec versions.

hodgestar:
Should we table this for the agenda of the next TSC meeting and ask those who would like to discuss it to attend? It would be helpful to set a clear direction for the future.

braised-babbage:
> Should we table this for the agenda of the next TSC meeting and ask those who would like to discuss it to attend? It would be helpful to set a clear direction for the future.

That seems reasonable. @jlapeyre can add this to the agenda

jlapeyre:
I agree heartily with my paraphrases of the arguments above:
* Jake: The reference parser is a software artefact. A semantic version attached to the artefact should identify fixed "observable" behavior: A fixed API and a fixed map of input to output. Changes to the spec may of course force a change in the behavior of the reference parser. Logically the meaning of a version of the spec and a version of the reference implementation dictate that they vary independently. And practically, they must as well.
* Erik's story: For neither the reference parser, nor the OQ spec has a procedure for updating version numbers attached to artefacts been articulated or put into practice. But the spec and the parser keep evolving. The need to address this deficiency is growing.

Further points
* It's been suggested that perhaps the parser should not be called a "reference" parser. But I think it's useful to have a reference parser and that should remain the main mission of the parser in this repo.
* The fact that the spec and the parser needing different versions has not been articulated clearly is in part responsible for the lack of timely and clear versioning and releases in practice. (That's almost a tautology)
* It's been suggested that reference parser could support flags requesting support for a particular version of the spec. Like gcc's [--std]( https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html#index-std-1).
    *  It was also pointed out that we don't have the resources to implement and maintain such flags.
    * It was suggested that we maintain simple, plain text or markdown file describing the relation between versions of the spec and the parser.
    * I think it may be a better idea to include information on which versions of the spec are supported in the parser itself. At least in an artefact that is distributed as the parser. In fact, it should be encoded programatically. But this could be quite simple. For example each version of the parser library has a fixed mapping of input to output. But it supports a query that returns a list of or range of spec versions that it supports.
    * It was pointed out that the parser at the moment only verifies (as part of its job) the syntax of OQ source, not the semantics. This makes it more likely that a single version of the parser supports multiple versions of the spec that differ in semantics but not in syntax.

hodgestar:
Superceded by #513.

hodgestar:
Superceded by #513.

PR #509, opened 2024-01-18
Update markdown documents to reflect results of TSC election of December 2023

jlapeyre:
This PR updates the markdown documents at the top level of the repo to reflect the results of the TSC election.

**Please remember** Do a squash-merge so that all the commits are squashed into one.

The TSC election closed on January 15, 2024.

The three candidates elected to the TSC were Erik Davis, Jialin Dou, and Simon Cross. The two outgoing members of the TSC are Bettina Heim and Philipp Schindler.

* Update governance.md to reflect TSC election of Dec 2023
* Update contributors.md to reflect results of Dec 2023 TSC election
* Remove NOMINATIONS.md for TSC election Dec-2023 Jan-2024
* Remove nomination and election notices from README.md

PR #511, opened 2024-01-23
Reno: only require release notes in changes to `source` directory

braised-babbage:
### Summary

In https://github.com/openqasm/openqasm/pull/506 we started using Reno for managing release notes. Along with this is check that notes are provided when there is a `git diff`. This PR tweaks that to only check for changes to the `source` directory (i.e. changes to the spec or parser & reference AST). My reasoning is that we do not need to provide release notes for changes to things like GitHub actions, project README, etc, so this check should make an effort to ignore such changes.

### DIFF
--- a/.github/renos_updated.sh
+++ b/.github/renos_updated.sh
@@ -3,8 +3,15 @@
 
 git fetch origin main
 
-CHANGED_FILES=$(git diff --name-only origin/main $GITHUB_SHA)
-for file in $CHANGED_FILES
+# We only require release notes for changes in the `source` directory.
+# Other stuff (governance info, github actions, etc) can be ignored.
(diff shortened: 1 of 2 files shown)


taalexander (commented):
Perhaps we should test this with a trivial update to a file in the source directory?

braised-babbage:
> Perhaps we should test this with a trivial update to a file in the source directory?

Ok, I tried it out, and saw the following failure (as expected) https://github.com/openqasm/openqasm/actions/runs/7631739064/job/20790381370. I've reverted the change in my latest commit.

taalexander (approved):
Thanks Erik!

jlapeyre:
* See #509

for an example of why we need this PR.

PR #512, opened 2024-01-31
Remove reserved keywords

braised-babbage:
This removes the subsection on reserved keywords. With the inclusion of the `switch` statement, we no longer have any keywords reserved which are not already in use by the language.

### DIFF
--- a/.github/renos_updated.sh
+++ b/.github/renos_updated.sh
@@ -11,7 +11,8 @@ if [ -z $CHANGED_SOURCE_FILES ]; then
     exit 0;
 fi;
 
-for file in $CHANGED_SOURCE_FILES
+CHANGED_RELEASE_NOTES=$(git diff --name-only origin/main $GITHUB_SHA -- releasenotes)
+for file in $CHANGED_RELEASE_NOTES
 do
    root=$(echo "./$file" | cut -d / -f 3 )
    if [ "$root" = "notes" ]; then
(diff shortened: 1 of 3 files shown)


jakelishman (dismissed):
This seems obviously correct.

jakelishman:
oh, I guess we have a CI job that requires release notes now?  That's fair - it's probably worth mentioning.

.github/renos_updated.sh
```diff
@@ -11,7 +11,8 @@ if [ -z $CHANGED_SOURCE_FILES ]; then
     exit 0;
 fi;
 
-for file in $CHANGED_SOURCE_FILES
+CHANGED_RELEASE_NOTES=$(git diff --name-only origin/main $GITHUB_SHA -- releasenotes)
+for file in $CHANGED_RELEASE_NOTES
```
braised-babbage, line 15: In a previous [PR](https://github.com/openqasm/openqasm/pull/511), I broke this script. `CHANGED_SOURCE_FILES` contains only changes to the `source` directory, so no notes would be detected. The above should fix this.

jakelishman: Easily enough done.  I guess at some point we might need to look into writing a manual "I know what I'm doing, skip this check" label or something, but definitely not in this PR.

jakelishman (approved):
Looks good to me, thanks for the fix to the reno script too!

PR #513, opened 2024-02-14
Add supported_versions to openqasm3.spec.

hodgestar:
### Summary

List the supported specification versions in the `openqasm3` package.

### Details and comments

In #508 and the following OpenQASM TSC meeting it was discussed that the `openqasm3` Python package version would be separate from that of the specification and that the package would support multiple specification versions. Thus it would be useful to list somewhere in the `openqasm3` package which versions of the specification are supported.

I considered adding a simple `SpecVersion` class, but I wasn't sure it was worth the (small) amount of complexity. A comparable specification version might be useful, but in the AST rather than this list of versions (e.g. so that users of the AST could check whether a version was `> SpecVersion(3, 1)`).

I'm happy to add a `SpecVersion` class and also use it in the AST, but I thought I'd start with this simpler implementation.

### Checklist

✅ I have added the tests to cover my changes.
✅ I have updated the documentation accordingly.
✅ I have read the CONTRIBUTING document.

### DIFF
--- a/releasenotes/notes/add-supported-spec-versions-to-openqasm3-pkg-7ea92eb1181bf0fd.yaml
+++ b/releasenotes/notes/add-supported-spec-versions-to-openqasm3-pkg-7ea92eb1181bf0fd.yaml
@@ -1,0 +1,7 @@
+---
+features:
+  - |
+    Added `openqasm3.spec.supported_versions` which lists the OpenQASM specification
+    versions supported by the `openqasm3` package. Currently the supported versions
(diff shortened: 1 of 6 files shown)


blakejohnson (approved):
I'm all in favor of simplicity until something more is needed. So, I am good with this approach.

blakejohnson:
It does look like you need to add a release note.

hodgestar:
Thanks @blakejohnson. Release note added.

PR #514, opened 2024-02-16
Release process updates

mbhealy:
Apologies for the delay, snow and family illness slowed things down.

### Summary
As discussed at the 02/07/2024 TSC meeting this PR makes a few minor updates to the contributing guidelines to spell out what was decided regarding the release process.

In general we agreed to:
- Release when the TSC feels that this is warranted (e.g. based on the number or nature of changes accumulated)
- Spec changes that require backwards incompatible changes at the parsing level warrants a major version bump. 
- Release v3.1 for the switch and other recent updates.

For the parser:
- The reference parser is an example implementation and not the true specification. It is used to vet proposed changes to the specification as a proof that they can be parsed and do not conflict egregiously with existing constructs.
- Regarding treatment of the OpenQASM version declaration in the parser:
  - It should generally be ignored by the reference parser for version-specific feature detection/support, but
  - Warnings for non-supported constructs in a specified version are appreciated but not required.
- The current list of versions supported by the parser will be listed within the parser itself. See https://github.com/openqasm/openqasm/pull/513
- The parser is ready for a 1.0 release - supporting v3.1 of specification

TODO in upcoming PRs:
- Release notes need to be updated to separate parser and specification release notes
- Parser release notes need to be updated and cleaned up
- Spec version 3.1 needs to be tagged

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,8 +77,8 @@ Please follow the next rules for the commit messages:
 ## Development Cycle
 
 The development cycle for OpenQASM is managed in the open using Github for project management.
-TODO: When preparing a new release changes for the released version should be identified
-in the release notes (See [issue #328](https://github.com/openqasm/openqasm/issues/328)).
+Release notes are collected separately for the specification and reference parser in
+the `releasenotes` directory.
 
 ### Semantic Versioning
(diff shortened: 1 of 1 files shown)


hodgestar (approved):
What's here looks good to me.

blakejohnson (approved):

PR #518, opened 2024-03-19
Add separate reno sections for specification and ast/grammar

mbhealy:
### Summary

This is the first alternative for releasing versions of the specification and ast/grammar separately, the other option is #519. 
Create separate reno sections for specification and ast/grammar updates. 

### Details and comments

I've added `spec_*` and `ast_*` sections to the reno notes, along with a template including those sections/subsections. I've also updated the existing notes with the correct sections.

PR #519, opened 2024-03-19
Move ast and specification release notes into separate directories

mbhealy:
### Summary

This is the second of two options for handling separate AST/grammar and specification release notes using reno. The first is #518 .

### Details and comments

In this version I have created new directories `spec_releasenotes` and `ast_releasenotes` and separated the existing notes between the two. I have also updated the contributing guidelines appropriately and the script that checks that a reno note has been created whenever there are changes in the source directory.


spec_releasenotes/releasenotes/config.yaml
```diff
@@ -0,0 +1,6 @@
+---
+encoding: utf8
+default_branch: main
+collapse_pre_releases: true
+branch_name_prefix: release/
```
hodgestar, line 5: Should the release branch prefixes be different for the spec and the Python package?

.github/renos_updated.sh
```diff
@@ -21,5 +23,5 @@ do
    fi
 done
 
-echo Please add or update a release note in ./releasenotes >&2
+echo Please add or update a release note in ./spec_releasenotes or ./ ast_releasenotes >&2
```
hodgestar, line 26:
```suggestion
echo Please add or update a release note in ./spec_releasenotes or ./ast_releasenotes >&2
```

mbhealy: Fixed in https://github.com/openqasm/openqasm/pull/519/commits/d6a45d0c94c6d86d61610542b7bb6e09349f9742

spec_releasenotes/releasenotes/config.yaml
mbhealy: Yes, I'm still working on how to get reno to correctly find the right tags/branches for each artifact. We may have to have separate `spec_release/*` and `ast_releast/*` branch names.

mbhealy: Ok, I think I've made reno find the right things. Reno keys on the tags to version each note, it ignores branches to determine which notes to apply. However, it uses branches to determine where to stop in the history for tags. So for the `stable/3.0` branch it will search for all tags that are newer than that branch and organize the included notes in the history back to `stable/3.0`.

So the question is, do we want to change our release process to use different stable branches for the AST and spec releases? I would advocate for yes, we should have something like`stable/spec/3.x` and `stable/ast/x.x' branches that we maintain, then we can update the `branch_name_prefix` and corresponding `_re` to match on them differently.

mbhealy:
I have an example version of a publishable build using this branch available at https://mbhealy.github.io/ currently. The release notes in the 3.0 and 3.1 versions don't contain anything currently because we would need to backport this PR there first.

spec_releasenotes/releasenotes/config.yaml
mbhealy: TSC discussion today resolved to maintain a single stable branch for both the AST and specification.

hodgestar (approved):

braised-babbage (approved):
Looks good to me. Not really an issue for this PR but I wondered if things might be clearer if we move the code for the Python package out from the current `source` directory (which contains the spec and `.rst` files). In a world where we had something like two disjoint directories

```
spec/
ast/
```

then e.g. we could have release notes for each stored in these. There is a common dependence on the ANTLR grammar, so that would have to be sorted out.

blakejohnson (approved):

blakejohnson:
Ok to merge, I assume?

PR #521, opened 2024-03-20
Add missing type annotation to loop variable in for loop examples

jlapeyre:
In several examples of `for` loops in the language specification there is no annotation, or type specification, on the loop variable. According to the spec omitting the annotation is not allowed.

This commit corrects each such instance by adding an annotation.


braised-babbage (approved):
👍

PR #522, opened 2024-03-21
Add release note for switch statement support.

blakejohnson:
I would be happy to write a more extensive note, but as I started to write more I felt like I was duplicating content from the spec itself. How expressive do we want these notes to be?

### DIFF
--- a/releasenotes/notes/switch-syntax-refactor-dd5253d79bc695aa.yaml
+++ b/releasenotes/notes/switch-syntax-refactor-dd5253d79bc695aa.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
+    Added support for switch statements, allowing for compact description of multi-way
+    branches that depend on a common control expression.


releasenotes/notes/switch-syntax-refactor-dd5253d79bc695aa.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Added support for switch statements, allowing for compact description of multi-way
+    branches that depend on a common control expression.
```
jlapeyre, line 5: This file should probably have a trailing new line.

jlapeyre (changes requested):
I agree that this is the appropriate amount of detail for this release note. If the concepts or structures added to the spec were more complicated, maybe more would be warranted. Not the case here.

releasenotes/notes/switch-syntax-refactor-dd5253d79bc695aa.yaml
blakejohnson: Fixed in beda3de

hodgestar (approved):

jlapeyre (approved):

PR #517, opened 2024-03-15
Document content of the standard library

jakelishman:
### Summary

This describes the effects of the `stdgates.inc` include file as part of the language specification.  I deliberately did not include the `examples/stdgates.inc` version of it in the specification to avoid implications that it *must* be implemented with that manner, but attempted to explain how implementations must behave *as if* if were defined like this, so that implementations are free to interpret the file in terms of `defcal` statements or the like (if they were to so choose).

This commit also includes a brief Sphinx extension, to make it possible to document OpenQASM gates in the full Sphinx structure.

### Details and comments

Close #516.

The text here is largely just a first draft - happy for any suggestions to the requirements, or different ways of presenting things.

It probably makes sense for me to split of the Sphinx extension into a separately installable Python package, but just for the purposes of demoing this PR, I've included it inline here.  If we want it, I'll sort out the legal stuff for my work on it on the IBM side (since here, it's just folded into the licensing for this repo) and move it into a separate repo.  The current implementation is bare bones and just enough to document what I needed to.

PR #523, opened 2024-03-22
Replace an occurrence of stdgates.qasm with stdgates.inc

jlapeyre:
In another PR We are adding behavior around stdgates.inc to the spec, and the name of the file is fixed. The appearance of `include "stdgates.qasm"` here is confusing because it looks like it is including the standard library declarations, but in fact is not.

* See #517

### DIFF
--- a/releasenotes/notes/remove-incorrect-stdgates-file-d6a93fe18dbe669e.yaml
+++ b/releasenotes/notes/remove-incorrect-stdgates-file-d6a93fe18dbe669e.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
+    Remove `include "stdgates.qasm";` from an example in the specification. This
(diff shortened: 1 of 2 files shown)


hodgestar (approved):

jlapeyre:
Any opinions on a release note here? I could do one of
* Add the no-release-note tag
* Write a release note

Hmmm. I'm inclined to write one. Before this change the spec might give someone the impression that `.qasm` was officially sanctioned. I actually was led astray by it and hardcoded it in the rust parser. Someone might misremember that the spec at one time sanctioned `.qasm` but now does not.

jlapeyre:
@hodgestar Thanks for approving earlier. I thought a release note might be in order, so I added one.

hodgestar (approved):
Release note is a good idea. Looks good.

PR #525, opened 2024-04-03
Update python package README.md: name of parser extra

garrison:
### Summary

I have updated the python package's README to refer to the extra `parser` rather than `parsing`, as the former is the name used in `setup.cfg`.

https://github.com/openqasm/openqasm/blob/main/source/openqasm/setup.cfg#L32


hodgestar (approved):
Thanks!

jlapeyre:
I think this is the kind of change that should *not* include a release note.

PR #529, opened 2024-04-17
Correct integer division in `pow` modifiers

### LINKED ISSUE #528: Integer math in stdgates.inc
### OpenQASM version:

3.1

The integer math in stdgates.inc for `sx` causes the power to round to 0 and implement the wrong gate. Raised originally in qiskit  https://github.com/Qiskit/qiskit/issues/12167

Inline the stgates.inc definition of `sx`

Define sx as sqrt(x)

Replace `1/2` with `0.5` or equivalent.

jakelishman:
### Summary

These would have been evaluted to the integer `0` instead of the intended floating-point value `0.5`.

### Details and comments

Fix #528

### DIFF
--- a/examples/stdgates.inc
+++ b/examples/stdgates.inc
@@ -1,4 +1,10 @@
-// OpenQASM 3 standard gate library
+// OpenQASM 3.0 standard gate library
+//
+// Note: the gates defined by this file are exactly the set that the OpenQASM
+// 3.0 specification defined, as are the unitary actions, but implementations
+// have some scope for how they handle the file internally.  See the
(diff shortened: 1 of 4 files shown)


hodgestar (approved):

jakelishman:
Me too - I actually wrote this PR before Lev made that issue, but after I had to run off very quickly at the end of the TSC meeting I didn't have time to push it.  Happy to switch to whichever.

hodgestar:
I'll put in a +0.5 for switching to `0.5` as slightly nicer looking.

hodgestar (approved):

PR #536, opened 2024-05-05
Fix incorrect syntax in broadcasting example.

dlyongemallo:
### Summary

The example given for broadcasting uses the incorrect syntax for declaring quantum registers. This is a minor edit to fix the typo.

### Details and comments

The correct syntax (see, e.g., [here](https://openqasm.com/language/types.html), or even just down below in this very same doc) is:

```python
qubit[SIZE] q;
```

The current documentation incorrectly puts the size after the register name:

```python
qubit q[SIZE];
```


jakelishman (dismissed):
Thanks!

PR #538, opened 2024-06-24
Release ast v1.0.0

braised-babbage:
This PR adds documentation around the release process, and updates the `openqasm3` package version to prepare for a release.

A few comments:
- I have added some notes about the release process from @jakelishman. We can tweak / update these after the release as need be.
- Regarding ANTLR versions, the latest on [pypi](https://pypi.org/project/antlr4-python3-runtime/) is 4.13.1, so we don't need to change anything here.
- Regarding the package version, at least a few months ago there seemed to be a general consensus from the members of the "profiles and releases working group" (@mbhealy, @braised-babbage, @hodgestar, and @jakelishman) that we could just move to a 1.0 release. I am happy to dial this back if this consensus does not hold. 
- We will also want to do a release of the OpenPulse package. I think that can lag this by a bit (and I wouldn't be surprised if there were still a few lingering tasks to resolve prior to cutting that).

### DIFF
--- a/ast_release.md
+++ b/ast_release.md
@@ -1,7 +1,5 @@
 # Deploying the OQ3 Python AST
 
-Author: Jake Lishman `jake.lishman@ibm.com`
-
 You need:
 - permission to push tags to `openqasm/openqasm`
 - a reviewer to approve a PR to the branch
@@ -11,7 +9,7 @@ Most of the deployment is automated by the GitHub workflows:
 - `/.github/workflows/deploy-ast.yml`
 which do what they say on the tin.
 
-The "deploy" workflow uses the organisation-level secret `OPENQASM_BOT_PYPI_TOKEN`, which is an API token for the PyPI user `openqasm-bot`, which I (Jake: `jake.lishman@ibm.com`) currently own.
+The "deploy" workflow uses the organisation-level secret `OPENQASM_BOT_PYPI_TOKEN`, which is an API token for the PyPI user `openqasm-bot` currently owned by Jake Lishman (`jake.lishman@ibm.com`).
 PyPI's authorisation policies may need us to change how we handle that in the future, if they bring in some of their proposed changes around API tokens and 2FA.
 
 ## Procedure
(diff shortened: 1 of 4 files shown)


ast_release.md
```diff
@@ -0,0 +1,48 @@
+# Deploying the OQ3 Python AST
+
+Author: Jake Lishman `jake.lishman@ibm.com`
```
jakelishman, line 3: Thanks for thinking of me, but no particular need to credit my quickly typed Slack message!  If you wanted to for history purposes, you can add me as a co-author to the commit that creates this file (I use `Co-authored-by: Jake Lishman <jake.lishman@ibm.com>` for work-related commits), but otherwise no worries.

If we write it in, it'll just stale as people edit things in the future.

ast_release.md
```diff
@@ -0,0 +1,48 @@
...
+- permission to push tags to `openqasm/openqasm`
+- a reviewer to approve a PR to the branch
+
+Most of the deployment is automated by the GitHub workflows:
+- `/.github/workflows/build-ast.yml`
+- `/.github/workflows/deploy-ast.yml`
+which do what they say on the tin.
+
+The "deploy" workflow uses the organisation-level secret `OPENQASM_BOT_PYPI_TOKEN`, which is an API token for the PyPI user `openqasm-bot`, which I (Jake: `jake.lishman@ibm.com`) currently own.
+PyPI's authorisation policies may need us to change how we handle that in the future, if they bring in some of their proposed changes around API tokens and 2FA.
```
jakelishman, line 15: PyPI's authorisation policies _did_ change since I wrote this, and now there's the concept of "trusted deployment environments".  We'd still be tied to `openqasm-bot` owning the package to some degree, but I can update the deployment to use the better system (no need to block this PR on that - I'll update it later).

ast_release.md
```diff
@@ -0,0 +1,48 @@
...
+
+## ANTLR considerations
+
+If we just generated one version of the ANTLR parser, the `openqasm3` Python package would only be compatible with the matching minor version of `antlr4-python3-runtime`.
+This causes problems when people want to use our package in conjunction with another ANTLR-generated parsing package, which may have used a different version of ANTLR.
+`openqasm3` has a dynamic import system for its ANTLR-generated components that chooses from the versions of those files the one that matches the installed version of `antrl4-python3-runtime`.
+The wheel-build action `build-ast.yml` uses the file `/source/openqasm/ANTLR_VERSIONS.txt` to decide which versions of the ANTLR compiler it should use to generate ANTLR files in the wheel.
+
+The Python ANTLR runtime is compatible with all releases of ANTLR that share the same `major.minor` version, and potentially differ in the patch number.
+For example, if version 4.11.1 of the ANTLR compiler (the Java bit) was used, the generated files will be compatible with versions 4.11.0, 4.11.1 and 4.11.1 (etc) of `antlr4-python3-runtime`, but versions 4.10.1 or 4.12.0 would not be.
```
jakelishman, line 41:
```suggestion
For example, if version 4.11.1 of the ANTLR compiler (the Java bit) was used, the generated files will be compatible with versions 4.11.0, 4.11.1 and 4.11.2 (etc) of `antlr4-python3-runtime`, but versions 4.10.1 or 4.12.0 would not be.
```

jakelishman (commented):
Thanks for doing this, Erik.  Seems like a good idea to do this to me.

hodgestar (approved):
Thanks @braised-babbage and @jakelishman. The documentation seems clear to me and the version bump looks good.

I learned about the ANTLR versioning considerations that I was not aware of before.

PR #543, opened 2024-07-10
Clarify that arrays and registers of size 0 are allowed

### LINKED ISSUE #532: Clarify whether arrays and bit registers of size 0 are allowed

I suggest that it should be stated explicitly whether bit registers or statically-sized arrays of size 0 are allowed by the spec.

Since it is not currently mentioned, I believe that registers with zero bits should be supported, but qe-qasm does not currently support them: https://github.com/openqasm/qe-qasm/issues/37.

For what it's worth, a zero-sized array on the stack [is disallowed in C](https://stackoverflow.com/questions/9722632/what-happens-if-i-define-a-0-size-array-in-c-c) by the standard.  However, I think there is a stronger case for allowing them in OpenQASM, for two reasons:

1. An unused empty array in a C can be eliminated from the code and replaced with `NULL` without the semantics of that code changing, AFAICT.  By contrast, removing an output array register of zero bits from an OpenQASM program would change the result; the register would no longer be included in the output.
2. OpenQASM code is often generated by a computer program, and the output is typically consumed by a computer program.  Running a program, with certain inputs, might sometimes result in a classical register of zero size, and it would be nice not to have to special case this both when the OpenQASM is generated and when the output is consumed.

If it is decided instead that arrays of size 0 should _not_ be allowed, then I suggest that it be clarified that every _dimension_ of the array must be at least 1 (e.g., `array[float[32], 3, 0, 2]` would contain zero elements and therefore also be disallowed).

garrison:
### Summary

Fixes #532.

### Details and comments

The primary change is to make it so a "size" can be a non-negative integer instead of a positive integer.  This works for qubits as well as classical bits/registers, but it leads to some clarifications being necessary for some other types:

- Integers of zero bits are explicitly forbidden by this PR.  An ~~unsigned~~ integer must have at least one bit.  ~~A signed integer must have at least two bits (the sign bit plus one integer bit).  The latter was probably implicitly assumed before, but now it is explicit.~~  NOTE: I saw nothing specifying that an integer must have a "normal" number of bits, i.e., 8, 16, 32, or 64, even though the examples tend to use these values.
- Floating point numbers must have a number of bits that corresponds to a valid IEEE 754 floating point type, so I believe it is unnecessary to explicitly say that zero bits are forbidden.
- Like unsigned integers, angles must have at least one bit.
- I added a statement permitting arrays to have dimension(s) which are of length zero.  I avoided the word "length" because it is currently not used to describe the individual dimensions given by the shape.  I am open to rephrasing this with guidance.
 
#### Remaining action items

- [x] add a release note
- [x] add tests for the parser
- [x] get tests to pass
- [x] revise release note: make it more specific

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -218,7 +218,7 @@ There are n-bit signed and unsigned integers. The statements ``int[size] name;``
 signed 1:n-1:0 and unsigned 0:n:0 integers of the given size. The sizes
 and the surrounding brackets can be omitted (*e.g.* ``int name;``) to use
 a precision that is specified by the particular target architecture.
-If provided, the ``size`` of an unsigned integer must be at least 1, and the ``size`` of a signed integer must be at least 2.
+If provided, the ``size`` of an integer must be at least 1.
 Bit-level operations cannot be used on types without a specified width, and
 unspecified-width types are different to *all* specified-width types for
 the purposes of casting.
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -785,7 +785,7 @@ and for multi-dimensional arrays subarray accesses can be specified using a
 comma-delimited list of indices (*e.g.* ``myArr[1, 2, 3]``), with the outer
 dimension specified first.
 
-One or more dimension(s) of an array can be zero, in which case the array has size zero.  An array of size zero cannot be indexed, e.g. given ``array[float[32], 0] myArray;``, it is an error to use either ``myArray[0]`` or ``myArray[-1]``.
+One or more dimension(s) of an array can be zero, in which case the array has size zero.  An array of size zero cannot be indexed, e.g. given ``array[float[32], 0] myArray;``, it is an error to access either ``myArray[0]`` or ``myArray[-1]``.
 
 For interoperability, the standard
 ways of declaring quantum registers and bit registers are equivalent to the
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -65,7 +65,7 @@ declares a reference to a quantum bit. These qubits are referred
 to as "virtual qubits" (in distinction to "physical qubits" on
 actual hardware; see below). The statement ``qubit[size] name;``
 declares a quantum register with ``size`` qubits.
-Sizes must always be :ref:`compile-time constant <const-expression>` positive
+Sizes must always be :ref:`compile-time constant <const-expression>` non-negative
 integers.
 Quantum registers are static arrays of qubits
 that cannot be dynamically resized.
@@ -218,6 +218,7 @@ There are n-bit signed and unsigned integers. The statements ``int[size] name;``
 signed 1:n-1:0 and unsigned 0:n:0 integers of the given size. The sizes
 and the surrounding brackets can be omitted (*e.g.* ``int name;``) to use
 a precision that is specified by the particular target architecture.
+If provided, the ``size`` of an unsigned integer must be at least 1, and the ``size`` of an unsigned integer must be at least 2.
 Bit-level operations cannot be used on types without a specified width, and
 unspecified-width types are different to *all* specified-width types for
 the purposes of casting.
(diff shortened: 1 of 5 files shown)


source/language/types.rst
```diff
@@ -218,6 +218,7 @@ There are n-bit signed and unsigned integers. The statements ``int[size] name;``
 signed 1:n-1:0 and unsigned 0:n:0 integers of the given size. The sizes
 and the surrounding brackets can be omitted (*e.g.* ``int name;``) to use
 a precision that is specified by the particular target architecture.
+If provided, the ``size`` of an unsigned integer must be at least 1, and the ``size`` of an unsigned integer must be at least 2.
```
jlapeyre, line 221:
```suggestion
If provided, the ``size`` of an unsigned integer must be at least 1, and the ``size`` of an signed integer must be at least 2.
```

jwoehr: **"a"** signed

jlapeyre:
> I saw nothing specifying that an integer must have a "normal" number of bits 8, 16, 32, or 64

Compare to Julia and Rust which are defined by an implementation rather than a spec (although Rust probably follows a more coherent design) . Julia `Int13` and Rust `i13` are not forbidden, but neither are the implemented and no one is interested in implementing them. The story in C is of course more complicated. But some motivations are shared with OQ3 since programs must be able to run on platforms with limited resources with varying widths of native types.

In OQ3, apparently a fully compliant implementation must handle 13 bit integers. I assume restricting the width of data types in OQ3 has been discussed in the past. Prima facie, it makes sense to restrict the widths somehow. For example require a minimum collection of data types, and allow implementations to support others.

jlapeyre:
> A signed integer must have at least two bits (the sign bit plus one integer bit). The latter was probably implicitly assumed before, but now it is explicit.

Otherwise, we would have a third representation of boolean values.

jakelishman:
Fwiw, a signed integer need not have two bits; `int[1]` can represent 0 and -1 in two's complement and +0 and -0 in one's complement. Not useful, but still well defined.

garrison:
> Fwiw, a signed integer need not have two bits; `int[1]` can represent 0 and -1 in two's complement and +0 and -0 in one's complement. Not useful, but still well defined.

I updated to allow for this in 19cd9fd3244f229962a37f75ae9e7f3287b19d97, as the spec is simpler this way.  We can always continue to discuss during the next TSC meeting.

levbishop:
This was discussed at TSC meeting and broadly approved. We'd like to have a test for the parser to confirm compliance with the given spec changes.

blakejohnson:
@garrison we are still hoping to see a test for the parser here.

garrison:
The one interesting thing I have found so far is that if a user attempts to provide an init expression for an empty array, it currently results in a `QASM3ParsingError`.  Of course, there is nothing to actually initialize in these cases, so there is no particular reason for the user to provide an init expression, but it still seems unexpected to me.  Some examples:

```openqasm
array[int, 0] a = {};
```
```openqasm
array[int, 2, 0] a = {{}, {}};
```
```openqasm
array[int, 0, 2] a = {};
```

On the other hand, the following program parses successfully, even though the dimension does not match:

```openqasm
array[int, 0] a = {0};
```

jakelishman:
I _think_ its unambiguous with the rest of the grammar - none of the other rules that involve `{}` are valid in the only locations that the array literal is (iirc).

garrison:
I modified the parser so that the tests that I added would pass in 8713166f1930a4.  I did this in the way most obvious to me.  I am very much a beginner with ANTLR, so this is likely shoddy work, but I wanted nonetheless to make it public for review while I try to figure out how to make it better.

blakejohnson (approved):

levbishop (approved):
Thanks for seeing this through. DIscussed at 2024-12-11 TSC meeeting

PR #552, opened 2024-09-18
Correct mathematical typos in `stdgates.inc` description

### LINKED ISSUE #545: Errors in the standard library documentation
### What is the file name(s) where you found the bug?

standard_library.rst

### What is the current text that needs to be fixed or removed? (You can copy and paste)

Definition of (at least) sdg (currently same as s) and rotation gates (currently 1/2 missing)

### What should the documentation say instead?

sdg: `i|1>` => `-i|1>`; rx/ry/rz/controlled: `theta` => `theta/2` (according to their definition in the paper in terms of `U`)

### Are there other related issues or pull requests about this bug?

_No response_

### Which files need to be updated?

_No response_

### What content needs to be updated or replaced? (You can copy and paste)

_No response_

### What new content needs to be added?

_No response_

jakelishman:
### Details and comments

Fix #545

### DIFF
--- a/source/language/standard_library.rst
+++ b/source/language/standard_library.rst
@@ -158,7 +158,7 @@ Single-qubit gates
 
       \texttt{sdg a;} \mapsto S^\dagger\colon\ \left\{\begin{alignedat}[c]2
          \lvert0\rangle &{}\to{}&  &\lvert0\rangle\\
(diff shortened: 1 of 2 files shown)


jlapeyre (approved):
Thank you !

PR #553, opened 2024-09-18
Declare `bit`/`bool` equivalence and differences between `bit[n]` and `array[T, n]`

### LINKED ISSUE #535: Clarify relationship between `bit`, `bit[1]` and `bool`

This stems from discussions we've had before, but was mostly triggered by an issue on Qiskit (Qiskit/qiskit#11870) that I forgot to bring over here between me and Blake, so I'll mostly just copy over the comments from that.

The original context is that Qiskit's legacy OQ3 importer rejected a program
```qasm
bit[4] c;
if (c[0] == "1") {
}
```
or the like, where `c[0]` (which I wrote the legacy importer to type as `bit`) is compared to `"0"` (which I typed `bit[1]`).  This is very related to #534.

jakelishman:
### Summary

This formalises the idea that `bit[n]` is a sequence-like type of which `bit[1]` is just the length-one variant.  This is distinct from the type `bit`, which is the scalar of it.  The `bit[n]` type is expected to be stored bit-packed (certainly when used in expression positions), which is reflected in how it casts to other integer types.

Taken alone, it is fairly meaningless to attempt to have a "bit-aligned" scalar type `bit`.  We don't attempt to define semantics for this in any sort ABI for OpenQASM 3; the only places this would be observable from within the program are in `array` and in `extern` calls, both of which it is forbidden from appearing in.  Instead, the semantics of scalar bit are simply set to have `bit` and `bool` be interchangeable in r-value usage.

An additional line about `qubit[n]` being equivalent to `array[qubit, n]` is deleted here, since some of the motivation for the rest of this commit is considering the alignment concerns of the `bit` and `bool` types; it doesn't really make sense to talk about classical alignment of a `qubit` type, and there doesn't appear to be any need to allow `array[qubit, n]` when `qubit[n]` already suffices.

### Details and comments

Close #535

### DIFF
--- a/source/language/types.rst
+++ b/source/language/types.rst
@@ -67,8 +67,7 @@ actual hardware; see below). The statement ``qubit[size] name;``
 declares a quantum register with ``size`` qubits.
 Sizes must always be :ref:`compile-time constant <const-expression>` positive
 integers.
-Quantum registers are static arrays of qubits
-that cannot be dynamically resized.
+Quantum registers cannot be resized after declaration.
 
 The label ``name[j]`` refers to a qubit of this register, where
 :math:`j\in \{0,1,\dots,\mathrm{size}(\mathrm{name})-1\}` is an integer.
@@ -191,7 +190,7 @@ Classical bits and registers
 There is a classical bit type that takes values 0 or 1. Classical
 registers are static arrays of bits. The classical registers model part
 of the controller state that is exposed within the OpenQASM program. The
-statement ``bit name;`` declares a classical bit, and or ``bit[size] name;`` declares a register of
+statement ``bit name;`` declares a single classical bit, while ``bit[size] name;`` declares a register of
 ``size`` bits. The label ``name[j]`` refers to a bit of this register, where :math:`j\in
 \{0,1,\dots,\mathrm{size}(\mathrm{name})-1\}` is an integer.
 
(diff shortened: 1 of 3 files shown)


source/language/types.rst
```diff
@@ -211,6 +210,26 @@ bit is on the right.
    // Declare and assign a register of bits with decimal value of 15
    bit[8] name = "00001111";
 
+
+The scalar type ``bit`` logically represents a single bit of data, but no restrictions or suggestions are made for how an implementation should choose to store a variable decalred as a scalar bit.
```
braised-babbage, line 214:
```suggestion
The scalar type ``bit`` logically represents a single bit of data, but no restrictions or suggestions are made for how an implementation should choose to store a variable declared as a scalar bit.
```

jlapeyre, line 214:
```suggestion
The scalar type ``bit`` logically represents a single bit of data, but no restrictions or suggestions are made for how an implementation should choose to store a variable declared as a scalar bit.
```

blakejohnson (dismissed):

braised-babbage (approved):
Ok, I read this and think it is clear enough. Thanks for doing this!

jakelishman:
I updated the title of the PR to reflect the two changes it makes, rather than one of them being hidden.

levbishop (approved):

braised-babbage (approved):

Issue #547, opened 2024-08-07
Consider allowing for `defcal` to be parameterized by classical types


braised-babbage:
### What is the expected behavior?

A `defcal` represents the implementation of a quantum operation (e.g. gate or measurement) in terms of a calibration grammar such as OpenPulse. It is typical that the body of a defcal depends on classical data, such as qubit frequency or waveforms. However, dependencies on non-angle types cannot currently be expressed via defcal parameters. In the parser this restriction is explicit, whereas it is perhaps implied by the [spec](https://openqasm.com/language/pulses.html#pulse-level-descriptions-of-gates-and-measurement). 

As one example of how users may wish to manipulate `defcal` parameters, imagine running some benchmarking protocol (e.g. randomized benchmarking) and varying the parameter of a waveform used for a gate. This program naturally mixes circuit-level notions with pulse-level notions.

Because we can't provide classical parameters directly to a gate implemented by a `defcal`, in practice one settles for a workaround. For the benchmarking example I gave above, the easiest thing would be for the relevant `defcal` to depend on a global variable. Alternatively, one might simply construct multiple versions of each `defcal` (one for each parameter value of interest). There's nothing intrinsically wrong with either of these approaches, but my point is just the current notion of `defcal` is less expressive than e.g. its counterpart as a subroutine. The world has settled on expressing subroutine dependencies through their argument list, because the result is easier to reason about. 

This is all complicated by the fact that defcals are intended to be coupled to gates, and in the context of a `gate` definition, non-`angle` parameters may make less sense. This came up in the discussion [here](https://github.com/openqasm/openqasm/pull/346). In order for defcal parameters to be useful, they must be able to appear in corresponding gate applications. So there are a few possibilities (although perhaps there are alternative framings of this):
1. neither `defcal` nor `gate` definitions allow non-`angle` parameters
2. we allow `defcal` and gate applications to have non-`angle` parameters, but `gate` definitions do not
3. we allow `defcal` and `gate` definitions to have non-`angle` parameters
Options (2) and (3) would require rethinking aspects of `gate`.

blakejohnson:
What about an option that kind of blends (2) and (3) where we allow non-`angle` parameters to gates but disallow their use in `gate` bodies? This would involve accepting the argument part of the extended gate definition syntax, but would not yet permit any expansion of `gate` bodies.

hodgestar:
[...]

PR #563, opened 2024-11-01
Add extended gate syntax

blakejohnson:
This is intended to address #547. What I have done is copy over the extended gate syntax from #346. My intention is to add further changes to the `defcal` syntax that references using non-`angle` parameters within `defcals`.

### DIFF
--- a/.github/workflows/tests-ast.yml
+++ b/.github/workflows/tests-ast.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         # Just using minimum and maximum to avoid exploding the matrix.
-        python-version: ['3.7', '3.12']
+        python-version: ['3.9', '3.12']
         antlr-version: ['4.7', '4.13']
     defaults:
       run:
--- a/.github/workflows/tests-ast.yml
+++ b/.github/workflows/tests-ast.yml
@@ -14,9 +14,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Just using minimum and maximum to avoid exploding the matrix.
-        python-version: ['3.9', '3.12']
-        antlr-version: ['4.7', '4.13']
+        include:
+          # Build on the oldest and latest supported Pythons
+          # and on the oldest and latest ANTLR version supported
+          # by each of those Pythons:
+          - python-version: '3.9'
+            antlr-version: '4.7'
+          - python-version: '3.9'
+            antlr-version: '4.13'
+          - python-version: '3.13'
+            antlr-version: '4.9'
+          - python-version: '3.13'
+            antlr-version: '4.13'
     defaults:
       run:
         working-directory: source/openqasm
--- a/ast_releasenotes/releasenotes/notes/extended-gate-syntax-15b2f58b731db7e4.yaml
+++ b/ast_releasenotes/releasenotes/notes/extended-gate-syntax-15b2f58b731db7e4.yaml
@@ -1,0 +1,8 @@
+---
+features:
+  - |
+    Introduces a new ast node type ``QuantumExtendedGateDefinition``` where the arguments
+    are ``List[ClassicalType]`` instead of ``List[Indentifier]``. A new node type was
+    introduced in order to avoid breaking backward compatibility with prior versions of
+    the reference parser. Consumers of the AST should now check for this new node type in
+    addition to the previously existing ``QuantumGateDefinition``.
--- a/source/grammar/qasm3Parser.g4
+++ b/source/grammar/qasm3Parser.g4
@@ -106,7 +106,7 @@ quantumDeclarationStatement: qubitType Identifier SEMICOLON;
 // Declarations and definitions of higher-order objects.
 defStatement: DEF Identifier LPAREN argumentDefinitionList? RPAREN returnSignature? scope;
 externStatement: EXTERN Identifier LPAREN externArgumentList? RPAREN returnSignature? SEMICOLON;
-gateStatement: GATE Identifier (LPAREN params=identifierList? RPAREN)? qubits=identifierList scope;
+gateStatement: GATE Identifier (LPAREN params=gateParameterList? RPAREN)? qubits=identifierList scope;
 
 // Non-declaration assignments and calculations.
 assignmentStatement: indexedIdentifier op=(EQUALS | CompoundAssignmentOperator) (expression | measureExpression) SEMICOLON;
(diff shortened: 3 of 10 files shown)


source/language/gates.rst
```diff
@@ -166,10 +170,57 @@ of the gate definition.
...
+     U(0, 0, alpha/k) a[1];
+     U(0, 0, alpha/(k**2)) a[2];
+   }
+
+
+For either kind of gate declaration, only built-in gate statements and calls to previously
+defined gates can appear in the body of the ``gate``.
 For example, it is not valid to
 declare a classical register in a gate body. Looping constructs over these quantum
 statements are valid.
```
blakejohnson, line 226: This text wasn't introduced in this PR, but I just realized that these two statements are in conflict with each other because the loop induction variable is effectively declared by a `for` loop statement.

blakejohnson:
Feedback today was:
* drop the colon between the classical parameters and quantum arguments
* drop registers as allowed quantum arguments (thus, no new broadcasting behavior)
* drop type specifiers for quantum arguments

I will do these things and then begin attempting the necessary changes in the parser.

source/grammar/qasm3Parser.g4
```diff
@@ -220,11 +220,17 @@ argumentDefinition:
     | (CREG | QREG) Identifier designator?
     | arrayReferenceType Identifier
 ;
+gateParameter:
```
blakejohnson, line 223: Name bikeshedding: to people prefer if this is called `gateArgument` instead?

blakejohnson:
GitHub doesn't offer the combination of Ubuntu 24.04 and python 3.7. So, we either need to bump the minimum python version to 3.8 or drop to Ubuntu 22.04.

jwoehr:
> GitHub doesn't offer the combination of Ubuntu 24.04 and python 3.7. So, we either need to bump the minimum python version to 3.8 or drop to Ubuntu 22.04.

We could drop 3.7 .AND. 3.8 
image

blakejohnson:
Oh, I see we already dealt with the same thing in terms of minimum python version in #570.

source/grammar/qasm3Parser.g4
braised-babbage: I don't have a strong preference. I think naively I've always thought of the parameters as those parts of the definition that determine the unitary operation, up to isomorphism, and then the arguments as specifying what qubit subspace this applies to.

With that said, there is a tiny bit of precedent for either, namely
1. In the Python AST, `QuantumGateDefinition` and `QuantumGate` use the word `argument` instead of `parameter`
2. In the spec, e.g. https://openqasm.com/language/gates.html#parameterized-gates, we tend to use the word _parameter_

source/language/gates.rst
```diff
@@ -180,9 +217,16 @@ and these symbols are scoped only to the subroutine body.
 
 An empty body corresponds to the identity gate.
 
+While the 'general' gate syntax allows for non-``angle`` parameters, OpenQASM current does
```
braised-babbage, line 220:
```suggestion
While the 'general' gate syntax allows for non-``angle`` parameters, OpenQASM currently does
```

ast_releasenotes/releasenotes/notes/extended-gate-syntax-15b2f58b731db7e4.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Introduces a new ast node type ``QuantumExtendedGateDefinition``` where the arguments
+    are ``List[ClassicalType]`` instead of ``List[Indentifier]``. A new node type was
+    introduced in order to avoid breaking backward compatibility with prior versions of
```
braised-babbage, line 6: I think even if we were OK with breaking backwards compatibility, there is still the question of how to distinguish the two kinds of gate definitions. The only alternative that comes to mind is to have some extra property on `QuantumGate` to flag that it is an extended definition (and this isn't any better than just adding a new definition, IMO).

braised-babbage (commented):
I dropped a few comments but this looks good to me.

source/grammar/qasm3Parser.g4
blakejohnson: Right, we seem to use "parameter" for this in the spec, but "argument" in the Python AST.

ast_releasenotes/releasenotes/notes/extended-gate-syntax-15b2f58b731db7e4.yaml
braised-babbage: To echo something I mentioned in the TSC meeting, as an implementor I prefer the `QuantumExtendedGateDefinition` since adding a new type like this gives a very legible "blast radius" to any necessary code changes I'd need to support (namely, additive code changes to some visitors).

.github/workflows/tests-ast.yml
```diff
@@ -14,18 +14,9 @@ jobs:
...
-              antlr-version: '4.7'
-            - python-version: '3.9'
-              antlr-version: '4.13'
-            - python-version: '3.13'
-              antlr-version: '4.9'
-            - python-version: '3.13'
-              antlr-version: '4.13'
+        # Just using minimum and maximum to avoid exploding the matrix.
+        python-version: ['3.9', '3.12']
+        antlr-version: ['4.7', '4.13']
```
blakejohnson, line 19: Reminder to myself to just grab the latest changes to this file from the `main` branch.

blakejohnson: Fixed in [3a94131](https://github.com/openqasm/openqasm/pull/563/commits/3a9413128657d260fcfb1a1c33110f13163584c0)

ast_releasenotes/releasenotes/notes/extended-gate-syntax-15b2f58b731db7e4.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Introduces a new ast node type ``QuantumExtendedGateDefinition``` where the arguments
```
hodgestar, line 4:
```suggestion
    Introduces a new AST node type ``QuantumExtendedGateDefinition``` where the arguments
```

source/grammar/qasm3Parser.g4
hodgestar: This grammar file also seems to use `argument`. E.g. `externArgumentList`. Should we change it?

source/language/gates.rst
hodgestar: Should we create a separate issue to clarify this?

hodgestar (commented):
I did a re-review. Looks good.

Bigger picture question: Would it be useful for the syntax to make it clear which parameters affect the unitary defined?

E.g.
```
gate rx(angle theta; bool drag) a { ... }
```
where the `;` indicates "arguments that affect the unitary end here".

source/grammar/qasm3Parser.g4
idavis, line 223: Perhaps `gateParameterList` should be a choice of `identifierList` or  list of typed identifier. This way the gate syntax can be validated at parse time rather than semantic analysis.

hodgestar:
> Bigger picture question: Would it be useful for the syntax to make it clear which parameters affect the unitary defined?
> 
> E.g. `gate rx(angle theta; bool drag) a { ... }`
> 
> where the `;` indicates "arguments that affect the unitary end here".

We discussed this suggestion at today's TSC meeting:

* @levbishop and I were keen to have a have way to make the conceptual distinction between parameters that affect the unitary and those that don't.
* The `;` separator seems awkward because its currently the statement delimiter

Some possible alternative ways of making the non-unitary parameters explicit (not all discussed at the meeting): 

* Use another separator character (perhaps `:` or `|`)
* Use a pseudo parameter like Python's `*` (e.g., `gate rx(angle theta, *, bool drag) a { ... }`)
* Use a keyword to modify the type (e.g. `gate rx(angle theta, nonunitary bool drag) a { ... }`)
* Use gate modifiers similar to the current `ctrl @` when passing the non-unitary parameters. The syntax for this gets messy, e.g., `defcalparams(True) @ rx(pi) a`.

It would perhaps be helpful to flesh out some of the examples we're considering more. Having a unitary that might use both some non-angle parameters and have some non-unitary parameters in the future might help clarify things.

blakejohnson:
Putting this on hold for now, waiting upon a strong use case.

PR #566, opened 2024-11-12
Recommend namespacing pragmas and annotations

hodgestar:
### Summary

This adds a recommendation to the directives specification suggesting that implementations namespace their pragmas and annotations to avoid collisions.

It reserves the use of the `openqasm` namespace for future use by the OpenQASM specification.

It extends the allowed annotations keywords to allow a dotted list of identifiers rather than a single identifier. At the cost of adjusting the parser to be a little more permissive, this allows pragmas and annotations to use identically namespacing that follows common conventions (e.g. Java and Python import namespacing). Using dot as the name separator allows underscore to be used within a name to separate words.

### Details and comments

Questions:

- Are we happy with extending the allowed annotation keywords as described? Does this require an OpenQASM specification version bump?
- Are we happy with the examples in the specification? I wasn't sure whether using `ibm.` was appropriate, but I also didn't want to make the examples too abstract.
- I would like to maintain a table of known namespaces, but the specification itself didn't feel like the correct place. Should I create a file somewhere else? If so, where? Or just add a small table to the specification for now?

### DIFF
--- a/source/language/directives.rst
+++ b/source/language/directives.rst
@@ -1,20 +1,50 @@
 Directives
 ==========
 
-OpenQASM supports directive mechanisms that allows other information to
+OpenQASM supports directive mechanisms that allow other information to
 be included in the program. Directives are either pragmas or annotations.
 Both are used to supply additional information to the compiler passes and the
 target system or simulator. Ideally the meaning of the program does not change
 if some or all of the directives are ignored, so they can be interpreted
 at the discretion of the consuming process.
 
+Pragma and annotation namespacing
+---------------------------------
+
+It is intended that different implementators of the specification might
+define their own pragmas and annotations. This creates the possibility that
+different implementations might define pragmas or annotations that collide.
+
+To avoid such collisions we recommend that implementations namespace their
+pragmas and annotations as follows:
+
+- Pragmas start with `pragma <namespace>.<name>(.<name>)*`.
+- Annotations start with `@<namespace>.<name>(.<name>)*`.
+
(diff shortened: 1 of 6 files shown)


source/language/directives.rst
```diff
@@ -40,10 +70,10 @@ the entire circuit.
    OPENQASM 3.0;
 
    // Attach billing information
-   pragma user alice account 12345678
+   pragma ibmq.user alice account 12345678
```
hodgestar, line 73:
```suggestion
   pragma ibm.user alice account 12345678
```

source/language/directives.rst
```diff
@@ -40,10 +70,10 @@ the entire circuit.
...
 
    // Assert that the QPU is healthy to run this circuit
-   pragma max_temp qpu 0.4
+   pragma ibmq.max_temp qpu 0.4
```
hodgestar, line 76:
```suggestion
   pragma ibm.max_temp qpu 0.4
```

braised-babbage:
I'm supportive of this idea, and taken on its own it seems pretty safe. My only request is that we consider how this comports with other potential notions of namespacing (e.g. as described in https://github.com/openqasm/openqasm/issues/365). For example, are there any caveats around the use of `.` as a separator more broadly (if we introduce namespaces as a notion outside of pragmas & annotations)? Is this even something we should be considering here?

blakejohnson (approved):
I feel like this is sufficiently contained that we can move forward without solving for namespacing in the broader language.

levbishop (approved):

PR #570, opened 2024-11-20
Update black to 24.10.

hodgestar:
### Summary

We had a dependabot alert to update black and its probably a little overdue in any case.

### Details and comments

The new black version made minor changes to:

- `source/openqasm/docs/conf.py`
- `source/openqasm/openqasm3/printer.py`

but no significant churn.

### DIFF
--- a/.github/workflows/tests-ast.yml
+++ b/.github/workflows/tests-ast.yml
@@ -23,7 +23,7 @@ jobs:
             - python-version: '3.9'
               antlr-version: '4.13'
             - python-version: '3.13'
-              antlr-version: '4.8'
+              antlr-version: '4.9'
             - python-version: '3.13'
               antlr-version: '4.13'
     defaults:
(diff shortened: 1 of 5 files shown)


hodgestar:
Hmm. This version of black is not available for Python 3.7. Let me see if there is a version that will satisfy both dependabot and still allow us to test against Python 3.7.

hodgestar:
The last version of black that supports Python 3.7 is 23.3.0 which is not a recent enough version to satisfy dependabot. Python 3.7 and 3.8 are now both end-of-lifed. Any objections to updating the antlr tests to run on Python 3.9?

jakelishman:
Python 3.7 and 3.8 are long EOL by now - I think it should be fairly uncontentious to drop support for them.

Fwiw, the 2025 Black style should be out relatively shortly, but it's not hard to make another PR. I suspect I was also lazy in setting up the CI matrix, if the style/lint checks run out of the same venv as the code tests - it shouldn't be an issue to separate them.

hodgestar:
Test passing. Ready for review.

levbishop (approved):
thanks

PR #576, opened 2025-01-06
feat: Add error listener for lexer on parse

ordinary-jamie:
### Summary

Add an `ErrorListener` that raises all errors to the lexer when parser permissive flag is unset.

### Details and comments

When parsing, callers have the option to set a `permissive` flag to let ANTLR attempt to recover from incorrect input. When this is unset, then the parser error handler is set to the bail error strategy. 

This PR adds a corresponding error listener for the lexer when it encounters an error, such as an unrecognised token.

An example use-case of this is added as a unit test; when the input has an invalid physical qubit reference, the current parser (without the error listener from this PR) will **incorrectly** produce an `ast.QuantumBarrier` with no qubits.

This PR provides callers the opportunity to catch this and rectify the program.

Note, that physical qubits must use an integer reference [1]:

> While virtual qubits can be named, hardware qubits are referenced by the syntax `$[NUM]`, where `NUM` is a non-negative integer.

#### References

[1] https://openqasm.com/language/types.html#physical-qubits
[2] https://www.antlr.org/api/Java/org/antlr/v4/runtime/ANTLRErrorListener.html


jakelishman (commented):
This looks very sensible and a clean addition, thanks.  Would you be able to add a release note with reno?

ordinary-jamie:
Thanks for the review, @jakelishman! Wasn't sure which section to file the release note under; decided with `features`, but let me know if you want this fixed.

jakelishman:
(sorry about the delay here) I think either features or fixes could be appropriate - however you prefer to think about it.

Looking at the failure (and trying to remember the process of this repo, sorry!), I think we have a non-standard location for `reno` notes, and the note needs to be in `ast_releasenotes` rather than `releasenotes`?  I'm not certain why we have the top-level `releasenotes` folder at all - maybe an accidental left-over.  Just a `git mv` of the release note should sort it out, I think.

Don't worry about the py3.7 test failure - I thought we'd removed 3.7 from the matrix, but I guess not.  The style failure is legit - might have been caused by a mismatched black version.

ordinary-jamie:
No worries @jakelishman. Pushed the fixes!

braised-babbage (approved):
LGTM -- thanks for adding this!

jakelishman (approved):
Sorry, I must have forgotten to come back to this.  Thanks for the changes!

Issue #577, opened 2025-01-08
[proposal]: Support `extern` identifiers


ordinary-jamie:
### Description

Currently, the `extern` keyword can be used in OpenQASM 3.0 to declare `extern` functions

https://github.com/openqasm/openqasm/blob/dfd1397fd4c8fd55b41b95d665ea41396138da42/source/grammar/qasm3Parser.g4#L108

Consider:

1. Extending this to also support `extern` declared variables/identifiers.
2. Defining a base `externTypes` in OpenQASM's grammar and requiring defcal grammars (e.g. OpenPulse) to override.

## Extensible defcal grammars

This will allow defcal grammars to define what types they wish to allow to be externally linked.
In the [example implementation](https://github.com/openqasm/openqasm/compare/main...ordinary-jamie:feat/add-extern-identifier?expand=1), each defcal grammar must override the `externType` rule.

This follows the pattern established by OpenPulse when introducing the `PORT`, `FRAME` and `WAVEFORM` scalar types:

https://github.com/openqasm/openpulse-python/blob/db7bf66e64be290b1d1485652d1c7e92ce07e499/source/grammar/openpulseParser.g4#L58-L60

## Issue in OpenPulse spec

The current `OpenPulse` spec erroneously specifies that it allows `extern port <IDENTIFIER>;` and `external frame <IDENTIFIER>;` statements

https://openqasm.com/language/openpulse.html#ports

> A port is only used to specify the physical resource on which to play a pulse or from which to capture data. The hardware can be accessed as OpenPulse port’s via extern identifier that specifies an external linkage that will be resolved at compile-time via vendor supplied translation units.
> 
> `extern port drive_port0`
> It is expected that a hardware vendor provide some documentation as to the associated functionality of a port (e.g. drive_port0 referse to the XY control line of a qubit 0).

However, attempting to parse a QASM program with this raises an error:

```py
import openpulse

qasm = """
OPENQASM 3.0;
defcalgrammar "openpulse";

cal {
    // No issue
    extern func(int) -> int;

    extern port drive_port0;
    extern frame xy_frame0;
}
"""

openpulse.parse(qasm)
```

Output:

```
Traceback (most recent call last):
  File "/python3.12/site-packages/openpulse/parser.py", line 88, in parse_openpulse
    tree = parser.calibrationBlock()
           ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #578, opened 2025-01-08
(Draft) Implement `extern` Identifiers

ordinary-jamie:
### Summary

Implements #577 

### Details and comments

Defcal grammar's allow the specification of externally linked resources, marked with the `extern` keyword. For example, OpenPulse allows declaring `extern port` or `extern frame` resources that are externally linked at compile time. However, the base grammar does not support such statements.

This commit adds a base parser rule for such externally linked resources.

### DIFF
--- a/source/grammar/qasm3Parser.g4
+++ b/source/grammar/qasm3Parser.g4
@@ -31,6 +31,7 @@ statement:
         | delayStatement
         | endStatement
         | expressionStatement
+        | externIdentifier
         | externStatement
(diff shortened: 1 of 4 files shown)


levbishop:
Discussed at TSC meeting of 2025-01-22. Decided won't merge for now, pending broader discussion about externs. (The fix to openpulse specifically is scheduled to fix with https://github.com/openqasm/openpulse-python/pull/37

Issue #584, opened 2025-02-11
[proposal]: built-in syntax for explicit no-op


jakelishman:
### Description

Have a spec-blessed way to signal an explicit no-op on a qubit, so that it's "used" by a `box` statement without ascribing any implementation-defined behaviour to it.

### Motivation

We currently have `box` in the spec, which can have an explicit duration (absolute, or stretchy).  The semantics of `box` imply synchronisation on all acted-on qubits in the box at the start and end of the box.  This can be used, for example, to describe dynamic decoupling on qubits, where the `box` is used to synchronise the starts and ends of the timing, using of stretch resolution.

In the fully-lowered, stretch-resolved dialects of OQ3, we don't need any explicit "virtual" no-op, because everything will have been replaced by `delay` instructions of absolute duration.  However, in the high-level virtual dialect, where qubits are given virtually and no stretch resolution has happened, there's no language-blessed way to mark "this qubit should idle for the length of the box" without introducing new stretches, or inserting a unitary gate statement that implements the identity operation (e.g. `delay[0dt] q[0];` or `id q[0];`).  It's less desirable to have these "semi-resolved" instructions, mostly because of the semantics - a `delay[0dt]` is only a no-op when applied to a single-qubit argument (otherwise it's a synchronisation point on two qubits), and `id` is defined in `stdgates.inc`, which isn't guaranteed to be included.

**This proposal is to have a spec-blessed way to mark a qubit as "behaves as if acted on in this scope", with no further semantics attached.**  This only practically has an effect within `box`, since no other scope implies synchronisation points on entry and exit.

Qiskit can represent this "`box` uses a qubit" form implicitly, but it can't represent this losslessly in OQ3 without choosing a magic representation itself

Possibilities:

1. bless `delay[0dt]` on a single qubit as meaning "explicit no-op", such that a parser could emit zero instructions in its own IR for such a node (e.g. in Qiskit, we _wouldn't_ emit an explicit `Delay(0dt)` instruction).
2. do nothing, and say that `id` or `U(0, 0, 0)` is sufficient to express this intent (with the same behaviour that a parser might choose to emit zero instructions in its own IR).
3. introduce a new keyword, like `noop` that accepts a qarg list (maybe like `barrier`), and can explicitly be ignored
[...]

PR #588, opened 2025-04-02
Add support for an explicit no-op instruction

jakelishman:
This requires compilers to validate the operands of the `nop` instruction, and has the effect of marking qubits as being "used" by a particular scope, but implies no further semantics on the qubits.

### Details and comments

See #584 for further details.


hodgestar (approved):

hodgestar:
We should also implement this in OpenPulse, with ports and frames as operands (not sure whether it should be only frame, or also port).

levbishop (approved):

nulinspiratie (approved):

levbishop:
Discussed at 2025-04-02 TSC meeting and approved to merge

PR #593, opened 2025-05-13
Update README.md Update call for secretary to November.

jwoehr:
Update call for secretary to November.


hodgestar (approved):
Thanks, @jwoehr.

Issue #590, opened 2025-04-09
[Bug]: `AliasStatement` type annotations


ShortDevelopment:
### OpenQASM version:

1.0.1

### Python version:

All

### Operating system:

All

### What is the current behavior?

The value of an `AliasStatement` may contain `IndexExpression`

https://github.com/openqasm/openqasm/blob/6c91bc6ab86c58a99186b312968c12bbbb821c71/source/openqasm/tests/test_qasm_parser.py#L1104-L1116

but the type annotations don't represent that

https://github.com/openqasm/openqasm/blob/6c91bc6ab86c58a99186b312968c12bbbb821c71/source/openqasm/openqasm3/ast.py#L1107

Note: According to the grammar all expressions seem valid

https://github.com/openqasm/openqasm/blob/6c91bc6ab86c58a99186b312968c12bbbb821c71/source/grammar/qasm3Parser.g4#L164

### Steps to reproduce the problem

Execute `mypy` on `test_qasm_parser.py`

### What is the expected behavior?

No type error

blakejohnson:
@hodgestar will take a first pass

PR #601, opened 2025-06-11
Update the type annotation of AliasStatement.value to match the grammar

hodgestar:
### Summary

Updates the type annotation of `AliasStatement.value` to `Expression` to match the grammar. Previous it was `Union[Identifier, Concatenation]`.

Addresses #590.

### Details and comments

There currently doesn't appear to be an intention to check these with a tool like mypy or pyright, but rather to address any issues as they're uncovered.


hodgestar:
Current status: On hold, pending an investigation by @nulinspiratie on what's needed to make a larger fix of issues reported by mypy.

hodgestar:
> Current status: On hold, pending an investigation by @nulinspiratie on what's needed to make a larger fix of issues reported by mypy.

Update: @nulinspiratie said he made some progress and fixed 50% of the issues, but 50% are more complex.

Issue #596, opened 2025-06-11
Support defcal call expressions


yitchen-tim:
### What is the expected enhancement?
For a program that includes a defcal call that has returns
```
OPENQASM 3;
bit[2] c;
defcal my_op1(bit b) $1 -> {};
defcal my_op2(bit b) $2 -> bit {};
rx(0.5) $1;
c[0] = measure $1;
my_op1(c[0]) $1;
c[1] = my_op2(c[0]) $2;
```
, the defcal call `my_op1(c[0]) $1;` is parsed as a `gateCallStatement`. However, `my_op2(c[0]) $2` is parsed as a base expression. This issue proposes to add support to defcal calls that has returns (e.g., the second defcal call in the above example). The use case includes specialized measurement instructions that is parameterized by pulse configurations.

### What does the solution looks like?
Parse defcal call `my_op2(c[0]) $2` as a `defcalCallExpression`. This include the following changes:
1. Define `defcalCallExpression`
```
defcalCallExpression:
    Identifier (LPAREN expressionList? RPAREN)? designator? gateOperandList;
```
This expression could use other naming, such as `gateCallExpression` since defcal is usually associated with a gate. However,  `defcalCallExpression` is more correct because we are treating operation that has returns, not gates. 

2. Update `assignmentStatement`
```
assignmentStatement: 
    indexedIdentifier op=(EQUALS | CompoundAssignmentOperator) (expression | measureExpression | defcalCallExpression) SEMICOLON;
```

3. Update `gateCallStatement` (optional): This part is optional, depending on whether we want to reuse `defcalCallExpression` in `gateCallStatement` or treat `defcalCallExpression` as a completely separate case. 
```
gateCallStatement:
    gateModifier* defcalCallExpression SEMICOLON
    | gateModifier* GPHASE (LPAREN expressionList? RPAREN)? designator? gateOperandList? SEMICOLON
;
```
### Related issues
https://github.com/openqasm/openqasm/issues/547 : In the discussion of #547, defcal is viewed as a general physical quantum operations (i.e., applying on physical qubits) and can include non-angle parameters. This request aligns with this view and proposes to support those physical quantum operations that has returns. 
https://github.com/openqasm/openqasm/issues/481: This request provides a more general solution than the one from #481. `measure_iq` or any "measurement with identifier" is a special case of defcal call with no parameter.
https://github.com/openqasm/openqasm/issues/256: Reset is a special case of defcal call with no parameter.

levbishop:
[...]

Issue #481, opened 2023-10-04
Grammar/AST does not support assignment with defined measurements


braised-babbage:
The OpenQASM spec provides for gate and measurement calibrations. In the case of measurements, these have defcals with return values. We don't require that these use the name `measure`. Among the examples in [this section](https://openqasm.com/language/pulses.html#pulse-level-descriptions-of-gates-and-measurement) we have

```
defcal measure_iq q -> complex[float[32]] { ... }
```

At present this is unusable, because [measure "expressions" ](https://github.com/openqasm/openqasm/blob/main/source/grammar/qasm3Parser.g4#L158)are hardcoded to use the `measure` token, and an [assignment statement](https://github.com/openqasm/openqasm/blob/main/source/grammar/qasm3Parser.g4#L106) only takes an expression or measure expression.

Given the various kinds of measurements that could occur on a given qubit, it's important that we support defcals with return values. This would require some changes to the grammar and AST.

PR #602, opened 2025-06-12
Allow generic measurement functions in measurement expressions

hodgestar:
### Summary

The specification and parser support defining measurement operations via `defcals`. For example the [defcal section](https://openqasm.com/language/pulses.html#pulse-level-descriptions-of-gates-and-measurement) has the following examples:

```
defcal measure $0 -> bit { ... }
defcal measure_iq q -> complex[float[32]] { ... }
```

However, the parser does not support *calling* measurement operations with names other than `measure`.

This MR adds the missing support via a new grammar expression and AST node. Adding a new node type ensures that current applications using the parser will continue to work for previously supported OpenQASM and fail with an unsupported node error when encountering the new syntax.

This MR was heavily inspired by #482.

See also: #481 (original issue) and #596 (second report of issue).

**Update:**

After @yitchen-tim [commented](https://github.com/openqasm/openqasm/issues/596#issuecomment-2967221436) on this PR in #596, I extended the PR to also support calling measurements defined with parameters:

```
defcal measure_rotated(angle[20] theta) q -> bit {  ... }
```

### Outstanding questions

- [x] Statements such as `measure_iq q;` and `measure_rotated(pi) q;` are parsed as calls to quantum gates. I'm not sure they can be easily distinguished. Is parsing them as gates acceptable? Decision: Yes, it is.
- [x] Measurements acting on multiple qubits remain uncallable -- i.e. `defcal measure_two_qubits q0, q1 -> bit[2] { ... }` . Should that also be fixed now? Yes, and implemented.

### Details and comments

- Quantum modifiers applied to measurements remain intentionally unsupported.

### DIFF
--- a/source/openqasm/openqasm3/parser.py
+++ b/source/openqasm/openqasm3/parser.py
@@ -637,7 +637,10 @@ class QASMNodeVisitor(qasm3ParserVisitor):
     @span
     def visitQuantumCallExpression(self, ctx: qasm3Parser.QuantumCallExpressionContext):
         if self._in_gate():
-            _raise_from_context(ctx, "cannot have a non-unitary 'measure' instruction in a gate")
+            _raise_from_context(
+                ctx,
+                "cannot have a non-unitary measure-like quantum call expression instruction in a gate",
+            )
         name = _visit_identifier(ctx.Identifier())
         arguments = (
             [self.visit(argument) for argument in ctx.expressionList().expression()]


idavis:
This PR looks like it is effectively adding arrow syntax to arbitrary "gate call" like structures.

From the PR desc "parser does not support calling measurement operations with names other than measure." - This is only true for the arrow syntax. As far as I can tell, defining a measurement with a `def` decl does everything this PR wants to achieve except the arrow syntax. From the spec:
```
 def pmeasure(angle[32] theta, qubit q) -> bit {
   rz(theta) q;
   h q;
   return measure q;
}
```

Using `gateOperand` (and gate syntaxt) also effective restricts the classical arguments to the measurement as `angle`s which is less functionality that the `def` measurement implementation. 

I don't really know `defcal`s so I might be missing something important. @jakelishman ?

yitchen-tim:
+1 to @jakelishman's idea of making it more like "quantum call expression". "Expression position" is more like a grammar rule describing where this expression can appear.

blakejohnson (approved):
Test coverage looks good.

PR #604, opened 2025-06-27
sizeof(arr) returns a `uint` when being passed an unspecified-length array reference

### LINKED ISSUE #603: `sizeof()` can't always return a `const uint`
### OpenQASM version:

3

### Python version:

N/A

### Operating system:

N/A

The issue is that `sizeof()` is a builtin that returns a `const uint`, but the length of arrays declared with the `#dim = const expression` syntax is not being known until runtime. These two facts are not compatible, because `sizeof` can't return a const value that is only known at runtime.

### More details

The first paragraph of the [Array in Subroutines](https://openqasm.com/language/subroutines.html#arrays-in-subroutines) section says:

"(...) The number of dimensions for all array parameters must be specified using the `#dim = const expression` syntax below, or specific lengths for each dimension must be provided. The unspecified-length version is provided because the lengths of the dimensions of array parameters (in the case of strided access) may not be known until runtime. (...)"

And then two paragraphs below it says:

"(...) `sizeof()` returns a const uint representing the length of the requested dimension of the array argument. (...)"

And there is an example using sizeof() with an array argument specified using the #dim = const expression` syntax.

N/A

The expected behavior is that `sizeof()` shouldn't return a `const uint` when taking as input an array which length isn't known at const time.

A possible solution is changing `sizeof` to return a `uint` for arrays that have non-const length. It could still return a `const uint` for arrays that have a const length (those declared with explicit size `array[int, 3, 4]`).

### Examples

```qasm
def runtime_arr_subroutine(readonly array[int, #dim = 2] arr) {
    // This works.
    uint dim_0_length = sizeof(arr, 0);

    // This doesn't work, because the length of the dimension is not const.
    const uint dim_1_length = sizeof(arr, 1);
}
```

```qasm
def const_arr_subroutine(readonly array[int, 2, 3] arr) {
    // This works.
    uint dim_0_length = sizeof(arr, 0);

    // This also works, because the length of the dimension is const.
    const uint dim_1_length = sizeof(arr, 1);
}
```

orpuente-MS:
Fixes #603 .

Currently the Spec says that `sizeof` returns a `const uint` when taking as input an array reference which length isn't known at const time. This PR adjust the Spec to say that `sizeof` returns `uint` when taking as input an array reference declared using the `#dim = const expression` syntax, which length is known at runtime.

### DIFF
--- a/source/language/subroutines.rst
+++ b/source/language/subroutines.rst
@@ -102,7 +102,9 @@ array parameters must be specified using the ``#dim = const expression``
 syntax below, or specific lengths for each dimension must be provided.
 The unspecified-length version is provided because the lengths of
 the dimensions of array parameters (in the case of strided access) may not be
-known until runtime. Passing multiple overlapping mutable references to the same
+known until runtime. The array references defined using the unspecified-length
+syntax have a different type from the ones defined using specific lengths for
+each dimension. Passing multiple overlapping mutable references to the same
 array to a subroutine is forbidden. However, the compiler will not always be
 able to resolve when this happens, and if it does, then no guarantees are made
 about the order that updates are made in.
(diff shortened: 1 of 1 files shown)


idavis:
Another note from the examples' arrays.qasm says that when dealing with #dim arrays the sizeof is explicitly not compile time (const) :
```openqasm
// When specifying array subroutine parameters, there is a second format where
// the sizes of the dimensions are not given explicitly, only the number of
// dimensions.  This is where the ``sizeof`` operator is most useful.  In these
// cases, `sizeof` is _not_ a compile-time constant.
def multi_dimensional_input(readonly array[int[32], #dim=3] my_array) {
    uint[32] dimension_0 = sizeof(my_array, 0);
    uint[32] dimension_1 = sizeof(my_array, 1);
    uint[32] dimension_2 = sizeof(my_array, 2);
}
```

braised-babbage:
This is sensible to me and I have no objections. Will defer to others if any strong opinions are voiced.

blakejohnson:
Consensus in the TSC meeting was to accept this PR, but we would like @jakelishman to take a look before finalizing.

blakejohnson (approved):

PR #607, opened 2025-07-03
Remove ~= assignment operator.

theinsanetramp:
### Summary

Remove the `~=` assignment operator from the grammar/parser.

### Details and comments

`~=` is never mentioned in the spec, never used in any of the examples, nor does anyone seem to know what it is supposed to do in the first place ([Slack thread](https://qiskit.slack.com/archives/C02FQB4F1M3/p1693332903688269)). Removing it resolves the confusion that was the impetus for https://github.com/openqasm/openqasm/issues/478, but does not resolve the issue of assignment operator documentation itself.


blakejohnson:
Thanks for the contribution!

PR #615, opened 2025-07-14
feat: Add error position info

ordinary-jamie:
### Summary

Update `openqasm3.parser.QASM3ParsingError` to include error position information

### Details and comments

* In #576 we added an error listener to raise the `QASM3ParserError`, however this PR forgot to propagate useful debugging information to the exception.
* This PR preserves line/column positional information in the exception object.
* Additionally, we delegates message formatting to the exception constructor; this will make it easier to maintain a standard message format (callers of the exception class don't have to worry about formatting).


blakejohnson (approved):
This looks like a nice enhancement. Thank you.

PR #624, opened 2025-09-03
[spec] Clarify const vs non-const durations

braised-babbage:
My notes from the 8/20 TSC meeting mention a few spec clarifications for `duration`, namely

- durationof returns a const duration
- durations can be const or non-const
- stretch expressions can be rejected by the type-checker because they involve non-const durations
- but some stretch failures are only caught later at stretch resolution time due to unsolvable constraints

These are related to https://github.com/openqasm/openqasm/issues/617, though with some tweaks based on the TSC discussion.

### DIFF
--- a/source/language/delays.rst
+++ b/source/language/delays.rst
@@ -148,6 +148,23 @@ backend-specific externs).
        // stretchy duration with backtracking by up to half b
        stretch e = -0.5 * b + c;
 
+Because `stretch` introduces a `const duration`, initialization with a non-`const` expression are a type error. On the other hand, because stretch resolution requires solving a system of constraints, a compiler may reject a well-typed program involving stretch durations due to a failure to resolve the constraint system.
+
+.. code-block::
+
+       input duration a; // not const
+       const duration b = 300ns;
+       stretch c;
+
+       // type error: const initializer expected
+       stretch d = c + a;
+
+       // stretch resolution error: 'e' resolved to -200ns
+       stretch e = b - 500ns;
+       delay[e] q0;
+
+
+
 Delays (and other duration-based instructions)
 ----------------------------------------------
 
--- a/source/language/delays.rst
+++ b/source/language/delays.rst
@@ -47,11 +47,11 @@ Below are some examples of values of type ``duration``.
 .. code-block::
 
        // fixed duration, in standard units
-       duration a = 300ns;
+       const duration a = 300ns;
        // fixed duration, backend dependent
-       duration b = 800dt;
+       const duration b = 800dt;
        // fixed duration, referencing the duration of a calibrated gate
-       duration c = durationof({x $3;});
+       const duration c = durationof({x $3;});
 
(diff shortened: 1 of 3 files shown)


source/language/delays.rst
```diff
@@ -127,23 +128,43 @@ these side effects. Also contrary to TeX, we prohibit overlapping gates.
...
-       duration b = durationof({x $0;});
+       const duration a = 300ns;
+       const duration b = durationof({x $0;});
        stretch c;
        // stretchy duration with min=300ns
        stretch d = a + 2 * c;
        // stretchy duration with backtracking by up to half b
        stretch e = -0.5 * b + c;
 
+Because `stretch` introduces a `const duration`, initialization with a non-`const` expression are a type error. On the other hand, because stretch resolution requires solving a system of constraints, a compiler may reject a well-typed program involving stretch durations due to a failure to resolve the constraint system.
```
jakelishman: Imo we shouldn't refer to "initialisation" of a `stretch`, because the `stretch a = <expr>;` statement doesn't _set_ `a` to the value of `<expr>`, it just introduces another lower-bound on the value of `a`.

I'd prefer
> Because `stretch` introduces a `const duration`, the optional lower bound express must also have type `const duration`.

(or something along those lines).

source/language/delays.rst
```diff
@@ -127,23 +128,43 @@ these side effects. Also contrary to TeX, we prohibit overlapping gates.
...
+
+       input duration a; // not const
+       const duration b = 300ns;
+       stretch c;
+
+       // type error: const initializer expected
+       stretch d = c + a;
+
+       // stretch resolution error: 'e' resolved to -200ns
+       stretch e = b - 500ns;
```
jakelishman: Imo this is not an error - it's just an ineffective bound because it's strictly weaker than the implicit `e > 0` bound that exists for all stretches.

jakelishman (commented):
Thanks Erik.  The actual bulk of the PR I'm fine with, but I think the new discussion around `stretch` is perhaps confusing the situation, since it's treating the "initialisation expression" of a stretch as an actual set, and not the lower-bound that it is (e.g. see the code block an comment here: https://openqasm.com/language/delays.html#operations-on-durations).

hodgestar (commented):
Looks good to me too, modulo Jake's comments and my one small suggestion to the change note.

blakejohnson:
I will take another look at this.

blakejohnson:
@mbhealy would be great if you could look at this, too.

source/language/delays.rst
mbhealy: The current spec says that passing negative values to `gate` or `box` definitions produces an error, in this case it's only used as a delay, so I don't think this necessarily would produce an error. To answer your question directly though I think that a conforming implementation *could* produce an error for this.

mbhealy (commented):
LGTM, agree with Jake's comments.

mbhealy (approved):
LGTM

PR #665, opened 2026-03-09
Docs: Fixes explicit state mapping for rz gate

OliverAh:
### Summary
Corrects the explicit state mapping of the rz gate given in the documentation. 
Fixes #660 

### Details and comments
The rz gate is defined as `rz(θ)=exp(-iZθ/2)`, where `Z` is the Pauli Z matrix. Therefore, mapping state `|0>` should result in a zero amplitude for state `|1>` and vice versa.

image


levbishop (approved):
Thanks for this.

PR #666, opened 2026-03-18
[Spec] Add explicit and implicit cast tables

### LINKED ISSUE #613: [proposal]: Casting table should be explicit regarding bit, bit[n], and bool
### Description

There are some ambiguities in the spec with respect to bool, bit, and bit[n]. I think in the casting table bit actually means bit[n]. And the casting rules for bit are just the casting rules for bool, or at least that should be the case according to the part of the spec where they say.

### Motivation

_No response_

### Authors

_No response_

### Examples

_No response_

orpuente-MS:
### Summary

Fixes #613 

This PR improves the type casting specification. It substitutes the **Allowed casts** table under **Types and Casting** -> **Casting Specifics** by two tables: **Explicit cast table** and **Implicit cast table**. These tables bring three improvements to the sepc:

1. Separate tables for explicit and implicit casts
2. Casting specifics for sized types (e.g. bit[n]).
3. The tables are complete, they include all types included in the OpenQASM specification.

TODO:
- [ ] I couldn't find any info on casting specifics between sized and unsized angles: `angle <-> angle[n]`

### Details and comments

I initially built these tables in the process of implementing the casting logic for the semantic pass of the OpenQASM compiler in the QDK. I had to gather information from different sections and examples of the OpenQASM spec and the C99 spec. Here are a few things that I found challenging when reading the **Allowed casts** table which led me to build these new tables:

1. The **Allowed casts** table does not differentiate between explicit and implicit casts.
2. It collapses all variants of a type (sized and unsized) into a single type: e.g. `bit`, `bit[n]`, and `bit[m]` with `n < m` are all represented as `bit`.
3. I had to reach for the C99 spec.
4. The `complex` type is missing, so I had to look for examples in other sections of the spec to understand its casting rules.

I figured other people could also face these challenges, so I am sharing these casting tables here as a PR.

Below are the same tables I included in the file changes, so that you can read them more easily while reviewing the PR.

### Explicit Cast Table

In this table `n` and `m` are positive integers s.t. `n < m`.

|            | bool | int | int[n] | int[m] | uint | uint[n] | uint[m] | float | float[n] | float[m] | angle | angle[n]               | angle[m] | bit | bit[n] | bit[m] | complex | complex[n] | complex[m] | duration |
| ---------- | ---- | --- | ------ | ------ | ---- | ------- | ------- | ----- | -------- | -------- | ----- | ---------------------- | -------- | --- | ------ | ------ | ------- | ---------- | ---------- | -------- |
| bool       | -    | Yes | Yes    | Yes    | Yes  | Yes     | Yes     | Yes   | Yes      | Yes      | No    | No                     | No       | Yes | Yes    | Yes    | No      | No         | No         | No       |
| int        | Yes  | -   | Yes    | Yes    | Yes  | Yes     | Yes     | Yes   | Yes      | Yes      | No    | No                     | No       | Yes | No     | No     | No      | No         | No         | No       |
| int[n]     | Yes  | Yes | -      | Yes    | Yes  | Yes     | Yes     | Yes   | Yes      | Yes      | No    | No                     | No       | Yes | Yes    | No     | No      | No         | No         | No       |
| int[m]     | Yes  | Yes | Yes    | -      | Yes  | Yes     | Yes     | Yes   | Yes      | Yes      | No    | No                     | No       | Yes | No     | Yes    | No      | No         | No         | No       |
| uint       | Yes  | Yes | Yes    | Yes    | -    | Yes     | Yes     | Yes   | Yes      | Yes      | No    | No                     | No       | Yes | No     | No     | No      | No         | No         | No       |
| uint[n]    | Yes  | Yes | Yes    | Yes    | Yes  | -       | Yes     | Yes   | Yes      | Yes      | No    | No                     | No       | Yes | Yes    | No     | No      | No         | No         | No       |
| uint[m]    | Yes  | Yes | Yes    | Yes    | Yes  | Yes     | -       | Yes   | Yes      | Yes      | No    | No                     | No       | Yes | No     | Yes    | No      | No         | No         | No       |
| float      | Yes  | Yes | Yes    | Yes    | Yes  | Yes     | Yes     | -     | Yes      | Yes      | Yes   | Yes                    | Yes      | Yes | No     | No     | No      | No         | No         | No       |
| float[n]   | Yes  | Yes | Yes    | Yes    | Yes  | Yes     | Yes     | Yes   | -        | Yes      | Yes   | Yes                    | Yes      | Yes | No     | No     | No      | No         | No         | No       |
| float[m]   | Yes  | Yes | Yes    | Yes    | Yes  | Yes     | Yes     | Yes   | Yes      | -        | Yes   | Yes                    | Yes      | Yes | No     | No     | No      | No         | No         | No       |
| angle      | Yes  | No  | No     | No     | No   | No      | No      | No    | No       | No       | -     | ??                     | ??       | Yes | No     | No     | No      | No         | No         | No       |
| angle[n]   | Yes  | No  | No     | No     | No   | No      | No      | No    | No       | No       | ??    | -                      | padding  | Yes | Yes    | No     | No      | No         | No         | No       |
| angle[m]   | Yes  | No  | No     | No     | No   | No      | No      | No    | No       | No       | ??    | truncation or rounding | -        | Yes | No     | Yes    | No      | No         | No         | No       |
| bit        | Yes  | Yes | Yes    | Yes    | Yes  | Yes     | Yes     | Yes   | Yes      | Yes      | No    | No                     | No       | -   | n = 1  | m = 1  | No      | No         | No         | No       |
| bit[n]     | Yes  | No  | Yes    | No     | No   | Yes     | No      | No    | No       | No       | No    | Yes                    | No       | Yes | -      | No     | No      | No         | No         | No       |
| bit[m]     | Yes  | No  | No     | Yes    | No   | No      | Yes     | No    | No       | No       | No    | No                     | padding  | Yes | No     | -      | No      | No         | No         | No       |
| complex    | No   | No  | No     | No     | No   | No      | No      | No    | No       | No       | No    | No                     | No       | No  | No     | No     | -       | Yes        | Yes        | No       |
| complex[n] | No   | No  | No     | No     | No   | No      | No      | No    | No       | No       | No    | No                     | No       | No  | No     | No     | Yes     | -          | Yes        | No       |
| complex[m] | No   | No  | No     | No     | No   | No      | No      | No    | No       | No       | No    | No                     | No       | No  | No     | No     | Yes     | Yes        | -          | No       |
| duration   | No   | No  | No     | No     | No   | No      | No      | No    | No       | No       | No    | No                     | No       | No  | No     | No     | No      | No         | No         | -        |

### Implicit Cast Table

NOTE: Only the behavior that is different from the Explicit Cast Table is filled in.

In this table `n` and `m` are positive integers s.t. `n < m`.

|            | bool | int | int[n] | int[m] | uint | uint[n] | uint[m] | float | float[n] | float[m] | angle | angle[n] | angle[m] | bit | bit[n] | bit[m] | complex | complex[n] | complex[m] | duration |
| ---------- | ---- | --- | ------ | ------ | ---- | ------- | ------- | ----- | -------- | -------- | ----- | -------- | -------- | --- | ------ | ------ | ------- | ---------- | ---------- | -------- |
| bool       | -    |     |        |        |      |         |         |       |          |          |       |          |          |     | No     | No     |         |            |            |          |
| int        |      | -   |        |        |      |         |         |       |          |          |       |          |          |     |        |        |         |            |            |          |
| int[n]     |      |     | -      |        |      |         |         |       |          |          |       |          |          |     | No     |        |         |            |            |          |
| int[m]     |      |     |        | -      |      |         |         |       |          |          |       |          |          |     |        | No     |         |            |            |          |
| uint       |      |     |        |        | -    |         |         |       |          |          |       |          |          |     |        |        |         |            |            |          |
| uint[n]    |      |     |        |        |      | -       |         |       |          |          |       |          |          |     | No     |        |         |            |            |          |
| uint[m]    |      |     |        |        |      |         | -       |       |          |          |       |          |          |     |        | No     |         |            |            |          |
| float      |      |     |        |        |      |         |         | -     |          |          |       |          |          |     |        |        | Yes     | Yes        | Yes        |          |
| float[n]   |      |     |        |        |      |         |         |       | -        |          |       |          |          |     |        |        | Yes     | Yes        | Yes        |          |
| float[m]   |      |     |        |        |      |         |         |       |          | -        |       |          |          |     |        |        | Yes     | Yes        | Yes        |          |
| angle      | No   |     |        |        |      |         |         |       |          |          | -     | No       | No       |     |        |        |         |            |            |          |
| angle[n]   | No   |     |        |        |      |         |         |       |          |          | No    | -        | No       |     |        |        |         |            |            |          |
| angle[m]   | No   |     |        |        |      |         |         |       |          |          | No    | No       | -        |     |        |        |         |            |            |          |
| bit        |      |     |        |        |      |         |         |       |          |          |       |          |          | -   | No     | No     |         |            |            |          |
| bit[n]     |      |     | No     |        |      |         |         |       |          |          |       | No       |          |     | -      |        |         |            |            |          |
| bit[m]     |      |     |        | No     |      |         |         |       |          |          |       |          | No       |     |        | -      |         |            |            |          |
| complex    |      |     |        |        |      |         |         |       |          |          |       |          |          |     |        |        | -       |            |            |          |
| complex[n] |      |     |        |        |      |         |         |       |          |          |       |          |          |     |        |        |         | -          |            |          |
| complex[m] |      |     |        |        |      |         |         |       |          |          |       |          |          |     |        |        |         |            | -          |          |
| duration   |      |     |        |        |      |         |         |       |          |          |       |          |          |     |        |        |         |            |            | -        |


mbhealy (dismissed):
This LGTM. Apologies for the delay in reviewing.

orpuente-MS:
@mbhealy Thanks for looking at it!

There are still four entries in the first table that need clarification, since I couldn't find any info on them in the specification. They are marked with `??` in the table, and correspond to these explicit casts:

- `angle` to `angle[n]` and `angle[m]`
- `angle[n]` and `angle[m]` to `angle`

We need to decide if we want to:
1. Allow these casts.
2. Disallow them.
3. Leave it as undefined behavior.

I would rather go with option 2, disallow them, to maintain consistency with the implicit casting table and with casts between `angle` and `bit[n]` being disallowed.

If that seems reasonable, I can make that change and then merge the PR.

orpuente-MS:
@mbhealy I went ahead and pushed the change I suggested. But it needs approval again. Could you take another look?

I will be on vacation until the end of month, but feel free to merge if the PR looks good to you.

nulinspiratie:
Hey @orpuente-MS, I think it's a great idea to have clarity on what casts are allowed and not, it's something that's been lacking.
The one thing I was wondering was whether we could condense the two 20x20 tables while still being explicit, which might help clarify the rules. After a few iterations (admittedly with Claude, though I checked the validity), the set we settled on consists of:
- Four rules
- One 8x8 table
- Two addendums to the table (we could consider migrating this to additional rules, but they felt more like addendums)
- Basic rules to cover implicit casts, which is a subset of the explicit cast table

I've added the details below. I think it matches your proposed set, with the added notes:

- `angle <-> angle[k]` still needs to be decided
- The explicit table stated that `float -> complex` is not allowed, whereas the implicit cast table stated this was allowed. If we adhere to the rule that the implicit casts are a subset of the explicit casts, I believe this should be allowed.

Keen to hear your thoughts, as well as the rest of the OpenQASM committee. Is this an improvement, or is it better to remain with the original 20x20 cast tables?


## The four rules

**R1 (Scalar bit equivalence).**
Scalar `bit` is interchangeable with `bool` for value casts, so wherever a rule
or table cell mentions `bool`, scalar `bit` behaves identically. Scalar `bit` is
*not* the register type `bit[1]`: the two are distinct types (`types.rst` lines
227 to 229). Scalar `bit` casts like `bool`, whereas `bit[1]` casts by width
match (R3).

```qasm
bit b = 1;
bool flag = bool(b);          // allowed: scalar bit and bool are interchangeable
int[32] n = int[32](b);       // allowed: b acts as bool, and bool -> int is allowed
bit[2] pair = bit[2](b);      // allowed: bool -> bit[n] of any width (note [a])
angle[1] ang = angle[1](b);   // not allowed: scalar bit is bool, not bit[1]; bool -> angle is No
```

**R2 (Same-family resize).**
Within `int`, `uint`, `float`, and `complex`, all size variants (sized or
unsized) cast freely in both directions.

```qasm
float[32] f = 3.14;
float[64] g = float[64](f);                      // allowed: widen within float
float[16] h = float[16](g);                      // allowed: narrow within float
complex[float[32]] c = complex[float[32]](2.0);
float[32] back = float[32](c);                   // not allowed: leaving the family is not a resize
```

**R3 (Width-matched bit bridge).**
`bit[k]` casts to and from `int[k]`, `uint[k]`, `angle[k]`, and other `bit[k]`
only when the widths are equal, using an exact little-endian bit copy. Width
mismatches require explicit slicing.

```qasm
bit[4] bits = "1010";
int[4]   i = int[4](bits);    // allowed: equal width, little-endian bit copy
angle[4] a = angle[4](bits);  // allowed: equal width
bit[4]   c = bit[4](bits);    // allowed: bit[k] <-> bit[k] at equal width
int[8]   j = int[8](bits);    // not allowed: 4 != 8, slice explicitly first
```

**R4 (Angle casts).**
- `float[*]` casts to `angle[*]` one way only, for any source and target size.
- `angle[n]` to `angle[m]` widening (`m > n`) zero-pads the least significant
  bits; `angle[m]` to `angle[n]` narrowing rounds or truncates, controlled by a
  `#pragma`.
- Unsized `angle` does not cast to or from sized `angle[k]`. (This is the open
  question flagged in the PR.)

```qasm
angle[20] a = angle[20](3.14);    // allowed: float -> angle (one way only)
float[64] f = float[64](a);       // not allowed: angle -> float
angle[32] wide = angle[32](a);    // allowed: widen, zero-pads least significant bits
angle[10] narrow = angle[10](a);  // allowed: narrow, rounds or truncates (#pragma)
angle un;
angle[16] s = angle[16](un);      // not allowed: unsized angle does not cast to/from angle[k]
```

---

## Explicit cast table

Legend:

- `Y` allowed
- `N` never allowed
- `R3` / `R4` governed entirely by that rule (defined above)
- `[a]` / `[b]` behavior not captured by a rule; see the lettered note below

| From \ To    | bool/bit | bit[n] | int/uint | float | angle | complex | duration |
|--------------|:--------:|:------:|:--------:|:-----:|:-----:|:-------:|:--------:|
| **bool/bit** | Y        | Y [a]  | Y        | Y     | N     | N       | N        |
| **bit[n]**   | Y        | R3     | R3       | N     | R3    | N       | N        |
| **int/uint** | Y        | R3     | Y        | Y     | N     | N       | N        |
| **float**    | Y        | N      | Y        | Y     | Y     | Y [b]   | N        |
| **angle**    | Y        | R3     | N        | N     | R4    | N       | N        |
| **complex**  | N        | N      | N        | N     | N     | Y       | N        |
| **duration** | N        | N      | N        | N     | N     | N       | Y        |

Scalar `bit` rides in the `bool/bit` row and column (R1), so the standalone
`bit[n]` row and column are sized registers (`n >= 1`), governed uniformly by
R3. Same-family resize (R2) makes every `Y` between `int`/`uint`, `float`, or
`complex` and their own size variants hold for all widths, so those cells need
no note. `int` and `uint` share one column because their behavior is identical.

Notes:

- **[a]** `bool` (and scalar `bit`) casts to `bit[n]` of any width: the low bit
  takes the value, the rest are zero. This is the one behavior that is neither a
  rule nor derivable from another cell.

  ```qasm
  bool flag = true;
  bit[4] bits = bit[4](flag);   // allowed: low bit takes the value -> "0001", rest zero
  ```
- **[b]** `float -> complex` is the one implicit promotion into `complex` (the
  imaginary part is set to zero). Since implicit casts are a subset of the
  explicit ones, it is also an explicit cast.

  ```qasm
  float[64] f = 2.0;
  complex[float[64]] c = f;                       // allowed: implicit promotion, imaginary part = 0
  complex[float[64]] c2 = complex[float[64]](f);  // allowed: also an explicit cast
  float[64] back = float[64](c);                  // not allowed: complex -> float
  ```
---

## Implicit casts

Implicit casts are a strict subset of the explicit ones:

- The same-family resize and the `bool` / scalar `bit` / `int` / `uint` /
  `float` numeric group cast implicitly (this is the all-`Y` block above).
- `float[*]` promotes implicitly to `complex[*]`, with the imaginary part set
  to zero.
- `angle[n]` and `angle[m]` resize implicitly, treated like unsigned integers
  (per the existing prose at `types.rst` lines 1049 to 1051).

Everything else requires an explicit cast. In particular, the width-matched bit
bridges (R3), `angle -> bool`, and `bool -> bit[n]` are explicit only.

orpuente-MS:
@nulinspiratie this is great. Thank you for putting time into this!

I have a few comments:

1. I agree that explicit casting from `float[*]` to `complex[*]` should be allowed.
2. I realized something when reading the `R3` and `R4` entries in your condensed explicit table: it is not enough to say `Y` to specify a cast is allowed, since that doesn't tell the reader the semantics of the allowed cast. So, even though my 20x20 table is "fully explicit", it doesn't talk about the semantics at all. So, what if we try to replace the remaining `Y` entries in your condensed tables by `R1` and `R2` where applicable?
3. I think it is better to have a second table for implicit casts, filling in the differences from explicit casts, instead of addendums to the explicit table. Especially now that you were able to condense it to an 8x8 table. The reason is that I believe a table format leaves no room for misinterpretation from the reader, and surfaces nice properties in the rule set, like symmetries in the table, which are nice to grasp visually if we try to make changes to the tables in the future.
4. I am wondering if it is worth including `duration` in the table at all, since casts between duration and other types are not allowed. There is a single entry from `duration` to `duration` that says `Y`, but this is not a cast, since it is from a type to itself; which is different from `float` to `float` which is really talking about `float` and `float[*]`.

With those comments in mind, what do you think about the following modification to your condensed table + addendums?

## Explicit cast table

| From \ To    | bool/bit | bit[n] | int/uint | float | angle | complex |
|--------------|:--------:|:------:|:--------:|:-----:|:-----:|:-------:|
| **bool/bit** | R1       | R6     | R7       | R7    | N     | N       |
| **bit[n]**   | R5       | R3     | R3       | N     | R3    | N       |
| **int/uint** | R5       | R3     | R2       | R2    | N     | N       |
| **float**    | R5       | N      | R2       | R2    | R4    | R2      |
| **angle**    | R5       | R3     | N        | N     | R4    | N       |
| **complex**  | R5       | N      | N        | N     | N     | R2      |

## Implicit cast table

**Note**: this table only fills in the behavior that is different from the explicit cast table.

| From \ To    | bool/bit | bit[n] | int/uint | float | angle | complex |
|--------------|:--------:|:------:|:--------:|:-----:|:-----:|:-------:|
| **bool/bit** |          | N      |          |       |       |         |
| **bit[n]**   |          |        | N        |       | N     |         |
| **int/uint** |          | N      |          |       |       |         |
| **float**    |          |        |          |       |       |         |
| **angle**    |          |        |          |       | N     |         |
| **complex**  |          |        |          |       |       |         |

## Legend for the explicit and implicit cast tables

The following rules specify the semantics of the allowed explicit and implicit casts:

**N**
Cast is disallowed.

**R1 (Scalar bit equivalence).**
Scalar `bit` is interchangeable with `bool` for value casts, so wherever a rule
or table cell mentions `bool`, scalar `bit` behaves identically. Scalar `bit` is
*not* the register type `bit[1]`: the two are distinct types (`types.rst` lines
227 to 229). Scalar `bit` casts like `bool`, whereas `bit[1]` casts by width
match (R3).

```qasm
bit b = 1;
bool flag = bool(b);          // allowed: scalar bit and bool are interchangeable
int[32] n = int[32](b);       // allowed: b acts as bool, and bool -> int is allowed
bit[2] pair = bit[2](b);      // allowed: bool -> bit[n] of any width (note [a])
angle[1] ang = angle[1](b);   // not allowed: scalar bit is bool, not bit[1]; bool -> angle is No
```

**R2 (C99-casts).**
Within `int`, `uint`, `float`, and `complex`, all size variants (sized or
unsized) cast freely in both directions; except for `complex -> int/uint/float`.
Mimmicking C99 semantics.

```qasm
float[32] f = 3.14;
float[64] g = float[64](f);                      // allowed: widen within float
float[16] h = float[16](g);                      // allowed: narrow within float
complex[float[32]] c = complex[float[32]](2.0);
float[32] back = float[32](c);                   // not allowed: casts from complex to int/uint/float
```

**R3 (Width-matched bit bridge).**
`bit[k]` casts to and from `int[k]`, `uint[k]`, `angle[k]`, and other `bit[k]`
only when the widths are equal, using an exact little-endian bit copy. Width
mismatches require explicit slicing.

```qasm
bit[4] bits = "1010";
int[4]   i = int[4](bits);    // allowed: equal width, little-endian bit copy
angle[4] a = angle[4](bits);  // allowed: equal width
bit[4]   c = bit[4](bits);    // allowed: bit[k] <-> bit[k] at equal width
int[8]   j = int[8](bits);    // not allowed: 4 != 8, slice explicitly first
```

**R4 (Angle casts).**
- `float[*]` casts to `angle[*]` one way only, for any source and target size.
- `angle[n]` to `angle[m]` widening (`m > n`) zero-pads the least significant
  bits; `angle[m]` to `angle[n]` narrowing rounds or truncates, controlled by a
  `#pragma`.
- Unsized `angle` does not cast to or from sized `angle[k]`. (This is the open
  question flagged in the PR.)

```qasm
angle[20] a = angle[20](3.14);    // allowed: float -> angle (one way only)
float[64] f = float[64](a);       // not allowed: angle -> float
angle[32] wide = angle[32](a);    // allowed: widen, zero-pads least significant bits
angle[10] narrow = angle[10](a);  // allowed: narrow, rounds or truncates (#pragma)
angle un;
angle[16] s = angle[16](un);      // not allowed: unsized angle does not cast to/from angle[k]
```

**R5 (truthy and falsy values)**
Zero values cast to `false` / `0`, and non-zero values cast to `true` / `1`.

**R6 (bool/bit to bit[n])**
`bool` (and scalar `bit`) casts to `bit[n]` of any width: the low bit
  takes the value, the rest are zero. This is the one behavior that is neither a
  rule nor derivable from another cell.

  ```qasm
  bool flag = true;
  bit[4] bits = bit[4](flag);   // allowed: low bit takes the value -> "0001", rest zero
  ```

**R7 (bool/bit to int / uint / float)**
`false` / `0` and `true` / `1` cast to the zero and one values of `int`, `uint`, and `float` respectively.

**Note: duration**
Casts between `duration` and other types are not allowed.

orpuente-MS:
Now, after fleshing out these condensed tables. I realized two things:

1. C99 allows explicit and implicit casts from `complex` to `bool`. So, maybe we should allow it too.
2. In my table 20x20 implicit cast table I had casts from `angle` to `bool/bit` disallowed. This was based on either an example or a paragraph of the spec I cannot find anymore. So, I removed the `No` when writing the condensed implicit table.

Any thoughts on point (1)?

mbhealy:
I really like the condensed table formats. 

I'm generally fine with allowing casts from `complex` to `bool`. 
I also think that casting from `angle` to `bool` should be allowed.

mbhealy:
> Unsized angle does not cast to or from sized angle[k]. (This is the open question flagged in the PR.)

I'm having difficulty remembering exactly why we wanted to disallow this. I can't think of a good objection, maybe someone else remembers?

nulinspiratie:
Hey @orpuente-MS, great work on the follow-up (and apologies for the delayed response).
In general, I agree with the points you made. Most importantly, pointing to the rules, where possible, makes sense. One thing I might want to propose here is that when the answer is a clear Y, it may be worth also stating that in the table for convenience, e.g. `Y (R1)`. That way, the reader does not necessarily need to consult the corresponding rule if the answer is always yes. What do you think?

Also happy to remove `duration` from this; the note clarifies it sufficiently.

And regarding casting complex to bool, my concern is that it can lead to machine-precision bugs. But I do see the value in sticking close to a known format, so I am fine either way. 

I'm happy with the condensed cast tables proposal by @orpuente-MS. I also won't seem to be able to join TSC meetings in the near future (overlapping meeting), so the in-person discussions on this will need to happen without me. I'm fine with whatever the TSC chooses regarding this PR

athirab751:
Hi, I'm Athira Babu, Data Scientist at IBM Kochi with a PhD in Mathematics — @jwoehr suggested I take a look at this from Issue #298.

The angle <-> angle[k] question caught my attention. An unsized angle has no fixed precision, so casting to angle[k] would silently require choosing one - and casting back would silently discard it. Both feel like they should be explicit decisions by the programmer, not something the compiler assumes. Disallowing these and requiring explicit slicing (consistent with R3) seems the safer path.

On complex -> bool - C99 treats a complex as false only when both components are zero, which is the standard mathematical definition of the zero element in set of complex numbers. That seems reasonable to follow here.

Issue #611, opened 2025-07-09
[proposal]: Inferred OpenQASM version and parser/compiler behavior should be defined in the spec


idavis:
### Description

- What are the inferred major and minor versions if the OPENQASM statement is missing?
- What is the expected behavior of a parser/compiler when a feature is used from a version newer than the declared version?

### Examples

Example: There is no version declared, and two 3.1 features are used here: anonymous scopes and switch statement.
```openqasm
{
  const int A = 0;
  const int B = 1;
  int i = 15;
  
  switch (i) {
  case A {}
  case B {}
  case B+1 {}
  default {}
  }
}
```

blakejohnson:
Thanks for this proposal, @idavis.

Row 14 in the table makes me uneasy, because having OpenQASM spec identifiers in include files seems like a good housekeeping strategy to mark author intent, particularly for a file which might get re-used in different contexts. Is there an alternative where we allow these identifiers in include files, but throw an error if the version numbers don't match the parent. Then these declarations become metadata that is dropped during textual inclusion.

idavis:
@blakejohnson This would require changing the spec and how include files are interpreted as include files are supposed to behave like they were defined inline at the site, which would put the statement in an illegal position. Do we want to open up that area for spec changes?

hodgestar:
@idavis For concreteness, is the main question here whether we would like to make the statements under `Summary of key rules` officially part of the specification?

hodgestar:
I would be very happy for a PR to be created from the proposed rules and TSC discussion.

PR #669, opened 2026-05-13
Defined version parsing behavior in the specification

idavis:
### Summary

Specifies and implements version parsing behavior brought up in #611

### DIFF
--- a/source/openqasm/openqasm3/parser.py
+++ b/source/openqasm/openqasm3/parser.py
@@ -34,8 +34,9 @@ __all__ = [
 ]
 
 import re
+import warnings
 from contextlib import contextmanager
-from typing import Union, TypeVar, List, Optional, Protocol, Tuple, cast
+from typing import Dict, Union, TypeVar, List, Optional, Protocol, Tuple, Callable, cast
 
 try:
     from antlr4 import (
@@ -58,6 +59,7 @@ from ._antlr.qasm3Lexer import qasm3Lexer  # type: ignore[import-not-found]
 from ._antlr.qasm3Parser import qasm3Parser  # type: ignore[import-not-found]
 from ._antlr.qasm3ParserVisitor import qasm3ParserVisitor  # type: ignore[import-not-found]
 from . import ast
+from . import spec
 
 _TYPE_NODE_INIT = {
     "int": ast.IntType,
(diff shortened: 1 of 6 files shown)


blakejohnson (commented):
The text changes to the spec look consistent with our discussions. I'm still reviewing the changes to the reference parser.

source/openqasm/openqasm3/parser.py
```diff
@@ -178,11 +180,87 @@ def parse_version(prog: str) -> Optional[Tuple[int, ...]]:
...
 
-MIN_SUPPORTED_VERSION = (3,)
-MAX_SUPPORTED_VERSION = (3, 1)
+def _parse_spec_version(v: str) -> Tuple[int, ...]:
+    """Convert a version string like '3.1' to a tuple like (3, 1)."""
+    return tuple(int(x) for x in v.split("."))
 
 
-def parse(input_: str, *, permissive=False, ignore_version=False) -> ast.Program:
+_parsed_versions = [_parse_spec_version(v) for v in spec.supported_versions]
```
blakejohnson, line 188: Is it time to add "3.2" to `spec.supported_versions`?

idavis: We should probably make 3.2 official in lock step with the `spec.supported_versions`

idavis: I don't remember why I added `"quantum_call_expression": (3, 2),`

source/openqasm/openqasm3/parser.py
```diff
@@ -194,19 +272,48 @@ def parse(input_: str, *, permissive=False, ignore_version=False) -> ast.Program
     :param ignore_version: If true, ignore the specified version of the OpenQASM program, and
         attempt to parse it anyway.  There is no guarantee that the output was syntactically or
         semantically valid for the given version if this is set.
+    :param include_resolver: An optional callable that takes an include filename and returns
```
blakejohnson, line 275: Why does `include_resolver` default to `None`? Are you trying to allow for magical availability of `stdgates.inc` or did you have something else in mind?

source/openqasm/openqasm3/parser.py
blakejohnson: > We should probably make 3.2 official in lock step with the spec.supported_versions

Sure. Then I guess tests can just override `MAX_SUPPORTED_VERSION` to ensure feature-gating works for un-released versions.

source/openqasm/openqasm3/parser.py
```diff
@@ -178,11 +180,87 @@ def parse_version(prog: str) -> Optional[Tuple[int, ...]]:
...
+        # we don't raise here, but we do remove the filename from the visited
+        # set so that other branches can include it.
+        _visited.discard(stmt.filename)
+
+
+def parse(
+    input_: str,
+    *,
+    permissive=False,
+    ignore_version=False,
```
blakejohnson, line 261: Doesn't `ignore_version` now need to default to `True` to be consistent with the spec change?

source/openqasm/openqasm3/parser.py
blakejohnson: > I don't remember why I added "quantum_call_expression": (3, 2),

That does feel wrong.

source/openqasm/openqasm3/parser.py
```diff
@@ -813,6 +945,7 @@ def visitMeasureExpression(self, ctx: qasm3Parser.MeasureExpressionContext):
 
     @span
     def visitQuantumCallExpression(self, ctx: qasm3Parser.QuantumCallExpressionContext):
+        self._check_feature_version(ctx, "quantum_call_expression")
```
blakejohnson, line 948: why?

source/openqasm/openqasm3/parser.py
blakejohnson: Nevermind, I was reading this backwards.

source/openqasm/openqasm3/parser.py
blakejohnson: I see the problem now. `parse` operates assuming the entire input is provided as a `str`. It can't make any assumptions about a file system or where to find included files.

source/openqasm/tests/test_qasm_parser.py
```diff
@@ -2639,14 +2640,348 @@ def test_comment_preservation():
...
+        source = """
+        OPENQASM 3.0;
+        qubit q;
+        bool b = true;
+        if (b) { x q; }
+        """
+        result = parse(source)
+        assert isinstance(result, Program)
+
+    def test_nop_rejected_below_3_2(self):
```
blakejohnson, line 2966: Do we want the positive version of this test as well? (requires overriding MAX_SUPPORTED_VERSION)

blakejohnson (commented):
I think the only thing still to resolve is the question about `quantum_call_expression`.

blakejohnson:
Some notes from the TSC discussion today:

* We discussed whether we wanted to provide an include resolver in the reference parser that loads from from a list of search paths.
  * We noted that the current parser doesn't actually include anything, but generates an appropriate AST node referencing the include.
  * Ian will open a issue to break off the conversation about OpenQASM's include loading mechanism
* We discussed the requirement that included files have a version <= to the outermost version

idavis:
Replaced by #679

PR #671, opened 2026-06-03
Docs: Updated the source/language/classical.rst to make description of while more precise

### LINKED ISSUE #670: Make description of `while` loop more precise
The following corrections (or equivalent) should be made:

* Avoid implying that the body is executed once unconditionally.
* `bool`  should be `<condition>`
* The text should refer to `<body>` rather than `program`

https://github.com/openqasm/openqasm/blob/28186ad4ea1ed4e1b7d0853985506b7320ed8a70/source/language/classical.rst?plain=1#L358-L362

Shanwis:
### Summary

Closes #670. Corrects the `while` loop description in `source/language/classical.rst`:

- Use `<condition>` instead of `bool` as the metavariable placeholder
- Rephrase to correctly describe pre-check semantics: "repeatedly executes `<body>` as long as the condition evaluates to true"
- Use `<body>` instead of "program" for consistency with the rest of the spec

### Details and comments

No grammar changes. Docs-only fix to align the while loop description with the existing `for` loop conventions and the actual semantics of the language.

### DIFF
--- a/source/language/classical.rst
+++ b/source/language/classical.rst
@@ -355,8 +355,8 @@ accessible after the loop.
 While loops
 ~~~~~~~~~~~
 
-The statement ``while ( bool ) <body>`` executes program until the Boolean evaluates to
-false [3]_. Variables in the loop condition statement may be modified
+The statement ``while ( <condition> ) <body>`` repeatedly executes ``<body>`` as long as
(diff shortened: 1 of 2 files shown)


Shanwis:
@jlapeyre is this what you had in mind?

hodgestar (approved):

Issue #622, opened 2025-08-19
[Spec issue]: `a[{1, 2, 3}]` indexing is underspecified and has a contradiction


orpuente-MS:
### OpenQASM version:

3.0, 3.1, Live

### Python version:

N/A

### Operating system:

N/A

### Missing info
The section on index sets is empty: <https://openqasm.com/language/types.html#index-sets-and-slicing>. I think something should be here, or maybe this header should be deleted?

### Naming convention in Grammar and Spec diverge

In the Grammar: A whole index with brackets (including `{a, b, c}`, a single expr, or a range) is called an indexOperator. And `{a, b, c}` is called a setExpression.
 
In the spec, a whole index is called an IndexSet, and a `{a, b, c}` index is not called by any names, it's just refered to as "a comma-separated list of integers contained in braces".

I think the Grammar and Spec would benefit from some common naming here.

### Contradiction about `{a, b, c}` indices

The spec says that ["a register cannot be concatenated with any part of itself"](https://openqasm.com/language/types.html#register-concatenation-and-slicing).

However, `{a, b, c}` indices allow just that. That is,

```
bit[8] a;
let a_alias = a[0:0] ++ a[2:2] ++ a[4:4];
```

which is disallowed, is equivalent to,

```
bit[8] a;
let a_alias = a[{0, 2, 4}];
```

which is allowed.

I think that to solve this contraction, either:
1. A register should be allowed to be concatenated with itself, or
2. `{a, b, c}` indices should not be allowed in alias statements. Which means they wouldn't get used anywhere, so in this case we could remove them from the Spec.

### Steps to reproduce the problem

N/A

### What is the expected behavior?

N/A

### Suggested solutions (Optional)

N/A

orpuente-MS:
@jakelishman , do you know if the indices in the `{a, b, c}` are meant to be const? Or can they be runtime values.

If they are not const, we won't be able to forbid multiple aliases to the same bit at editing/design time.

jakelishman:
I think registering aliasing as a whole was vaguely supposed to be a compile-time thing, but I don't think it was fully formalised either way.  The `let` syntax was introduced before my involvement with the language.

PR #676, opened 2026-07-22
Accept array concatenation in declarations

### LINKED ISSUE #432: []: Concatenation in ClassicalDeclarationStatement raises error during parsing
### OpenQASM version:

3.0 (0.4.0)

### Python version:

3.11

### Operating system:

macOS 13.1 (Ventura)

Parsing a ClassicalDeclarationStatement with a concatenation operator raises an error (I think the parser is trying to match to a semicolon when it instead encounters the concatenation operator).

Create a new Conda environment

```
conda install pip
pip install openqasm3
pip install antlr4-python3-runtime
pip install ipython
ipython -i
```

```python
import openqasm3

qasm_str = """
array[int[8], 2] first = {0, 1};
array[int[8], 3] second = {2, 3, 4};

array[int[8], 5] concat = first ++ second;
"""

openqasm3.parse(qasm_str)
```

raises error when processing the last line, a ClassicalDeclarationStatement with the concatenation operator ('++') 

This is indicated to be possible by the live specification on array slicing (https://openqasm.com/language/types.html#array-slicing), the example used here are the first 3 lines from the code example in the array slicing section. 

I think this error is caused by the double plus concatenation operator only be used for `aliasExpression` statements and not for regular `expression` statements which can be used in `classicalDeclarationStatement`. 

Based on the discussion in https://openqasm.com/language/types.html#register-concatenation-and-slicing this seems very deliberate, so I don't think the solution is to just add the '++' operator to the 'expression' grammar. 

_No response_

### LINKED ISSUE #620: [Grammar Bug]: `++` operator only appears in alias expr but it's used for array concatenation
### OpenQASM version:

3.0, 3.1, Live

### Python version:

N/A

### Operating system:

N/A

The Grammar only allows the `++` operator in alias statements. However, the Spec introduces array concatenation using the `++` operator, and uses it as the right-hand side of classical declaration statements.

N/A

The Grammar should allow `++` concatenation expressions as the right-hand side of classical declaration statements.

N/A

sankalpsthakur:
### Summary

Accept array concatenation expressions as initializers in classical declarations in both the reference grammar and the Python AST parser.

**User impact:** Valid OpenQASM 3 programs that concatenate arrays in declarations (as shown in the spec's array-slicing examples) previously failed to parse, blocking toolchain authors and compiler front-ends.

Fixes #620
Fixes #432

### Details and comments

OpenQASM permits array concatenation with `++` in declaration initializers, but the declaration grammar currently routes through an expression rule that excludes concatenation. The Python parser also assumes a single expression node in this position.

This change adds the declaration-only grammar alternative and folds concatenation operands into nested `Concatenation` AST nodes. The fold is iterative so long declarations do not hit Python's recursion limit. The iterative left fold is equivalent to the previous reverse-iterator recursive construction: both yield a left-associative tree `(a ++ b) ++ c`. Tests cover indexed operands, three operands, shared alias/declaration associativity, and a 1,101-operand declaration while preserving the existing assignment grammar boundary.

Previously reported validation during development (not rerun for this description update):

- 186 reference-grammar tests passed with ANTLR 4.9.2
- 737 Python parser/AST tests passed across supported ANTLR runtimes, including 4.7.2, 4.9.2 and 4.13.0
- Black, Pylint, mypy, and `git diff --check`

### AI/LLM disclosure

AI coding tools were used to help draft this change: OpenAI Codex for the implementation and associativity follow-up commits, and Cursor for the release-note reference update. Each commit carries an `AI assistance:` trailer identifying the tool. Codex also assisted with this description update.


source/openqasm/openqasm3/parser.py
```diff
@@ -893,21 +893,23 @@ def visitLiteralExpression(self, ctx: qasm3Parser.LiteralExpressionContext):
...
-            return add_span(ast.Concatenation(lhs=lhs, rhs=rhs), combine_span(lhs.span, rhs.span))
+        return self._visit_concatenation(ctx.expression())
 
+    @span
+    def visitConcatenationExpression(self, ctx: qasm3Parser.ConcatenationExpressionContext):
+        return self._visit_concatenation(ctx.expression())
+
+    def _visit_concatenation(self, expressions):
         # This iterator should always be non-empty if ANTLR did its job right.
-        iterator = reversed(ctx.expression())
```
blakejohnson, line 909: The original code made an explicit choice to walk the expression with a reversed iterator. That appears to be dropped in the changeset. What is the justification for this change?

sankalpsthakur: Good catch — the reversed walk was intentional in the old recursive helper, and the left-associative AST shape is preserved.

**Why reverse was used originally**

The recursive visitor walked operands from right to left so that recursion could assemble a *left*-associative tree:

```text
a ++ b ++ c  =>  (a ++ b) ++ c
```

The reverse was an artifact of that recursive construction, not a separate semantic choice.

**Why the iterative version walks forward**

This PR shares one helper for alias and declaration concatenations, and declaration initializers can be very long (the suite includes a 1,101-operand case). A recursive walk hits Python’s recursion limit on those inputs, so the helper is an iterative left fold:

```python
lhs = visit(e0)
for e in e1..en:
    lhs = Concatenation(lhs, visit(e))
```

That produces the *same* left-associative tree as reverse+recurse. Walking the iterative fold with a reversed iterator would invert operand order / associativity, so reverse is not restored here.

Comments in `_visit_concatenation` now document this, and tests assert the shared left-associative shape for both `let … = a ++ b ++ c` and declaration initializers, plus the long left-spine chain.

Happy to adjust further if you’d prefer a different presentation of that history in the comment.

blakejohnson:
This changeset appears to be generated by an AI tool. Can you please document which AI tool you used in the commit messages?

sankalpsthakur:
Re the reversed iterator: the old code walked operands in reverse so a recursive visitor could still build a left-associative tree `(a ++ b) ++ c`. The new helper does the same fold iteratively, left to right — walking forward is what keeps that shape. Reversing the iterator here would flip the tree; tests for alias/declaration associativity and the 1,101-operand chain lock in the left-associative result.

sankalpsthakur:
I'd rather keep the commits focused on the change itself. Happy to answer any technical questions on the diff though.

sankalpsthakur:
Yes—AI tools were used. I have amended the commit messages to disclose them directly: OpenAI Codex for the implementation and associativity follow-up commits, and Cursor for the release-note reference update. The rewrite is metadata-only; the resulting code tree is unchanged. Sorry that my earlier response did not answer the question directly.

sankalpsthakur:
The commit messages already carry this — each commit has an `AI assistance:` trailer (`090a31b` and `f2a6791`: OpenAI Codex; `d32f1f6`: Cursor, with a `Co-authored-by` trailer). They may not be visible from the Files view. I've left them as-is; happy to also add a disclosure section to the PR body if that's the more discoverable place for it.

sankalpsthakur:
added the disclosure section to the pr body.

idavis:
We need to look at #622 first - then we can come back to this.

PR #678, opened 2026-08-04
Define include-file resolution and idempotent inclusion

### LINKED ISSUE #677: [proposal]: Define include-file resolution and idempotent inclusion
### Description

OpenQASM 3 defines `include "filename";` as textual insertion and defines nothing else. There is no rule for how a target name becomes source text, no rule for relative targets, no notion of file identity, no duplicate rule, and no cycle rule. This proposal supplies those rules without changing the grammar.

The proposal has seven parts.

**1. An include target is an opaque logical name.**

Resolution is an implementation-defined mapping from an include target plus the canonical name of the including source unit, to a canonical name plus source text. The specification defines three terms and does not define a lookup algorithm:

* *include target*: the string literal written in the `include` statement.
* *canonical name*: the identity produced by resolution. Two includes that resolve to the same canonical name refer to the same source unit.
* *source unit*: the content associated with one canonical name.

This is the posture C, C++, and Fortran already take for `#include` and `INCLUDE`. It is what allows filesystem, in-memory, notebook, editor-buffer, and sandboxed hosts to all be conforming.

**2. A mandated observable contract.**

Lookup stays implementation-defined; observable behavior does not. Within a single compilation:

* resolving the same target from the same includer shall yield the same canonical name
* the same canonical name shall yield the same source text
* a relative target shall resolve against the canonical name of the including source unit, never against a process working directory
* an inclusion cycle must be diagnosed
* implementations must support an include depth of at least 15

Stability is scoped to a single compilation deliberately. Resolution results legitimately change between compilations under live editing, which is why TypeScript ships an explicit invalidation hook. A globally stated stability requirement would be unimplementable in a language server.

**3. Provenance identity: a source unit is incorporated at most once per canonical name.**

A subsequent `include` that resolves to an already-incorporated canonical name is a no-op: it does not insert that source unit's statements a second time. Consequently, declarations from a source unit incorporated through one include site are not redeclared when another include resolves to the same canonical name. Declarations from distinct canonical source units are each incorporated. Any name conflicts are handled by the language's ordinary name, scope, and overloading rules, not by a separate include-resolution collision rule.

The rule is stated at the source-unit level, not the symbol level. This matters: pragmas, `cal` blocks, and top-level statements are not symbols, and a symbol-level rule would leave them duplicated while a source-unit-level rule does not. Only the source-unit-level form is self-consistent.

This is the part that fixes the reported problem, and it needs no new syntax.

**4. Existing `defcal` overloading remains permitted.**

The incorporation-once rule does not change the language's existing treatment of defcal. Declarations from distinct incorporated source units remain subject to the ordinary redeclaration rules, except that multiple defcal statements for the same operation remain permitted as overloads, as specified today.

**5. A reserved canonical name for the standard library.**

The canonical name `stdgates.inc` is reserved for the standard library and does not depend on a literal file existing. Implementations that make the standard library available should consider it locatable independently of ordinary include search-path settings, and may synthesize it rather than supply a literal file. The standard library is by far the most likely double include, so without this the mechanism fails on its most important case.

The reference parser represents this reserved source unit as an empty virtual unit when callers explicitly enable include resolution. This preserves canonical identity, provenance, and incorporation-once behavior without adding parser-owned gate declarations. Downstream semantic tooling is responsible for making the specified gates available, resolving their symbols, and type checking their uses.

**6. No promise about physical file identity.**

Two distinct canonical names that happen to resolve to the same physical content are two source units. The specification makes no claim about inodes, symlinks, case-insensitive filesystems, or content equality.

This is a deliberate limit, not an oversight. There is no portable definition of physical file identity: Clang uniques by inode, GCC compares content, MSVC documents that aliasing defeats canonical-path identification, and Python reports `st_ino` as zero on Windows.

**7. `.` and `..` are prohibited in include targets.**

This is the cheapest traversal defense available at the specification level because it requires no filesystem semantics at all. Protobuf and Bazel both do it. Traversal through include mechanisms is an exploited vulnerability class.

### Guidance: include files that read the includer's scope

The scoping chapter's own example is not self-contained: `my_definitions.qasm` computes `int j = i + 5;` where `i` is declared by the includer, and a comment blesses this. Under part 3, such a unit is incorporated at its *first* include site only.

Because OpenQASM has one flat global scope, `i` can be declared only once in an entire program, so this cannot produce a differing value. What it does change is that a second include of such a unit becomes a silent no-op where today it is a duplicate-declaration error. The proposal keeps the rule universal and adds guidance that a source unit which reads its includer's scope is not safe to include more than once. It does not deprecate includer-scope inheritance.

### No escape hatch

There is deliberately no mechanism for incorporating the same source unit twice. No surveyed module system provides one at the include level, and the C patterns that motivate one, X-macros and the `<assert.h>` carve-out, require macros or conditional compilation that OpenQASM does not have.

### What this does not change

No grammar change. `includeStatement: INCLUDE StringLiteral SEMICOLON;` is untouched, and the `.`/`..` prohibition constrains string content rather than syntax.

### Non-normative note on host integration

The specification will carry a non-normative note describing four host load behaviors, so implementations in any language can offer a mirrored surface:

1. **No resolver.** Includes are not loaded. The include statement is retained unresolved.
2. **A mapping** from canonical name to source text. The keys are canonical names. This is the form notebooks, test harnesses, and virtual filesystems want.
3. **A callable** receiving the include target *and the including source unit's canonical name*, returning source text plus a canonical name, or nothing to indicate the target is not its concern. The includer must be passed, or relative resolution is impossible.
4. **A resolver object**, supporting composition and fallthrough.

The reference parser also composes its reserved-name standard-library resolver ahead of the caller-selected resolver when a caller opts into resolution through forms 2 through 4. This is parser policy, not a fifth host integration form: it supplies canonical identity and provenance for `stdgates.inc`, while downstream semantic tooling supplies the standard-library declarations.

The note is non-normative because requiring these forms would contradict part 1.

### Alternatives considered

**An opt-in `once` marker** (`pragma openqasm.once`, `@openqasm.once include "x.inc";`, or `include once`). Rejected because both cheap forms are blocked. An includee-side pragma contradicts the specification's own guidance that pragmas "should avoid stateful or positional interactions to avoid unexpected behaviors between included source files", which names included files explicitly. An includer-side annotation reproduces GCC's `#import`, which GCC's manual calls not well designed because it forces every user to know a file is include-once. `include once` needs new grammar. Part 3 dissolves the problem without any of them.

**Benign redefinition by textual identity** ("a `gate` declaration textually identical to an existing one is a no-op"). Rejected because token identity is unsound for declarations. C and C++ specify it only at the preprocessor layer, over context-free token blobs, and C++ documents character-for-character identical class definitions that still violate the One Definition Rule because an earlier differing declaration changes what the identical tokens mean. OpenQASM is more exposed than C++ here, not less: the scope is flat and global, and gate bodies reference other gates by bare name.

**Special-casing the standard library only.** Rejected as the primary fix, retained as part 5. It is cheap and the specification already permits treating `stdgates.inc` as virtual, but it does nothing for user libraries, and every implementation already special-cases the standard library differently and incompatibly.

**Specifying nothing and leaving resolution implementation-defined.** Rejected because that is the status quo, and the implementations summarized below already differ in relative lookup, standard-library precedence, duplicate handling, and cycle handling. This proposal preserves the defensible part of the position by leaving the lookup algorithm implementation-defined while constraining observable behavior.

**A module or namespace system instead of textual inclusion.** Rejected for now, not forever; this is issue #365. The documented cost is large: C++20 modules still lack header-unit support across compilers, and Sass needed from 1.23.0 to 1.80.0 plus a dedicated migrator to deprecate `@import`. That is disproportionate to unblocking #611.

**Deferring same-name conflict diagnosis to the use site**, as Haskell does. Rejected because Haskell can defer safely: a qualified name always repairs the ambiguity. OpenQASM has no namespacing, qualification, or aliasing, so deferral would leave the conflict unrepairable and merely later.

**A mapping-only host API with no callable form.** Rejected as sufficient, accepted as one of four supported forms. A mapping cannot express relative resolution or lazy lookup, but it is the right convenience form for in-memory, notebook, and test hosts.

### Reference parser

The `openqasm3` reference parser implements the parser-facing include-resolution behavior: target validation, resolver stability checks, cycle and depth diagnosis, once-per-canonical-name incorporation, reserved `stdgates.inc` identity, and the four resolver forms. It remains a structural AST parser and does not perform general semantic duplicate-declaration validation.

When a caller explicitly enables include resolution, the reference parser resolves exact `stdgates.inc` before caller-supplied mappings, callables, or resolver objects. It records an empty source unit with the reserved canonical name and does not ship, load, or synthesize standard-gate declarations. The standard library is defined by the specification; downstream compiler tooling performs symbol resolution, type checking, and selects language-version-specific gate implementations. With no resolver, `stdgates.inc` remains unresolved exactly like every other include target. The parser does not load the standard library from a process working directory or from a repository example file.

### Tier 2, explicitly deferred

* `include once` as a first-class modifier, if explicit control is later preferred to the implicit rule. Requires a grammar change.
* Namespacing or a module system: issue #365.
* Whether an included file may carry its own version statement: issue #611. This proposal makes each source unit addressable by canonical name, which is a prerequisite for answering it, but does not answer it.

### Motivation

**Two independently valid files cannot be combined.**

Both files below are valid OpenQASM 3 today, and each can be verified on its own. Combining them is an error, because the standard gates get declared twice and the specification states that identifiers declared by `gate`, `def`, and `defcal` "may not be shadowed or otherwise redeclared".

There is no workaround. C solves this with wrapper `#ifndef` guards, which GCC calls "the standard way", but OpenQASM has no conditional compilation and no macros, so an author cannot write one.

**This is a regression from OpenQASM 2.**

OpenQASM 2 stated: "The path is specified relative to the current working directory." OpenQASM 3 dropped that sentence and did not replace it. OpenQASM 3 is therefore strictly *less* specified than OpenQASM 2 on include resolution, and no issue in this repository remarks on the omission. This proposal restores a lost guarantee rather than inventing a new capability.

The working-directory rule itself should not be revived. It is not implementable in the hosts people actually use: Jupyter has no supported API for a notebook's own path and its maintainers have called the question ill-defined, `__file__` is optional by specification, `sys.path[0]` is documented as a potentially unsafe path, and the process working directory is documented as time-dependent. Includer-relative resolution is what every modern comparator uses.

**The specification already assumes the mechanism it never defines.**

The standard-library chapter tells implementations that `stdgates.inc` "should be locatable, regardless of any other settings that they might apply around include search paths". That is an include-resolution rule written without an include-resolution model behind it.

**Six implementations have each invented a different answer.**

| Implementation | Loads includes? | Resolution | Standard library | Duplicate handling | Cycles |
|---|---|---|---|---|---|
| `openqasm3` reference parser | Opt-in | mapping, callable, or resolver object; no ambient lookup | reserved virtual source unit; downstream semantics | incorporated once by canonical name | explicit error |
| Qiskit `qasm2` / `qasm3` | Yes | ordered include path, first match | built-in `qelib1.inc` table always beats a real file | none | none |
| Qiskit `openqasm3_parser` | Yes | absolute, then search paths, then `QASM3_PATH` env var | synthesized, file never read | none | `// FIXME: prevent a file from including itself` |
| staq / qasmtools | Yes | real file first, built-in as fallback | inverse of Qiskit's precedence | none | none |
| pyqasm | Yes | includer dir, then working directory | hoisted | preprocessor skips silently by basename; AST visitor hard-errors | explicit error |
| Braket simulator | Yes | bare `open()` against the working directory | none | none | none |
| microsoft/qdk | Yes | includer-relative only | name-matched, gates injected | hard error, even for a diamond | explicit error |

Two of these disagree on standard-library precedence in exactly opposite directions. One is internally inconsistent with itself. A program that works on one toolchain can silently mean something different on another.

**This blocks in-flight work.**

Issue #611 needs to decide how version statements in included files behave. Earlier TSC notes on PR #669 recorded that the parser did not load includes and that a separate issue would define the mechanism; the reference parser implementation now supplies that mechanism while retaining version statements on included source units without applying them.

Issue #365 already enumerated the option space in 2022, including header guards, include-once, a module system, and allowing redefinition. It records "no immediate consensus" and was deferred until after the 3.0 release. This proposal takes the narrowest option from that list that actually solves the problem.

**Related:** #611, #365, #230, #507, #516, PR #669, PR #566, PR #435.

### Authors

@idavis

### Examples

The problem. Each library file is independently valid:

```text
// a.inc
include "stdgates.inc";
gate my_a q { h q; }
```

```text
// b.inc
include "stdgates.inc";
gate my_b q { x q; }
```

Without the incorporation-once rule, combining them fails:

```text
OPENQASM 3.0;
include "a.inc";
include "b.inc";
// ERROR without incorporation once: 'h', 'x', and every other standard gate is redeclared
// by the second inclusion of stdgates.inc.
```

Under this proposal the second inclusion of the canonical name `stdgates.inc` contributes nothing. A semantic implementation makes the standard-library gates available once; the reference parser records the virtual source identity and leaves those gate symbols to downstream tooling.

Relative resolution is against the includer, not the working directory:

```text
// lib/gates/extra.inc, canonical name "lib/gates/extra.inc"
include "helpers.inc";   // resolves to lib/gates/helpers.inc
```

Same-named declarations from different source units are incorporated as distinct source units. A semantic implementation may reject ordinary redeclarations according to the language's name, scope, and overloading rules, but include resolution itself does not define an additional collision rule:

```text
OPENQASM 3.0;
include "vendor_a.inc";   // declares: gate flip q { x q; }
include "vendor_b.inc";   // declares: gate flip q { y q; }
// The include mechanism incorporates both source units. Ordinary semantic validation
// determines whether the duplicate `flip` declaration is accepted.
```

`defcal` is unaffected, preserving current behavior:

```text
OPENQASM 3.0;
include "cal_q0.inc";   // defcal rx(angle) $0 { ... }
include "cal_q1.inc";   // defcal rx(angle) $1 { ... }
// No error: multiple defcal statements overload rather than redeclare.
```

Prohibited include targets:

```text
include "./gates.inc";      // ERROR: '.' component
include "../lib/gates.inc"; // ERROR: '..' component
```

Cycles are diagnosed:

```text
// x.inc includes y.inc, y.inc includes x.inc
// ERROR: inclusion cycle.
```

idavis:
## Functional summary

This branch makes include resolution an opt-in parser capability. It introduces `IncludeEnvironment` as the single boundary that combines source lookup with owner-local library bindings. Callers provide one `includes=` value to `openqasm3.parse` instead of coordinating source data and binding maps through separate parser arguments.

> Implements / Closes #677

## Resolve sources in memory

`parse` accepts a source mapping through its single `includes=` channel. The parser resolves the include, records both source units, and links the include statement to the resolved source index.

```python
import openqasm3

source = """\
OPENQASM 3.0;
include "gates.inc";
qubit q;
custom q;
"""
includes = {
    "gates.inc": "gate custom q { U(0, 0, 0) q; }\n",
}

program = openqasm3.parse(source, includes=includes, name="main.qasm")

assert [unit.canonical_name for unit in program.sources] == [
    "main.qasm",
    "gates.inc",
]
assert program.statements[0].target == "gates.inc"
assert program.statements[0].source == 1
```

Mappings, callables, and `IncludeResolver` implementations are normalized into the same resolver contract. `FileSystemResolver` is available when source text must be loaded from an explicit set of search roots.

## Scope library names by owner

`IncludeEnvironment.scoped` binds the apparent library names written by each source owner to canonical library names. A source cannot reach a library merely because some other source loaded it.

```python
import openqasm3
from openqasm3.resolver import IncludeEnvironment

sources = {
    "main.qasm": 'OPENQASM 3.0; include "@v/gates.inc";',
    "@vendor/gates.inc": "gate custom q { U(0, 0, 0) q; }",
}
environment = IncludeEnvironment.scoped(
    sources,
    {
        "": {"v": "vendor"},
        "vendor": {},
    },
)

program = openqasm3.parse(
    sources["main.qasm"],
    includes=environment,
    name="main.qasm",
)

assert program.sources[1].canonical_name == "@vendor/gates.inc"
```

An empty owner binding map fails closed. The resolver reports a missing target instead of exposing a library outside that owner's declared scope.

## Preserve or flatten the include structure

Resolved sources remain separate from the root AST. Printing preserves the original include statement, while `flatten_includes` produces a statement list with included declarations spliced into place.

```python
rendered = openqasm3.dumps(program)
flattened = openqasm3.flatten_includes(program)

assert 'include "@v/gates.inc";' in rendered
assert len(flattened.statements) >= len(program.statements)
```

`stdgates.inc` is always resolved before caller backends as a synthetic empty source unit. Downstream compiler tooling remains responsible for standard-gate definitions and type checking.

## Compatibility boundary

Parsing without `includes=` retains unresolved-AST behavior. `Include.filename` is renamed to `Include.target`; `filename` remains a deprecated compatibility alias. Include failures expose stable identities for missing, invalid, ambiguous, cyclic, overly deep, and inconsistent resolution conditions.


source/language/includes.rst
```diff
@@ -0,0 +1,225 @@
...
+
+* resolving the same include target from the same containing source unit shall
+  produce the same canonical name;
+* a given canonical name shall produce the same source text;
+* an include target that is *relative*, in whatever sense the implementation
+  defines relativity, shall be resolved with respect to the canonical name of
+  the containing source unit, and shall not be resolved with respect to a
+  process working directory or any other ambient state;
+* an implementation must diagnose an inclusion cycle; and
+* an implementation must support an inclusion depth of at least 15.
```
jwoehr: Perhaps leave inclusion depth implementation-defined?

jwoehr (dismissed):

source/language/includes.rst
idavis: I had the same thought, but I put it here to trigger a conversation. I'm happy to go either way, but we'll need a good default for a reference implementation.

source/language/includes.rst
```diff
@@ -0,0 +1,244 @@
...
+locations, against an in-memory table supplied by an embedding host, against
+unsaved editor buffers, or by any other means.
+
+Although the mapping is implementation defined, its observable behavior is not.
+Within a single compilation of a single program:
+
+* resolving the same include target from the same containing source unit shall
+  produce the same canonical name;
+* a given canonical name shall produce the same source text;
+* an include target that is *relative*, in whatever sense the implementation
```
hodgestar: We shouldn't leave the question of which include targets are relative up to the implementation.

For example, in `mygates.inc`:

```
include "one_qubit_gates.inc";
...
```

The specification should say whether `"one_qubit_gates.inc"` should be treated as relative to `"mygates.inc"` or not.

If it isn't specified, libraries want to include their own components won't have a behaviour to rely on.

In the example above, my intention was that `"one_qubit_gates.inc"` would be a relative include, but currently this is not guaranteed by the proposed spec update.

idavis: what if you have `one_qubit_gates.inc` defined in another working folder that you want to be able to reuse without copying?  Adding that folder as a search path a filesystem based resolver would allow looking locally then falling back to the search path. Making it part of the spec prohibits that.

source/language/directives.rst
```diff
@@ -51,6 +51,10 @@ intact for future passes.  Pragmas should avoid stateful or positional
 interactions to avoid unexpected behaviors between included source files. If the
 position is relevant to a pragma, an annotation should be used instead.
 
+A pragma contained in an included source unit is processed when that source
+unit is incorporated into the program, which happens at most once however many
+``include`` statements resolve to it.  See :ref:`include-once`.
```
jakelishman, line 56: Just for clarity: this statement is purely informative, right?  As I understand the intention of `include-once`, this is just a direct consequence of that, and any other behaviour would be very misleading.

source/language/includes.rst
```diff
@@ -0,0 +1,244 @@
...
+
+Resolution
+----------
+
+The following terms are used throughout this section.
+
+*Include target*
+   The string literal written in an ``include`` statement.  An include target
+   is an opaque logical name.  It is not required to be a file-system path, and
+   an implementation is not required to interpret it as one.
```
jakelishman, line 33: If we keep this: using `include "filename";` as the example statement in the first paragraph of this file is misleading: `include "target";` might help poke people to not think of it as a file name.

source/language/includes.rst
idavis: I think specifying exact load order is better served by a manifest/project.

source/language/includes.rst
```diff
@@ -0,0 +1,244 @@
...
+than redeclaration.  ``defcal`` statements originating from different source
+units overload one another in exactly the same way as ``defcal`` statements
+written in a single source unit.
+
+Repeated inclusion
+~~~~~~~~~~~~~~~~~~
+
+There is no mechanism for incorporating the same source unit into a program
+more than once.  A source unit intended to contribute different content at
+different points of a program cannot be expressed by repeated inclusion.
```
jakelishman, line 409: This is definitely a breaking change to the language, so we'll need to bump the language version.  (Not saying that's a problem.)

jakelishman (commented):
Overall, I think this is trying to solve a very real problem, but I'm not sure that trying to solve it within `include` itself leads to the best solution.

I understand the main motivation here to be to avoid "library" files from defining the same `gate`s (or other things that can't be shadowed) more than once when more than one file includes the same library.  I think this PR is trying to thread the needle between `include` statements historically always having been files (or at least files that a compiler _pretends_ exist), and the massive difficulty in canonicalising based on that.

I think that this change is already breaking the language semantics and requires a new language version (not a bad thing).  Given we _are_ doing that, though, I think we should consider if it's easier not to try to redefine `include` statements, but instead jump straight to using a more modern PL solution to this problem, like some concept of "modules" in the language.

For example: could we leave `include` as before this PR, including all the horrible multi-definition errors, and then _add_ `import <module>;` instead (plus/minus _also_ defining namespacing)?  Now we do uniqueness by the resolved "complete" module name (where we have control on how relative/absolute module names are spelled), and the rules for "how to look up a module in the filesystem" can come _after_ canonicalisation, rather than before, so they're easier.

source/language/directives.rst
idavis: Correct. And pragmas aren't 'ordered' in any way so no matter when they are included the behavior should be the same.

source/language/includes.rst
idavis: It depends on how you define breaking. No legal program will change behavior. If a program had a multiple include of a file, it would have been invalid and shouldn't have compiled. This change opens programs that would have been invalid to potentially now being valid. Any existing valid program should have no change in behavior.

jakelishman: > If a program had a multiple include of a file, it would have been invalid and shouldn't have compiled.

This part isn't true because there wasn't any symbol hygiene in includes previously.  For example `my_include.qasm`:
```qasm
a = a + 1;
```
and my main program:
```
OPENQASM 3.0;
output bit b;
uint a = 0;
include "my_include.qasm";
include "my_include.qasm";
include "my_include.qasm";
if (a == 3) {
   x $0;
}
b = measure $0;
```

I think this is a legal program in the previous version of the spec, whose output should be (noise/resets aside) 1 every time.  Adopting this spec change would cause it to be 0 every time instead.

idavis: Oy, that is both correct and horrible.

idavis: While that is true, part of me wonders if that is a use case we are willing to break without a full version bump.

jakelishman: Oh yeah, I wasn't suggesting it needed to be a new major version, just a minor realistically - my point was just motivating the top-level PR comment of (paraphrased) "since we're changing _some_ semantics, can we do it cleaner with a new system than trying to retrofit the old one?".

My only real concern with this (albeit a decent one) is having a program validity condition based on some sort of canonicalisation _after_ filename-but-also-maybe-not resolution means we have to specify way more stuff than if we can specify a different look-up system without the baggage of needing to support things that smell like filenames for backwards compatibility.

idavis:
@jakelishman @hodgestar Based on the feedback and the TSC call I think I was far too flexible and while import order was implicitly defined, resolution was far to relaxed. I'm reworking the PR to make this much more explicit and defined.

idavis:
@jakelishman @hodgestar @jwoehr Based on our conversation, I'll update the PRs to cover:
- a new `import` keyword and stmt that carries the semantics that we've discussed
- The `include` statement is deprecated and behavior is implementation defined.

Questions that I think we need to get a handle on:
- What does it mean to import a library vs a file in a library? This is important for the proposed `import "@std"` or `import "stdgates"` as those are effectively a libray import vs a file in a library, unless we want `import "@std/gates.inc"`
- How can an implementation supply their own magic library like `stdgate.inc`?

source/language/includes.rst
idavis: Closing as the PR is completely reworked.

PR #679, opened 2026-08-05
Define version behavior for included OpenQASM files

idavis:
## Functional summary

This branch layers one effective OpenQASM version over the complete program. The root source selects the version used to validate every source unit, while each source retains its own explicitly declared version as author-intent
metadata.

> Specifies and implements version parsing behavior brought up in #611, designed to supercede #669 based on #678

## Apply one effective version

The root header selects the effective version for the complete include graph. An included source keeps its explicit header as metadata but is parsed under the root's effective version.

```python
import openqasm3

source = """\
OPENQASM 3.1;
include "gates.inc";
qubit q;
custom q;
"""
includes = {
    "gates.inc": """\
OPENQASM 3.0;
gate custom q { U(0, 0, 0) q; }
""",
}

program = openqasm3.parse(source, includes=includes, name="main.qasm")
included = program.sources[1]

assert program.version == "3.1"
assert program.effective_version == "3.1"
assert included.version == "3.0"
assert included.effective_version == "3.1"
```

An included header may declare the root version or an earlier version. A later declaration is rejected because one compilation cannot switch language rules partway through its include graph.

## Enforce feature introduction versions

The parser recognizes OpenQASM 3.0, 3.1, and 3.2 and rejects syntax used before its introduction version.

```python
import openqasm3
from openqasm3.parser import QASM3ParsingError

try:
    openqasm3.parse("OPENQASM 3.0; switch (0) { case 0 {} }")
except QASM3ParsingError as error:
    assert "'switch' requires OpenQASM 3.1 or later" in str(error)
else:
    raise AssertionError("OpenQASM 3.0 must reject switch statements")

program = openqasm3.parse("OPENQASM 3.1; switch (0) { case 0 {} }")
assert program.effective_version == "3.1"
```

Anonymous scopes and `switch` require OpenQASM 3.1. `nop` requires OpenQASM 3.2. Version-related parsing failures expose stable optional error identities for callers that need machine-readable diagnostics.

## Handle a missing root header

A headerless root defaults to OpenQASM 3.0 and emits `MissingVersionWarning`. Headerless included sources inherit the root effective version without producing an independent default.

```python
import warnings

import openqasm3
from openqasm3.parser import MissingVersionWarning

with warnings.catch_warnings(record=True) as captured:
    warnings.simplefilter("always")
    program = openqasm3.parse('include "stdgates.inc";')

assert program.version is None
assert program.effective_version == "3.0"
assert [warning.category for warning in captured] == [MissingVersionWarning]
```

## Compatibility boundary

The `parse` signature and include-resolution model remain unchanged. Callers that intentionally accept unsupported versions or bypass feature gates can use `ignore_version=True`; the resulting AST still records the supplied effective version. Flattening includes preserves the program-wide effective-version metadata.


jwoehr (dismissed):
LGTM

